PR #9, opened 2019-02-27
GF#20480, removed yearIsValid()

patrick-werner:
arIsValid() currently checks if the year i is i >= 1800 && i <= 2100

The FHIR spec only has a regex to validate dates and doesn't has restrictions on min or max years, i propose to just get rid of yearIsValid() and do only the regex check.


patrick-werner:
Tracker: GF#20480

grahamegrieve:
see discussion here: https://chat.fhir.org/#narrow/stream/179165-committers/topic/yearIsValid().20in.20InstanceValidator

grahamegrieve:
This was changed directly after the zulip discussion.

PR #12, opened 2019-03-21
Set id value urn

patrick-werner:
zulip discussion:
https://chat.fhir.org/#narrow/stream/179165-committers/topic/IdType.20sets.20urn.3A*.20for.20IDs/near/161373153


jamesagnew:
Hi @patrick-werner -

I'm not clear on why this is a good thing? If we see an ID like "urn:uuid:12345" why do we want to strip the prefix? This will definitely break how some stuff works in HAPI FHIR (transaction processing in the JPA server being an obvious example)

patrick-werner:
hey @jamesagnew,

this change is needed to comply to the the FHIR spec:
http://build.fhir.org/datatypes.html#id

> id:
>Any combination of upper- or lower-case ASCII letters ('A'..'Z', and 'a'..'z', numerals ('0'..'9'), '-' and '.', with a length limit of 64 characters. (This might be an integer, an un-prefixed OID, UUID or any other identifier pattern that meets these constraints.), | Regex: [A-Za-z0-9\-\.]{1,64}

patrick-werner:
sorry for not giving more context, here is the use-case:
When parsing bundles with parser.setOverrideResourceIdWithBundleEntryFullUrl(true); and you are using absolut URLs (http://...) you don't notice the replacement as everything is stripped from the URL to leave the ID, if you do the same with urns:
```
entry id: e3ddbded-31d3-4505-805a-03b0d941621e
fullUrl: urn:uuid:e3ddbded-31d3-4505-805a-03b0d941621e
```
you end up with a resource with has an ID of  _urn:uuid:e3ddbded-31d3-4505-805a-03b0d941621e_ which isn't compliant to the FHIR ID datatype.

So my issue is that IdType.setValue() behaves not consistently, but strips inputs to proper IDs for FullUrls, but just sets the input for urn: based values. So i am in favor of a more consistent implementation.
To solve the transaction problem we could have a boolean flag on IdType?

grahamegrieve:
@jamesagnew what do you want to do with this?

PR #28, opened 2019-05-23
Add test case for validator complaining about display names that are …

lmckenzi:
…defined in the value set.


grahamegrieve:
Why the changes  to the snapshot generation tests?

lmckenzi:
They weren't valid (missing mandatory elements).  I'd fixed them when working on a test case that I ended up not needing.  Figured I may as well commit the fix.

PR #52, opened 2019-06-20
Condition Convertor: Fix convertCondition*Status

tobiasgirardet:
FHIR specifies the defining URL for the condition's clinical status value set to be [0]:

`http://hl7.org/fhir/ValueSet/condition-clinical`

However, is seems that the converter contains a typo as the code currently uses this URL:

`http://hl7.org/fhir/condition-clinical`

This PR addresses this issue and makes sure that the converter creates valid FHIR Condition resources.

But, since a lot of FHIR resources in the real world don't use the correct URL, this PR also makes the converter a little bit more robust by accepting the following URLs as well:

`http://terminology.hl7.org/CodeSystem/condition-clinical`
(which is actually the HL7 code system that the values set for this status is based upon)
`http://hl7.org/fhir/condition-clinical`
(the formally solely used URL)

The same issues mentioned here also apply to the condition's verification status, which is fixed in this PR as well.

[0] http://hl7.org/fhir/valueset-condition-clinical.html

### DIFF
--- a/org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/VersionConvertor_30_40.java
+++ b/org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/VersionConvertor_30_40.java
@@ -8253,19 +8253,19 @@ public class VersionConvertor_30_40 {
     org.hl7.fhir.r4.model.CodeableConcept cc = new org.hl7.fhir.r4.model.CodeableConcept();
     switch (src) {
       case ACTIVE: 
-        cc.addCoding().setSystem(VALUE_SET_CONDITION_CLINICAL_URL).setCode("active");
+        cc.addCoding().setSystem(CODE_SYSTEM_CONDITION_CLINICAL_URL).setCode("active");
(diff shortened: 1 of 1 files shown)


grahamegrieve:
http://hl7.org/fhir/ValueSet/condition-clinical is not correct. the value set URL never appears in an instance. http://terminology.hl7.org/CodeSystem/condition-clinical is correct - please update to use that

tobiasgirardet:
You are right. That was a misunderstanding on my part. I changed the URL that is used for the newly created resources. However, do you agree that the converter should still be robust and accept more than this specific URL as input?

grahamegrieve:
Well.. not sure. I'm doubtful about accepting the value set URL... Are you seeing them in the wild? I feel as though people would take the fact that it supports that as blessing that it's ok...

tobiasgirardet:
I agree. It is better to check if resources are valid before passing them to the converter. I removed the value set URL. However, the same is true for the 'legacy' URL http://hl7.org/fhir/condition-clinical . Should this be removed as well?

PR #59, opened 2019-07-09
CDA cda-core-2.0 tests and fixes

oliveregger:
Changed the CDARoundTripTests that it loads the hl7.fhir.cda model.

I had to adapt the XmlParser in one specific instance. All CDA Entries level have in text Element the type 'ED' which is defined as a choice type for string and binaryBase64. I handled this special case directly in the XmlParser, if there is a better way, I'm happy to add a more elegant version.

I had to adjust the model, i changed a few v3 datatypes to constraint instead of specialization (e.g http://hl7.org/fhir/cda/StructureDefinition/ST), because otherwise the snapshot got filled with multiple elements, see https://github.com/HL7/cda-core-2.0/pull/11/commits/1cdb38f17f6d7f1a7cfb91400e97f18787c40bc0 and https://github.com/HL7/cda-core-2.0/pull/11.

However the IG Publisher does not apply currently the constraint rules on the Logical Model, I will provide a pullrequest for that too, that means, that only if the pull request for ig publisher and then the pull request for the cda-core-2.0 model will be applied, this tests here succeed.

### DIFF
--- a/.project
+++ b/.project
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.hl7.fhir.core</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.m2e.core.maven2Nature</nature>
+	</natures>
+</projectDescription>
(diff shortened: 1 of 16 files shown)


oliveregger:
The ED data type is depended on the representation attribute and defaults to TXT, added the type decision depending on the representation attribute.

Additionally the logical model snapshot generation has been refined, if it is a specialisation, elements which are not used from the base definition (max 0) are removed. This allows, to cancel out the id, extension attributes of Element.

CDA validation works now, in the ccda example document there is an empty <effectiveTime/> tag which fails the validation.

oliveregger:
has been addressed by newer pull requests

PR #65, opened 2019-07-29
scanDirectory now scans all subfolders

patrick-werner:
If you pass the validator-cli a folder all subfolders are scanned recursively.


grahamegrieve:
Will accept this once there's a flag to scan sub-directories

PR #72, opened 2019-09-03
Generalize liquid

fil512:
This PR supports the hapi-fhir PR that adds Liquid Template support to hapi-fhir: https://github.com/jamesagnew/hapi-fhir/pull/1215

The two PR depend on one another and need to be merged at the same time.

About these changes:

- Version-specific LiquidEngine classes are replaced with a single version-independent LiquidEngine.
- The original LiquidEngine tests pass with this new version-independent class
- The bridge between LiquidEngine and the version-specific FHIRPathEngine is an interface called IFluentPath.
- IFluentPath and it's supporting interfaces (IExpressionNode, INarrativeConstantMap...) live in hapi-fhir-base [perhaps they should move to a package within `fhir.core`?]
- Version-specific implementations of IFluentPath are in `fhir.dstu3`, `fhir.r4` and `fhir.r5`.
- I didn't test the `fhir-dstu3` implementation--it looks like testing in that version requires setup outside of maven.  (Would appreciate if @grahamegrieve tries running the `LiquidEngineTests` I added to dstu3).


grahamegrieve:
> IFluentPath and it's supporting interfaces (IExpressionNode, INarrativeConstantMap...) live in hapi-fhir-base [perhaps they should move to a package within fhir.core?]

I think so. Is that the source of the mutual dependency? I think that should be avoided if possible.

Please explain the dependencies this introduces: 
- caffeine
- org.apache.commons.collections4.map

Why "FluentPath"? If this is to differentiate a version-independent interface to the FHIRPath engine to a version dependent one, then the name should indicate that

grahamegrieve:
@markiantorno what do you think we should do with this?

markiantorno:
Let me have a chat with @fil512 about this. Usually when it comes to merges this old...it's tricky.

jamesagnew:
One thing I'll mention just 'cause I see it in the discussion..

We only called the interface IFluentPath because back then that's what FHIRPath was called. That interface has since been renamed to IFhirPath (it hadn't been renamed yet when that PR was written but it was earlier this year).

PR #73, opened 2019-09-05
Fixed issues with version-conversion differentiating profile and targ…

lmckenzi:
…etProfile

Fixed problem where parser would get stuck in an infinite loop when allowing unexpected elements and actually encountering one
Fixed problem with localizing relative links in narrative if the reference starts with ../


grahamegrieve:
I don't understand the proposed changes to VersionConvertor_30_50.java?

lmckenzi:
In 3.0, you'd have a separate repetition for each "Reference".  In 5.0, all references get combined together into a single Reference repetition with multiple target profiles.  That works fine because all of the target profiles are distinct.  However, if in 3.0 you also had *profile* declarations (defining rules for the Reference data type itself rather than the target resource), those were all getting added as profiles on the single Reference.  Problem is that they could be (and typically were) the same.  So you'd end up with multiple declarations of the same 'profile' on the Resource element.  The new code checks for duplicates and only propagates a given profile once.

PR #76, opened 2019-09-07
Add pattern support to 2016may

lmckenzi:


grahamegrieve:
you know, I will merge this, but it's a bit odd, since it's being made in R3, but it's stated to be for 2016may. Is that really want? The R3 profile utilities are defunct for any tools we maintain

lmckenzi:
The 2016May stuff is authored using STU3 (but profiles resources expressed as 2016May)

PR #92, opened 2019-10-18
Fix precision and validator issues

### LINKED ISSUE #91: Date formatting should honour year/month precision
# Description
The FHIR dateTime datatype allows to use incomplete dates with just a year or with a year and month. However, in narrative generation (and maybe other situations as well), these incomplete dates are rendered as full dates, defaulting to the first day of the month (if only the day is absent) or to the first day of the year (if day and month are absent).

# Proposed fix
Formatting of the dateTime data type happens in the `toHumanDisplay()` or `toHumanDisplayLocalTimezone()` methods of the `BaseDateTimeType` class. This class does recognize the lower resolutions of year and month.

However, these methods leverage `FastDateFormat` of the Java platform to provide locale-aware date displays. And `FastDateFormat` doesn't have built-in formats for rendering dates in these lower resolutions.

One possible way to fix this is to modify the default `MEDIUM` format to display only what's needed (see [reference](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html)). This is quite tricky of course (as always with date and time), but with a proper fallback I think it might work.

For the year-only resolution, we could define a new format as:
```
private static final FastDateFormat ourHumanYearFormat = FastDateFormat.getInstance(ourHumanDateFormat.getPattern().replaceAll("[^y]", ""));
```
This filters out anything that's not part of the year rendering, while keeping the specific year format intact for the `MEDIUM` format.

The year/month resolution is quite a bit trickier, as there are many ways to arrange year, month and days across locales, with a choice of punctuation marks and such. A safe approach might be to replace the entire block of day+month with just the month. This leaves the formatting in relation to the year intact. If day and month are not adjacent to each other, a fallback is provided. Example:
```
private static final FastDateFormat ourHumanYearMonthFormat;
static {
  String format = "yyyy, MMM"; // Fallback for when the medium format doesn't use month and day of month adjacent to each other
	
  String medium = ourHumanDateFormat.getPattern();
  Pattern pattern = Pattern.compile("(M([^\\p{Lower}\\p{Upper}]*d)|(d[^\\p{Lower}\\p{Upper}]*)M)");
  Matcher matcher = pattern.matcher(medium);
  if (matcher.matches()) {
    format = matcher.replaceAll("M");
  }
  ourHumanYearMonthFormat = FastDateFormat.getInstance(format);
};
```

Finally, the getHumanDisplay...() methods can be modified to:
```
switch (getPrecision()) {
  case YEAR:
    return ourHumanYearFormat.format(value);
  case MONTH:
    return ourHumanYearMonthFormat.format(
  case DAY:
    return ourHumanDateFormat.format(value);
  ...
```

jamesagnew:
Fixes #91 

* [x] Honour the precision when formatting datetime values with less than DAY precision
* [x] The validator should report an error when a JSON decimal number contains a trailing dot (e.g. `123.`)

### DIFF
--- a/org.hl7.fhir.dstu2/src/main/java/org/hl7/fhir/dstu2/model/BaseDateTimeType.java
+++ b/org.hl7.fhir.dstu2/src/main/java/org/hl7/fhir/dstu2/model/BaseDateTimeType.java
@@ -25,6 +25,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.Validate;
 import org.apache.commons.lang3.time.DateUtils;
 import org.apache.commons.lang3.time.FastDateFormat;
+import org.hl7.fhir.utilities.DateTimeUtil;
 
 import java.text.ParseException;
 import java.util.*;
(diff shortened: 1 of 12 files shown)


jamesagnew:
@grahamegrieve do you mind to give this a review? I had to modify the JsonParser so I think this merits a look

PR #93, opened 2019-10-18
mitigate snapshot generation for cda-core-2.0 model with id/extension

oliveregger:
This pull requests tries to mitigate the problems raised by [GF#25031](https://gforge.hl7.org/gf/project/fhir/tracker/?action=TrackerItemEdit&tracker_item_id=25031) and [cda-core-2.0 issue](https://github.com/HL7/cda-core-2.0/issues/21).

The proposed intermediate fix until the principal problem has been addressed is to adapt the snapshot generation process. If in a logical model the differential the elements (id, extension) are introduced with a different representation (change from XmlText to XmlAttribute or inverse) the differential element is integrated instead of the base definition. This is the same behaviour if for a logical model only the snapshot would be defined and not relaying on any differential snapshot generation. Relying on a differential snapshot generation instead of providing a (wrong) snapshot has the benefit of not allowing to introduce potential inconsistencies. 

Two (logical2, logical2) snapshot generation have been added to illustrate the input, expected output snapshot generation.

This is however not a solution to the base problem that:
- if a logical model has a different representation of id, extension then the FHIR base Element can have a conflict with the base definition
- An identical name for an attribute and child element can be represented in an xml schema but will fail the constraints of a StructureDefintion (e.g. snapshot.element.all(id) and snapshot.element.id.trace('ids').isDistinct()) because they differ only in the representation attribute which is not present in the id element.

--
Background ( see also [GF#25031](https://gforge.hl7.org/gf/project/fhir/tracker/?action=TrackerItemEdit&tracker_item_id=25031) and [cda-core-2.0 issue](https://github.com/HL7/cda-core-2.0/issues/21) )

Some V3 Acts have an id element inside, e.g:

ClinicalDocument.id [1..1]
SubstanceAdministration.id [0..*]
This V3 Acts are represented as Logical Models in FHIR. Logical Models have a base of "Element" or another logical model. The base content of the FHIR Element has already an id and an extension element, which have different datatypes (string) and defined cardinality.

This leads to the following problems:

SubstanceAdministration.id: illegal constrained type http://hl7.org/fhir/cda/StructureDefinition/II from string in http://hl7.org/fhir/StructureDefinition/Element
Element SubstanceAdministration.id: derived max (*) cannot be greater than base max (1)
A related problem occurs with the II datatype which has an extension attribute which conflicts with the extensions defined in Element.

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/ProfileUtilities.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/ProfileUtilities.java
@@ -466,7 +466,7 @@ public class ProfileUtilities extends TranslatingUtilities {
         baseSnapshot = cloneSnapshot(baseSnapshot, base.getType(), derived.getType());
       }
       processPaths("", derived.getSnapshot(), baseSnapshot, diff, baseCursor, diffCursor, baseSnapshot.getElement().size()-1, 
-          derived.getDifferential().hasElement() ? derived.getDifferential().getElement().size()-1 : -1, url, webUrl, derived.present(), null, null, false, base.getUrl(), null, false, new ArrayList<ElementRedirection>(), base);
+          derived.getDifferential().hasElement() ? derived.getDifferential().getElement().size()-1 : -1, url, webUrl, derived.present(), null, null, false, base.getUrl(), null, false, new ArrayList<ElementRedirection>(), base, derived.getKind());
       if (derived.getDerivation().equals(TypeDerivationRule.SPECIALIZATION)) {
         for (ElementDefinition e : diff.getElement()) {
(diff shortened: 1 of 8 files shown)


grahamegrieve:
I'm kind of cool about this work around. I feel like this is going in the wrong direction.

oliveregger:
Looking forward how the discussion continues on [Zulip](https://chat.fhir.org/#narrow/stream/179177-conformance/topic/Loglcal.20Models.20.26.20Element), personally I would be also in favor of an abstract base with no id, extension.

oliveregger:
I agree with you @grahamegrieve that this pull request went into the wrong direction and I close it now. It provoked validation errors that the id is then not in order for a ClinicalDocument (in Element in in the attribute, in a CDA it comes much later). I think the only way to solve it, is to introduce an abstract Base with no id/extension or another way stating that behaviour.

PR #111, opened 2019-12-05
With revised code, dirForFile sometimes returns null and it's no long…

lmckenzi:
…er necessary


grahamegrieve:
Made a different change that actually solved the problem

PR #113, opened 2019-12-11
extracted errorMessage building

patrick-werner:


org.hl7.fhir.validation/src/main/java/org/hl7/fhir/r5/validation/InstanceValidator.java
```diff
@@ -859,7 +859,7 @@ public void validate(Object appContext, List<ValidationMessage> errors, Element
   private void checkElementUsage(List<ValidationMessage> errors, Element element, NodeStack stack) {
      String elementUsage = element.getUserString("elementSupported");
     hint(errors, IssueType.INFORMATIONAL, element.line(),element.col(), stack.getLiteralPath(), elementUsage==null || elementUsage.equals("Y"),
-        String.format("The element %s is not marked as 'mustSupport' in the profile %s. Consider not using the element, or marking the element as must-Support in the profile", element.getName(), element.getProperty().getStructure().getUrl()));	  
+        "The element " + element.getName() + " is not marked as 'mustSupport' in the profile " + element.getProperty().getStructure().getUrl() + ". Consider not using the element, or marking the element as must-Support in the profile");
```
jamesagnew, line 862: If you're tweaking the way this is generated, it'd be more efficient to use the parameterized form of hint(..) so that the error message string isn't built unless the rule actually triggers

patrick-werner: will do, this is only an intermediate step for better parsability on my side.

jamesagnew:
Ah right. Sorry, you already explained this to me. :)

PR #120, opened 2019-12-31
fixed nullPointer

patrick-werner:
When validating against an IG with profiles where derivation isn't set the validator throws some null pointers here:
https://github.com/hapifhir/org.hl7.fhir.core/blob/630ab0afcba49043ff2b145217e6735baddfaba2/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/ProfileUtilities.java#L489
As derived.getDerivation() returns null.

This PR checks for null first.


grahamegrieve:
Did something slightly different after discussion with Patrick

PR #121, opened 2020-01-10
FHIR to CDA with FHIR mapping language and cda-core-2.0

oliveregger:
this pull requests enable the java validator to run transforms from fhir to cda, see https://github.com/ahdis/cda-core-2.0/tree/develop/maps for the different maps. The following changes have been made:

- adjusted snapshot generation for xmlAttributes, for a specialization the order is still important (first specialized attributes have to be still on top, nested attributes then in order)
- fixed nullpointer exception when converting narrative text from fhir to cda
- fhirpath enhancement that conformsTo works also on profiles for types
- xsi:type issues (in serialization full url was add instead of type)
- fix for creating xsi:type (e.g. effectiveTime with EIVL_TS) in FHIR Mapping language (xsi:type was serialized as element)
- support for multiple namespaces in cda (e.g. pharm)


grahamegrieve:
I have lots of questions about this PR, but it's all too much for a single PR. Can you re-submit it as a set of individual PRs so we can work through the changes please

PR #128, opened 2020-01-29
Wrap all conversion accessors

markiantorno:


jamesagnew:
I love how quickly that source parser library allowed you to pull this change together.

@grahamegrieve - thought for another day, but I wonder if it would be feasible one day to have the Java converters generated automatically based on the FHIR mapping language transformation files..

grahamegrieve:
looks good to me. 

In theory, it would be possible to generate these from the mapping language, but it would be a massive task to reconcile the current versions - java is authoritative for conformance resources

jamesagnew:
Yeah makes sense.

PR #132, opened 2020-02-12
Updated accessors in main convertor files.

markiantorno:
@grahamegrieve Let me see if the tests pass for this pull req first, then i'll merge this, and apply the same changes to the other convertor files. I've made the changes you suggested, ie. non-generic conversions that fail at runtime.


markiantorno:
Need to step through the load package and figure out why it's loading for Grahame and not for me. 
Could be a caching issue...or something else?

PR #135, opened 2020-02-14
updating unit tests so they pass on my mac. Also updating test to Junit5

markiantorno:


grahamegrieve:
Right. no comment needed in the change log

PR #136, opened 2020-02-17
Fix ordering issue in converter

jamesagnew:
@markiantorno - The convertType method appears to be doing the types alphabetically, but this doesn't always work as there is a hierarchy. E.g. you have to test for UnsignedIntType before testing for IntegerType since the former is a subclass of the latter.

This PR fixes this for the 10/30 converters but I'm not sure if these are generated or not (i.e. what's the best strategy for fixing it for the other versions?)


grahamegrieve:
the best way is to iterate the primitives twice, once checking for things that don't inherit from primtiveType and once for what is left over

markiantorno:
The conversion issues you mentioned @jamesagnew are taken care of in this pull req (https://github.com/hapifhir/org.hl7.fhir.core/pull/132). I no longer convert using the generic method, and instead convert using the specific method (ie convertUnsignedInt). Grahame had requested this fix as well. 

I think Grahame just did the merge this morning, so the issues should be solved. I appreciate your patience.

PR #140, opened 2020-02-21
Instance validator code org

markiantorno:


grahamegrieve:
can you explain the thinking behind this?

markiantorno:
As I'm going through the code to refactor for the localization, I just thought this would make it easier to read. The class sizes seemed a little large to me, and I didn't want to add more code to them to make them even larger, without looking at any dependencies that could be pulled out into their own class files. There didn't seem to be a reason to have these classes directly nested in the main validator class as they were not tightly coupled, and were just data classes with the exception of one.

 There were two other classes in the validator that were very tightly bound with the validator code...so I could not pull those out. As I get a better understanding, I may refactor those as well.

At the end of the day, these changes are just for quality of life, and readability.

PR #141, opened 2020-02-24
Validator cleaning

markiantorno:
@grahamegrieve here are the small changes we discussed in out meeting for the validator packages


jamesagnew:
Nice!

PR #151, opened 2020-03-07
fixed convertors calling Enumeration with deprecated constructor

patrick-werner:
some convertor classes were calling Enumeration with the deprecated constructor causing NPEs


patrick-werner:
preapplied to: #153

PR #152, opened 2020-03-10
Fix for the issue 1740 (https://github.com/jamesagnew/hapi-fhir/issues/1740)

surajpjannu:
Describe the bug
=============
OperationDefinition has a field called resource. Which has a cardinality of 0..*.
anything otherthan array should be validation failure.
But if we try passing string as an input the validation is passing.
Tested with Some other resource as well, where the field is List<?> (where ? can be code,string etc) all or showing the same behaviour.

Posted a question in Google HAPI group but couldn't get the satisfactory answer. https://groups.google.com/forum/#!topic/hapi-fhir/locCV7PicNw

So raised a bug in hapi-fhir project.
https://github.com/jamesagnew/hapi-fhir/issues/1740

Soultion
======
In JsonParser, conditions was merged in one if ccondition. i think its better to split it.
Made corresponding changes for R4 FHIR versions.

But for DSTU3 , DSTU2016MAY and R5 corresponding changes has to be done.


patrick-werner:
fixed: c0c5b7f7acebe471b4d30182ce4ea8f1b9254341

Issue #167, opened 2020-04-07
Fix Broken Validator Version Flag


alexanderkiel:
If I call the validator with:

```
java -jar validator.jar -version 4.0.1 -ig ig.pack examples
```

It tries to detect the FHIR version from the IG name `ig.pack` but can't. So it uses the default with is `null`. The earlier detected version is overridden with `null`. The validator crashes.

If I however call it like this:

```
java -jar validator.jar -ig ig.pack -version 4.0.1 examples
```

The version is taken from the version flag because it comes after the detection from the IG flag.

PR #168, opened 2020-04-07
Fix Broken Validator Version Flag

alexanderkiel:
I'm not sure if that is the way the version detection though IG should work. But that is my take on it.

Closes: #167


markiantorno:
Makes sense.

PR #178, opened 2020-04-21
Fix for os agnostic utilities tests

markiantorno:
Previously, paths were hard-coded to only pass on windows machines. I assume I am not the only one who has chosen a life a suffering, and programs on Linux.  This fix allows these tests to pass on Linux and OS X machines as well.


markiantorno:
I put this in the main testing PR, disregard.

PR #185, opened 2020-04-29
Fix for HAPI-FHIR issue 1820

markiantorno:
https://github.com/jamesagnew/hapi-fhir/issues/1820

Fix for the above issue.


markiantorno:
So.

I had a lot of trouble getting the tests to pass on this one, and I discovered it was an issue with some parallelism related to the JUnit5 runner, the JUnit Vintage runner, and the version of Surefire we were using.

This whole JUnit5 thing involved a lot more than I realized.

I nuked all traces of the previous JUnit tests, updated the surefire version to the latest, and cleaned up some of the tests and updated them where needed. Now the vintage runner in no longer in the project, and this pull request has in turn become massive.

[God has cursed me for my hubris, and my work is never finished](http://www.youtube.com/watch?v=GQHVQvePdtI)

PR #186, opened 2020-05-01
ConceptMap updates for changed relationship types

rhausam:
ConceptMap updates for 'source-is-narrower-than-target' (formerly 'broader') and 'source-is-broader-than-target' (formerly 'narrower') relationships.


pieter-edelman-nictiz:
(please ignore, I used the wrong number in a tag and can't seem to undo it now)

PR #197, opened 2020-05-14
Deserializing CDA ST and ED elements from JSON fix

oliveregger:
i'm trying to split up hapifhir/org.hl7.fhir.core#121 in different steps.

this pull requests test the serialization/deserialization of the the ST definition from the cda-core logical model. the cda title is of type ST. see added test data which needs to be added from this pull request https://github.com/FHIR/fhir-test-cases/pull/19.

with xml everything works fine, the json parser stumbles during deserializing (not serializing) that the ST type is a choice type but has only one Element (ED the base has two). With attached patch for the JsonParser the tests work and the cda.title assertion works.


grahamegrieve:
Why is this change required? 

add || property.getDefinition().getPath().endsWith("data[x]")

oliveregger:
data[x] has in the ST snapshot only one type (dataString), so it does not go into the choice deserialization part (reguires more than one type in the isChoice function). With the || it overwrites the isChoice function. I did not wanted to adapt the isChoice function because i was afraid of side effects.

PR #198, opened 2020-05-14
Add check to avoid NPE

lmckenzi:


grahamegrieve:
Thanks but will solve this differently

PR #218, opened 2020-06-03
FHIR Mapping Language support for conformsTo function

oliveregger:
The FHIR Mapping Language can use FHIRPath function like the conformsTo function. The FHIR Mapping Language implementation is based on Element and not on Resources, the conformsTo function in StructueMapUtiliites has been extended that it works also in the FHIR Mapping Language context.

1. A test case has been written and added in the pull request, see https://github.com/FHIR/fhir-test-cases/pull/41 to call the conformsToFunction (in https://github.com/FHIR/fhir-test-cases/pull/41/commits/b518e95f3c422fc43627c0a2efd0c4ed5a5bc316#diff-a59de46d0f4e5b94ea8e023ff66c9d0d)
2. The r4 fml tests and data has been removed since the test was disabled and the test maps are now in the fhir-test-case project.
3. The  r5 fml test has been reenabled, the test has been moved to the org.hl7.fhir.validation test package because StructureMapUtilties need an InstanceValidatorFactory


grahamegrieve:
Why does this add tests cases here instead of to test cases?

oliveregger:
> Why does this add tests cases here instead of to test cases?

the test data itself is added to test cases (see pr FHIR/fhir-test-cases#41), this pr here only reenables the test itself to run on the test case data, removes the old r4 test data and adds the conformsTo for Element. Should it be done differently?

PR #236, opened 2020-06-09
XmlWriter for UTF-8 should not encode umlaut in attributes (elementmodel)

oliveregger:
XmlWriter is used in the serialisation of the Element in r5.elementmodel (e.g. after a FHIR Mapping Language transform) and XmlWriter does currently encode umlaute in attributes if we have an UTF-8 encoding.

We get  

```xml
<family value="Br&#xF6;nnimann-Bertholet"/>
```
instead of
```xml
<family value="Brönnimann-Bertholet"/>
```

XmlWriter uses already in other places XMLUtil.escapeXML which encodes depending on the specified encoding of the output. This PR adds this function also for the attribute and provides a unittest to illustrate the behaviour.

### DIFF
--- a/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xml/XMLWriter.java
+++ b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xml/XMLWriter.java
@@ -246,7 +246,7 @@ public class XMLWriter extends OutputStreamWriter implements IXMLWriter {
 					write(element[0]);
 					write("=\"");
 					if (element[1] != null)
-						write(xmlEscape(element[1]));
+						write(XMLUtil.escapeXML(element[1], charset, false));
 					write("\"");
 				}
 			}
@@ -254,18 +254,6 @@ public class XMLWriter extends OutputStreamWriter implements IXMLWriter {
 		return col;
 	}
 
-	 protected String xmlEscape(String s) {
-     StringBuilder b = new StringBuilder();
-     for (char c : s.toCharArray()) {
-       if (c < ' ' || c > '~') {
-         b.append("&#x");
-         b.append(Integer.toHexString(c).toUpperCase());
-         b.append(";");
-       } else
-         b.append(c);
-     }
-     return b.toString();
-   }
 	/* (non-Javadoc)
 	 * @see org.eclipse.ohf.utilities.xml.IXMLWriter#attribute(java.lang.String, java.lang.String, java.lang.String, boolean)
 	 */
--- a/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xml/XMLWriter.java
+++ b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xml/XMLWriter.java
@@ -246,7 +246,7 @@ public class XMLWriter extends OutputStreamWriter implements IXMLWriter {
 					write(element[0]);
 					write("=\"");
 					if (element[1] != null)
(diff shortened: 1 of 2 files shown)


oliveregger:
this PR provokes some double encoding??? I have to investigate why the test fails , will update PR later ... 


actual:
  <concept>
    <code value="_ActInformationAccessCode"/>
    <display value="ActInformationAccessCode"/>
    <definition value="**Description:** The type of health information to which the subject of the information or the subject&amp;#39;s delegate consents or dissents."/>

expected:
        <td>Designation of a concept&#39;s state.  Normally is not populated unless the state is retired..</td>

grahamegrieve:
Are you sure about the double encoding? Doesn't seem obvious how

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xml/XMLWriter.java
```diff
@@ -246,26 +246,14 @@ private int writeAttributeSet(boolean special, int col, int wrap) throws IOExcep
 					write(element[0]);
 					write("=\"");
 					if (element[1] != null)
-						write(xmlEscape(element[1]));
+						write(XMLUtil.escapeXML(element[1], charset, false));
```
oliveregger: i introduced the strange duplicate behaviour hier. element[1] ist already xmlEscape'd from the setAttribute functions. however the \r and \n are not yet converted to one line. the local xmlEscape function is doing that (c < ' ') but also for the umlaute (c > '~').

oliveregger: the NarrativeGenerationTests for the file actcode expects that the attributes are on one line and CR, LF are encoded in the attribute itself (e.g. same as  actcode-input.xml):

actcode-expected.xml: 

```xml
  <concept>
    <code value="_ActControlVariable"/>
    <display value="ActControlVariable"/>
    <definition value=".... snip ... pressure.&#xD;&#xA;&#xD;&#xA;Control variables ...snip..."/>
```

oliveregger:
now both tests should run through. the original escapeXml function is changed that only the characters below ' ' (like CR, LF) are escaped, the others are escaped already in the xxxAttribute function.

1. An open question for me is, if the provided actcode resource for the NarrativeGenerationTests should have &#xD;&#xA; inside in the attribute value and therefore expect it afterwards also for comparing (and failing the tests).

2. If it is required for XmlWriter that attributes are always serialized on one line this could be also coded directly with the XmlUtils.escape function (setting the third isNoLines to true).

@grahamegrieve if for this PR changes according to 1 or 2 should be made, I'm happy to update this PR.

grahamegrieve:
@oliveregger I've let this languish with guilt. Is it still a problem? I have a vague memory of fixing this in some other commit, maybe?

oliveregger:
@grahamegrieve thanks for getting back to it, a will retest and update the PR (or close if not necessary anymore), but won't be this week possible to do it for me

oliveregger:
@grahamegrieve : wish you  a happy new year! I revisited the PR. The testFunction to check that umlaute  are not encoded in UTF-8 in attributes fails without the PR (see org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/ResourceRoundTripTests.java, testSerializeUmlaut()). I rebased the PR to master and test works then fine. If you have any comments to the PR please let me know.

dhaag84 (approved):

dhaag84:
Hello everyone, I also would like to have those fixes upstream. Looks all good to me.

PR #237, opened 2020-06-09
override Element extension methods to also check for modifierExtensions

grahamegrieve:


jamesagnew:
Ahh got it- Of course, superclass doesn't have modifiers.

I'm not opposed, this is how HAPI's old structures worked too- there was a single set of method for dealing with both modifier and normal extensions.

Should probably have a javadoc though that indicates this behaviour.

grahamegrieve:
ok. Added javadoc as suggested

PR #249, opened 2020-06-16
Adding more verbose output for bad file path validator input

markiantorno:


markiantorno:
@grahamegrieve I have updated to move the new error message to the i18n tooling. However, I found some issues with the tools that needed to be addressed. I added unit tests to verify.

PR #260, opened 2020-06-25
Validate choice types with constrained out options

jamesagnew:


grahamegrieve:
thanks but decide to implement this slightly differently

PR #271, opened 2020-07-09
Fix bundle rendering

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/BundleRenderer.java
```diff
@@ -198,7 +198,7 @@ private void renderResponse(XhtmlNode root, BundleEntryResponseComponent respons
   }
 
   private void renderRequest(XhtmlNode root, BundleEntryRequestComponent request) {
-    root.para().addText("Response:");
+    root.para().addText("Request:");
```
markiantorno, line 201: Should request and response be put in two variables or is this the only spot they're used?

grahamegrieve: Only spot. There's a wider conversation to be had around i18n for the rendering but not as part of this PR

PR #274, opened 2020-07-15
Handle changing resource names when converting DataRequirements

grahamegrieve:


org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/VersionConvertorResourceNameMapper.java
```diff
@@ -0,0 +1,767 @@
...
+    if (name.equals("Task")) { return null; }
+    if (name.equals("TestReport")) { return null; }
+    if (name.equals("TestScript")) { return "TestScript"; }
+    if (name.equals("ValueSet")) { return "ValueSet"; }
+    if (name.equals("VisionPrescription")) { return "VisionPrescription"; }
+    return null;
+  }
+
+
+}
```
markiantorno, line 765: So, for something like this with so many if conditions, I would lean towards using a switch statement instead:

    

      switch (name) {

      case "AllergyIntolerance": return "AllergyIntolerance";

      case "Appointment": return "Appointment";

        ...
      default: return null;

    }

From a readability perspective, it's a lot less code and easier to parse visually, and switch statements are much more performant than if conditions, especially when the number of choices is large.

One other thing that you could do that would really clean it up and remove all the "..." Strings would be to use <class>.class.getName() for the types you were checking and returning. ex:

`case org.hl7.fhir.dstu2.Appointment.class.getName(): return org.hl7.fhir.dstu3.Appointment.class.getName();`

but I understand if that is too tedious. It just takes out any magic Strings, which aren't great.

grahamegrieve: you can do switch statements on strings? since when? 

But these absolutely are magic strings...

grahamegrieve: Did some of the changes. but the case statements have to be literal strings

PR #285, opened 2020-07-27
fix evaluation of codes in code systems that are labelled content = fragment

grahamegrieve:


grahamegrieve:
@markiantorno please review. I will try and figure out why this branch claims so many changes next time...

PR #294, opened 2020-08-06
Gg v5.0.20 work

grahamegrieve:


org.hl7.fhir.dstu2/src/main/java/org/hl7/fhir/dstu2/utils/client/FHIRToolingClient.java
```diff
@@ -87,6 +87,9 @@
 	public static final String DATE_FORMAT = "yyyy-MM-dd";
   public static final String hostKey = "http.proxyHost";
 	public static final String portKey = "http.proxyPort";
+  private static final int TIMEOUT_NORMAL = 1;
+  private static final int TIMEOUT_OPERATION = 2;
+  private static final int TIMEOUT_OPERATION_LONG = 3;
```
markiantorno, line 92: Can we put this in an enum?

org.hl7.fhir.dstu2016may/src/main/java/org/hl7/fhir/dstu2016may/utils/client/FHIRToolingClient.java
```diff
@@ -87,6 +88,9 @@
 	public static final String DATE_FORMAT = "yyyy-MM-dd";
   public static final String hostKey = "http.proxyHost";
 	public static final String portKey = "http.proxyPort";
+  private static final int TIMEOUT_NORMAL = 1;
+  private static final int TIMEOUT_OPERATION = 2;
+  private static final int TIMEOUT_OPERATION_LONG = 3;
```
markiantorno, line 93: Yeah, this is used in a few different places. Putting it in it's own class as an enum is the right approach.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/client/FHIRToolingClient.java
```diff
@@ -85,9 +85,10 @@
 	public static final String DATE_FORMAT = "yyyy-MM-dd";
 	public static final String hostKey = "http.proxyHost";
 	public static final String portKey = "http.proxyPort";
-  private static final int TIMEOUT_NORMAL = 1;
-  private static final int TIMEOUT_OPERATION = 2;
-  private static final int TIMEOUT_OPERATION_LONG = 3;
+  private static final int TIMEOUT_NORMAL = 15;
+  private static final int TIMEOUT_OPERATION = 30;
+  private static final int TIMEOUT_OPERATION_LONG = 60;
+  private static final int TIMEOUT_OPERATION_EXPAND = 120;
```
markiantorno, line 91: This is also in a couple places, might be good to put it in an enum or const as well.

PR #304, opened 2020-08-11
fix validator leak

grahamegrieve:


org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/tests/TestLoadIg.java
```diff
@@ -0,0 +1,96 @@
...
+        System.gc();
+        System.out.print("loading: allocated memory " + getUsedMemoryAsMbs() + " MB, ");
+        System.out.print("free memory " + getFreeMemoryAsMbs() + " MB, ");
+        System.out.println("max memory " + getTotalMemoryAsMbs() + " MB");
+
+        // The method under test:
+        hl7Validator.loadIg(id + (version != null ? "#" + version : ""), true);
+      }
+    } catch (Exception e) {
+      e.printStackTrace();
```
radamson, line 40: I'm not sure catching an exception would reliably check for a future regression here and probably wouldn't occur even with the existing leak after 3 iterations unless the system only provides a max memory of 1GB.

Another option could be to check the allocated memory after the first run, and then assert that after 5 runs the allocated memory hasn't increased by an unacceptable amount (e.g. allocated memory hasn't grown more than 25% or whatever is reasonable/expected).

grahamegrieve: I'm not sure that Java memory is consistent enough to make executable rules around this. but Mark is going to work further on the other test and I'll let him think about this.

radamson: I was just going to update and say since `System.gc()` only _suggests_ that the GC be run that the check might be a little aggressive and unpredictable.

I'm not super familiar with all the internal workings, but looking again a better test may be to check for duplicates in the `CanonicalResourceManger` `list` since it seems like that was the thing that was actually fixed.

grahamegrieve: Well, there's still actually a leak there. The first test doesn't leak noticeably because I fixed a bunch of other problems that made the original leak irrelevant. The second test cases exposes that directly. I considered making a test case for

PR #326, opened 2020-08-29
add readme for convertors

ItayGoren:


markiantorno:
Hey, this is great, but it might be better to put it in the github project wiki section instead. Realistically, we need a lot more documentation for the various modules, and i don't think it would be great practice to have separate README files in the various sections. Linking to the wiki from the main README would probably be the best way to go about this.

ItayGoren:
I agree we need more documentation but it's a start :) How can i get edit permissions for the wiki? Never worked with it before.

markiantorno:
I've pasted what you have submitted into a wiki page for the convertors:
https://github.com/hapifhir/org.hl7.fhir.core/wiki/Convertors

Thanks for taking the time to write this up. As we get more documentation, we can look at a better landing page for the wiki, and better organization. If you have any suggestions, please let me know!

markiantorno:
Closing, as the documentation was added to the wiki.

PR #329, opened 2020-08-31
changing validator cli return obj to include more information on error location

markiantorno:
Changed ValidationOutcome to include a list ValidationMessage instead of straight error messages so that line and column numbers can be extracted downstream.


grahamegrieve (approved):
ok

PR #333, opened 2020-09-02
Fix typo in ParametersRenderer and flag additional issue

lmckenzi:


grahamegrieve:
I added a test case for this, and so ended up fixing this in another PR instead. thanks.

PR #340, opened 2020-09-08
add patch wildcard for ig dependencies

oliveregger:
The [NPM fhir specification](https://confluence.hl7.org/pages/viewpage.action?pageId=35718629#NPMPackageSpecification-Versionreferences) says:

> When packages point to dependencies they should refer to the whole package version number and not use wildcards, except for the patch version in a semver version reference:
> 
>  "hl7.fhir.core" : "3.0.x"
> This x here means that it should a package resolver should accept the package with the highest found patch number.

This PR adds supports for the following use cases:
- if an ig dependency is defined as .x the latest patch version will used from the file package cache or installed from the package server
- if the ValidationEngine is called and a patched version is already specified before the dependency, the patched version will be used.

e.g.:
ch-core is depending on ch-epr-term 2.0.x, see [ig](https://github.com/hl7ch/ch-core/blob/oe_dependsonigversion/input/ch.fhir.ig.ch-core.xml)

```
java -jar org.hl7.fhir.validator_cli.jar -version 4.0.1 -ig ch.fhir.ig.ch-core#current HPWengerRole.xml 
```
will download ch-epr-term 2.0.4 (latest published version) into the package cache and will be used for validation

```
java -jar org.hl7.fhir.validator_cli.jar -version 4.0.1 -ig ch.fhir.ig.ch-epr-term#2.0.3 -ig ch.fhir.ig.ch-core#current HPWengerRole.xml 
```
will download ch-epr-term 2.0.3 into the package cache and will be used for validation, ch-epr-term 2.0.4 will not be used

```
java -jar org.hl7.fhir.validator_cli.jar -version 4.0.1 -ig ch.fhir.ig.ch-epr-term#2.0.4 -ig ch.fhir.ig.ch-core#current HPWengerRole.xml 
java -jar org.hl7.fhir.validator_cli.jar -version 4.0.1 -ig ch.fhir.ig.ch-core#current HPWengerRole.xml 
```
ch-epr-term 2.0.4 will be used for validation

This PR is depending on a PR for the IG Publisher to support building implementation guide, will follow after this one, since the IG Publisher is depending on this project.

### DIFF
--- a/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/VersionUtilities.java
+++ b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/VersionUtilities.java
@@ -193,6 +193,16 @@ public class VersionUtilities {
     }
   }
 
+  public static String getPatch(String version) {
+    if (version == null)
+      return null;
+    if (Utilities.charCount(version, '.') == 2) {
+      String[] p = version.split("\\.");
+      return p[2];  
+    }
+    return null;
+  }
+
   public static boolean isSemVer(String version) {
     if (Utilities.charCount(version, '.') != 2) {
       return false;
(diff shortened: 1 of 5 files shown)


oliveregger:
An additional PR for the IG Publisher is not necessary, as soon as the dependency to this project is added it works.

grahamegrieve (approved):
ok

PR #355, opened 2020-09-22
adding conversion from 30-40 for Device as per client request 

markiantorno:
So, @grahamegrieve, there was a support ticket put in at Smile that I was asked to look into, Device resources were not being converted dstu3 <-> r4. This should fix this, however, I was uncertain of if I managed to get everything correct.


grahamegrieve (approved):
ok

PR #357, opened 2020-09-25
GitHub packages

markiantorno:
Added support to deploy to both github packages and sonatype independently


grahamegrieve (approved):
well, looks ok to me - as if I would know

PR #360, opened 2020-10-06
fixed bug in conversion from dstu2 to r4 in timing resource

ItayGoren:


grahamegrieve (approved):
ok. thanks. we really need to do this everywhere

PR #361, opened 2020-10-07
performance: improve speed of comparing UriTypes

winne42:
While searching for performance bottlenecks in HAPI FHIR, we noticed that comparing FHIR resources was unnecessarily expensive: `equalsDeep()` calls `normalize()` even in case of string equality, which leeds to tons of unneeded URI objects; creation of these URI objects is in turn very expensive because it involves parsing the string constructor argument.

This very simple fix reduced our CPU load very significantly. There is probably more tuning potential in `normalize()`, but that's another story.


grahamegrieve (commented):
look good

grahamegrieve (approved):

PR #391, opened 2020-12-01
Load context from classpath allowing duplicates

azizjonh:
Added ability to specify whether to allow duplicates when loading context from classpath. In similar fashion to loading from package.


grahamegrieve (approved):
ok

PR #393, opened 2020-12-02
remove unnecessary (& problem causing) Fragment-Host from OSGi manifest

bdenton:
Adding "Fragment-Host" to the OSGi manifest is not needed and it short-circuits the normal OSGi class loading architecture. This has been creating issues for the few of us that are trying to build OSGi-based solutions for FHIR


grahamegrieve:
@markiantorno please review

jamesagnew (approved):

PR #412, opened 2021-01-12
Updated the info.cqframework dependency versions

KevinDougan-SmileCDR:
@markiantorno 
The existing v1.4.9 `info.cqframework` libraries are no longer available, so this PR will upgrade those libraries to the latest version.


markiantorno (approved):
+1

PR #417, opened 2021-01-18
Validator evaluate FHIRPath statements: Use FHIRPathEngine directly form validator

oliveregger:
The validator_cli.jar has a -fhirpath option to run fhirPath statements against resources. 

```
java -jar validator_cli.jar /Users/oliveregger/Documents/github/ch-orf/input/examples/bundle/referral-min-mapped.xml -version 4.0.1 -fhirpath 'entry[0].resource.subject.resolve().name.given'
````
However the FHIRPathEngine was currently not configured together with the validator, especially the resolve() function within a bundle does not work that way.

Bundle.entry[0].resource.subject.resolve().name.family
-> if the subject references a patient resource this should return the family name.

if the FHIRPathengine is used directly from the Validator the resolve function within a bundle is implemented and works.

This PR access the FHIRPathEnginge from the validator instead of creating a new one.


grahamegrieve (approved):
ok

PR #422, opened 2021-01-24
fixes missing depencency: okhttp3/Request$Builder

patrick-werner:
This PR adds the okhttp dependency to the validation module which is introduced in the R5 package as optional.

 Terminology server http://tx.fhir.orgException in thread "main" java.lang.NoClassDefFoundError: okhttp3/Request$Builder


markiantorno (changes requested):
please add the optional tag to this dependency

markiantorno (approved):
:+1: [thanks](https://www.youtube.com/watch?v=1mzUvDokTLE)

PR #425, opened 2021-01-24
Pw fix okhttp dependency exception

patrick-werner:


markiantorno (changes requested):
Please mark the dependency as optional

markiantorno:
Actually this should be fine. No one is importing the CLI as a dependency

PR #427, opened 2021-01-27
Memory and Performance enhancements

JPercival:
The motivation for this PR is to speed up the IG Publisher builds and reduce its memory usage. It was tested and benchmarked by building the [opioid-cds-r4](https://github.com/cqframework/opioid-cds-r4) IG.

The core change is to reuse the same instance of the FHIRPathEngine where possible in the JSONParser. This necessitated some changes in the FHIRPathEngine and supporting classes. All the changes were made to be backwards compatible.

In benchmarking, overall memory usage of the publisher was reduced by ~60% and the "Load Content" portion of the IG Build is about 6x faster, reducing overall build time by about 2 minutes.

Before:
Sending Usage Stats to Server (06:43.0788)
Memory (MB): Use = 6262, Free = 3628, Total = 9890, Max =11264

After:
Sending Usage Stats to Server (04:33.0507)
Memory (MB): Use = 2660, Free = 962, Total = 3623, Max =11264

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/ProfileUtilities.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/ProfileUtilities.java
@@ -330,6 +330,18 @@ public class ProfileUtilities extends TranslatingUtilities {
   private boolean autoFixSliceNames;
   private XVerExtensionManager xver;
 
+  public ProfileUtilities(IWorkerContext context, List<ValidationMessage> messages, ProfileKnowledgeProvider pkp, FHIRPathEngine fpe) {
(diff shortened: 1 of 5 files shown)


markiantorno:
Looks good to me, but @grahamegrieve should have the final say here.

grahamegrieve (commented):
ok. I'll take this further after the PR is accepted. Creating a FHIRPath Engine wasn't supposed to be costly

grahamegrieve (approved):

JPercival:
In my testing something like ~300k instances of the FHIRPathEngine were created, so even the relatively small cost in creating the HashMaps for caching the type information from the StructureDefinitions was expensive in aggregate.

PR #432, opened 2021-02-03
adding test for observation conversion 10_40

ItayGoren:
edit a test while debugging a bug - was found in another place but the test is still usefull


markiantorno (changes requested):
Please add a comment in the RELEASE_NOTES.md file to indicate the changes you have made.

ItayGoren:
added

PR #452, opened 2021-03-07
Fix validator crash when a nested folder is present in -ig <path>

vadi2:


markiantorno:
Unless I'm viewing this incorrectly...and I very well might be, it's saying that the only changes here are three lines, and those changes are indentations? Can you please detail the changes made?

vadi2:
Yeah for some reason codecov thought I changed a lot more than just 3 lines, huh? It makes the diff in Github unreadable.

The actual change is in https://github.com/hapifhir/org.hl7.fhir.core/pull/452/files#diff-02509189ba844ec01bf04b989cce1e7b4b8138c66812654ddc4f1bcb2fc69864R565, and faulty logic did not skip folders correctly when the `recursive` flag was `false`.

PR #457, opened 2021-03-12
Create hapi base package

markiantorno:
Base package from HAPI fhir has been moved into the core libraries. This was done to remove the circular dependency in the core, where HAPI depended on the core resources, and the core referenced HAPI as the parent pom file.


grahamegrieve:
so what's with this?

PR #467, opened 2021-03-24
add default status value to medication request conversion 10_40

ItayGoren:


markiantorno:
Please update the RELEASE_NOTES file if you are making a change.

ItayGoren:
added thanks

PR #469, opened 2021-03-29
support all the gender values

ItayGoren:
as documented here: https://www.hl7.org/fhir/DSTU2/valueset-administrative-gender.html
it is possible to specify M,F,A,O,U as genders in dstu2. added support for such values.
In addition, added support for M,F,UN,UNK for stu3

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,0 +1,1 @@
+* support all the possible gender values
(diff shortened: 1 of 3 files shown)


grahamegrieve:
Well, https://www.hl7.org/fhir/DSTU2/valueset-administrative-gender.html does not mean that it is possible to specify M,F,A,O,U as genders in dstu2; that's just telling you what the equivalent codes are in V2 and CDA

PR #474, opened 2021-04-09
Updates needed to support change of ConceptMap 'source'and 'target' e…

rhausam:
…lements to 'sourceScope' and 'targetScope'.


grahamegrieve:
I did this work as part of the general uplift of R5 code in core. I'm going to close this PR - check against the current version please.

Issue #453, opened 2021-03-09
VersionConvertor_30_50 convertTypeRefComponent does not copy and convert aggregation from R5 to DSTU3


tyfoni:
We have upgraded from hapi-fhir 5.0.2 to hapi-fhir 5.3.0, and the BaseValidator no longer validates the aggregation mode.
We have profiled the EpisodeOfCare and limits the diagnosis.condition to be "referenced": 

```

    <element id="EpisodeOfCare.diagnosis.condition">
      <path value="EpisodeOfCare.diagnosis.condition" />
      <type>
        <code value="Reference" />
        <targetProfile value="http://ehealth.sundhed.dk/fhir/StructureDefinition/ehealth-condition" />
        <aggregation value="referenced" />
      </type>
    </element>

```

We have a test that verifies that adding a condition "contained" will fail validation (hapi-fhir 5.0.2):

ValidationMessage[level=ERROR,type=STRUCTURE,location=Parameters.parameter[0].resource.ofType(Bundle).entry[0].resource.ofType(EpisodeOfCare).diagnosis[0].condition,message=Reference is contained which isn't supported by the specified aggregation mode(s) for the reference]

```
addValidationMessage:398, BaseValidator (org.hl7.fhir.validation)
addValidationMessage:393, BaseValidator (org.hl7.fhir.validation)
rule:255, BaseValidator (org.hl7.fhir.validation)
checkReference:2304, InstanceValidator (org.hl7.fhir.validation.instance)
checkChild:4151, InstanceValidator (org.hl7.fhir.validation.instance)
validateElement:4008, InstanceValidator (org.hl7.fhir.validation.instance)
checkChild:4182, InstanceValidator (org.hl7.fhir.validation.instance)
validateElement:4008, InstanceValidator (org.hl7.fhir.validation.instance)
startInner:3468, InstanceValidator (org.hl7.fhir.validation.instance)
start:3393, InstanceValidator (org.hl7.fhir.validation.instance)
validateResource:4723, InstanceValidator (org.hl7.fhir.validation.instance)
validateContains:3871, InstanceValidator (org.hl7.fhir.validation.instance)
checkChild:4165, InstanceValidator (org.hl7.fhir.validation.instance)
validateElement:4008, InstanceValidator (org.hl7.fhir.validation.instance)
checkChild:4182, InstanceValidator (org.hl7.fhir.validation.instance)
[...]

PR #501, opened 2021-05-18
Issue #453: add copying of aggregation mode for conversion from 50 to 30

tyfoni:
VersionConvertor_30_50 convertTypeRefComponent does not copy and convert aggregation from R5 to DSTU3 #453


grahamegrieve (approved):
looks good

Issue #502, opened 2021-05-19
Column number is inadvertently set to line number in Validator output


pieter-edelman-nictiz:
When I validate a resource using the latest FHIR Validator, in all the reported errors the column number is inadvertently set to the line number. For example:
```
FHIR Validation tool Version 5.3.14 (Git# d17a8da923f9). Built 2021-05-14T15:06:07.75Z (4 days old)
  Java:   1.8.0_292 from /usr/lib/jvm/java-8-openjdk-amd64/jre on amd64 (64bit). 3500MB available
  Paths:  Current = /mnt/c/Users/Edelman/Nictiz-R4-Zib2020, Package Cache = /home/pieter/.fhir/packages
  Params: -version 4.0 -ig resources/ -ig qa -recurse -profile http://nictiz.nl/fhir/StructureDefinition/ProfilingGuidelinesR4-StructureDefinitions-Zib-Profiles resources/zib/zib-BloodPressure.xml
--- snip ---
Done. Times: Loading: 00:09.0590, validation: 00:03.0307. Memory = 616Mb

Success: 0 errors, 2 warnings, 0 notes
  Warning @ StructureDefinition.differential.element[11].sliceName (line 162, col162) : sd-pg-13: 'Slice names should be camelCased starting with a lowercase letter.' Rule 'Slice names should be camelCased starting with a lowercase letter.' Failed
  Warning @ StructureDefinition.differential.element[14].sliceName (line 189, col189) : sd-pg-13: 'Slice names should be camelCased starting with a lowercase letter.' Rule 'Slice names should be camelCased starting with a lowercase letter.' Failed
```

Line 162 and 189 are indeed the locations of the offending elements, but the column numbers are reported incorrectly.

pieter-edelman-nictiz:
Note: this seems not to happen universally. I only can reproduce it on constraint failures.

PR #504, opened 2021-05-19
Fix for a bug where the line number was reported as column number when validating constraints

pieter-edelman-nictiz:
This should fix issue #502.


grahamegrieve (approved):
thanks

PR #506, opened 2021-05-20
If there's an ampersand that doesn't actually have a closing semi-col…

lmckenzi:
…on, the process eventually runs out of characters and then fails doing a charAt(0).  Catch this issue and throw a useful exception.

Specific example within file that caused an issue:
            feedbackurl = feedbackurl + '&';


grahamegrieve (approved):
thanks

PR #508, opened 2021-05-20
Allow extensions in checkFixedValue if a pattern instead of a fixedValue is defined on the element

### LINKED ISSUE #507: Extensions should be allowed on elements even if a pattern on the same element does not contain them
See https://github.com/FHIR/fhir-test-cases/pull/90

In the test case, the referenced profile contains a pattern on Condition.code.coding. In the instance some extensions are added to the same element which are not part of the pattern. The InstanceValidator handles the pattern the same as a fixed value and rejects the Coding with "No extensions allowed, as the specified fixed value doesn''t contain any extensions". If a pattern is used instead of a fixed value in the profile, extensions should be allowed.

alexzautke:
Fixes #507

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,1 +1,0 @@
-* update so I can push a new validator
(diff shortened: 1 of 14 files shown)


grahamegrieve (approved):
makes sense

PR #512, opened 2021-05-24
added regex for countrycodes identifier validation

patrick-werner:
- checks if a canonical url matches `^http:\/\/hl7.org\/fhir\/sid\/passport-[a-zA-Z]{3}$`
- missing: check if 3 letter code is a valid country code

Jira tracker: https://jira.hl7.org/browse/FHIR-19671


grahamegrieve (commented):
There are other places in the code that reference sids. This change should be made in all those places

grahamegrieve:
No. On further thought, and discovering I had made other SID related canges, I have condensed all the SID stuff into a single new class, which includes your code. So I am abandoning this PR. Please add a test case to the test cases instead.

PR #519, opened 2021-05-26
Conversion update ii

markiantorno:


grahamegrieve (approved):
Well, OK, but are going to move to not have static methods for the convertors - that's another commit? 

Not really a fan of stream methods either - they always seem to be more obscure and harder to debug to me.

PR #526, opened 2021-06-02
Fixed issue where when validating with no terminology server and a va…

lmckenzi:
…lue set with only an expansion (no compose), the 'inferred' code system wasn't being populated and validation was then failing on a coding with no specified code system


grahamegrieve (approved):
ok

PR #537, opened 2021-06-14
can not convert boolean abatement from dstu2 to r4

ItayGoren:


ItayGoren:
In the documentation there is no specification how to convert abatementBoolean to R4 and it is currently crashing.
Because R4 doesn't have such field, we ignore it in the conversion.

grahamegrieve (approved):
ok

PR #541, opened 2021-06-18
Static method convertor

markiantorno:
@grahamegrieve 

I pulled apart everything in the 40_50 convertor libraries. This is the same approach I want to take with the other version convertors, but I wanted to run it by you first and discuss. After we re-org all the convertor classes like this, I will make a factory class that we can just instantiate and call `convertResource(...)` on for all convertors. I was thinking we could use a syntax like `ConvertorFactory().with(BaseAdvisor).convertResource()`.

Let's talk about this Monday or Tuesday? Or even later today if you have time. It's a pretty substantial refactor. But, I think it opens up a lot of different options for structuring it going forward.


markiantorno:
Also, I updated to the latest test cases version 1.1.66-SNAPSHOT and they don't pass on this branch, or on master. I assume you are working on that?

PR #551, opened 2021-06-27
fixing procedure conversion from dstu2 to r4

ItayGoren:
the field dstu2.Procedure.notPerformed is optional. fixing a bug in convert for procedure which didn't convert the status field if notPerformed was not given


markiantorno (changes requested):
Resolve the merge conflicts and we can close this.

grahamegrieve:
We should close this one too?

grahamegrieve:
Close these due to merge conflicts. If the underlying issues are still a problem, re-submit based on the current code.

PR #554, opened 2021-07-04
Fixed issue where validator fails to detect invariant issue when ther…

lmckenzi:
…e are multiple profiles that use the same invariant on the same element.  Also addressed issue revealed by the fix where the context resource wasn't being set properly when checking certain invariants (identified by one of the existing test cases).

See new test case in https://github.com/FHIR/fhir-test-cases/pull/93

### DIFF
--- a/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
+++ b/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
@@ -4528,10 +4528,6 @@ public class InstanceValidator extends BaseValidator implements IResourceValidat
     String thisExtension = null;
     boolean checkDisplay = true;
 
-    // Lloyd: Changed onlyNonInherited to false because inherited invariants can be overridden and it's important to check both.  
-    // Caching logic should make checking inherited constraints more efficient now.
-    // Added choice around checkInvariants because azatadha/example-params-1 test was treating the '%resource' in the bdl-3 constraint for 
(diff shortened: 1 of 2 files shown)


org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
```diff
@@ -4528,7 +4528,16 @@ public void checkChildByDefinition(ValidatorHostContext hostContext, List<Valida
     String thisExtension = null;
     boolean checkDisplay = true;
 
-    checkInvariants(hostContext, errors, profile, typeDefn != null ? typeDefn : checkDefn, resource, ei.getElement(), localStack, true);
+    // Lloyd: Changed onlyNonInherited to false because inherited invariants can be overridden and it's important to check both.  
+    // Caching logic should make checking inherited constraints more efficient now.
+    // Added choice around checkInvariants because azatadha/example-params-1 test was treating the '%resource' in the bdl-3 constraint for 
+    // evaluating the Bundle rule as Parameters, not Bundle
+    SpecialElement special = ei.getElement().getSpecial();
```
markiantorno, line 4531: So, 
1. Please put this in the RELEASE_NOTES.md file, and not here. 
2. If you want to keep a comment here, which would be great, please reformat it just to say what the code does, no need for your name or added verbs describing the work process

markiantorno (changes requested):

markiantorno (approved):

PR #558, opened 2021-07-14
changing PackageHacker.useSecureReferences to true by default

markiantorno:


grahamegrieve (commented):
Well, I'm ok with the change to secure references, but where did the other changes come from?

grahamegrieve (commented):
so what happened to this change? it hasn't been made?

PR #566, opened 2021-07-27
Adding accessToken parameter to hl7validator for use in HTTP Authorization header calls to TX

david-simons:
In response to this Zulip chat:
https://chat.fhir.org/#narrow/stream/179239-tooling/topic/HL7.20Validator.20authentication.20support.20for.20terminology.20service

### DIFF
--- a/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/client/FHIRToolingClient.java
+++ b/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/client/FHIRToolingClient.java
@@ -58,10 +58,6 @@ import org.hl7.fhir.utilities.ToolingClientLogger;
 import org.hl7.fhir.utilities.Utilities;
 import org.hl7.fhir.r4.model.Enumerations;
 import org.hl7.fhir.r4.model.DateTimeType;
-//import org.hl7.fhir.r4.model.Enumeration;
-//import org.hl7.fhir.r4.model.Enumerations.FHIRVersion;
-//import org.hl7.fhir.r4.model.Enumerations.PublicationStatus;
-//import org.hl7.fhir.r4.model.codesystems.CapabilityStatementKind;
 
 /**
  * Very Simple RESTful client. This is purely for use in the standalone 
@@ -197,12 +193,11 @@ public class FHIRToolingClient {
       capabilities = (CapabilityStatement)utils.issueGetResourceRequest(resourceAddress.resolveMetadataUri(true), getPreferredResourceFormat(), "CapabilitiesStatement-Quick", TIMEOUT_NORMAL).getReference();
     } catch(Exception e) {
         // THIS IS A WORKAROUND FOR THE TERMINOLOGY SERVER RESPONDING WITH A 404 TO /metadata
-		CapabilityStatement cs = new CapabilityStatement();
-		cs.setStatus(Enumerations.PublicationStatus.ACTIVE);
-		cs.setDateElement(DateTimeType.now());
-		cs.setKind(CapabilityStatement.CapabilityStatementKind.INSTANCE);
-		cs.setFhirVersion(Enumerations.FHIRVersion._4_0_0);
-		return cs;
+		capabilities = new CapabilityStatement();
+		capabilities.setStatus(Enumerations.PublicationStatus.ACTIVE);
+		capabilities.setDateElement(DateTimeType.now());
+		capabilities.setKind(CapabilityStatement.CapabilityStatementKind.INSTANCE);
+		capabilities.setFhirVersion(Enumerations.FHIRVersion._4_0_1);
 		//handleException("An error has occurred while trying to fetch the server's conformance statement", e);
     }
     return capabilities;
--- a/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/client/FHIRToolingClient.java
+++ b/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/client/FHIRToolingClient.java
@@ -108,7 +108,7 @@ public class FHIRToolingClient {
 	public FHIRToolingClient(String baseServiceUrl) throws URISyntaxException {
 		preferredResourceFormat = ResourceFormat.RESOURCE_XML;
 		detectProxy();
-		initialize(baseServiceUrl, "");
+		initialize(baseServiceUrl, null);
 	  }
 	
 	public FHIRToolingClient(String baseServiceUrl, String accessToken) throws URISyntaxException {
--- a/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/client/FHIRToolingClient.java
+++ b/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/client/FHIRToolingClient.java
@@ -96,7 +96,6 @@ public class FHIRToolingClient {
 	
 	private String base;
 	private ResourceAddress resourceAddress;
-	private String accessToken;
 	private ResourceFormat preferredResourceFormat;
 	private int maxResultSetSize = -1;//_count
 	private CapabilityStatement capabilities;
@@ -108,21 +107,16 @@ public class FHIRToolingClient {
 	public FHIRToolingClient(String baseServiceUrl) throws URISyntaxException {
 		preferredResourceFormat = ResourceFormat.RESOURCE_XML;
 		detectProxy();
-		initialize(baseServiceUrl, null);
+		initialize(baseServiceUrl);
 	  }
 	
-	public FHIRToolingClient(String baseServiceUrl, String accessToken) throws URISyntaxException {
-    preferredResourceFormat = ResourceFormat.RESOURCE_XML;
-    detectProxy();
-    initialize(baseServiceUrl, accessToken);
-  }
-  
   public FHIRToolingClient(String baseServiceUrl, String accessToken, String username, String password) throws URISyntaxException {
     preferredResourceFormat = ResourceFormat.RESOURCE_XML;
     utils.setUsername(username);
     utils.setPassword(password);
+	utils.setAccessToken(accessToken);
     detectProxy();
-    initialize(baseServiceUrl, accessToken);
+    initialize(baseServiceUrl);
   }
   
 	public void configureProxy(String proxyHost, int proxyPort) {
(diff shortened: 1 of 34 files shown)


david-simons:
Got it to build - now need to test it against our TX server with an OAuth2 access (bearer) token supplied:
`-txAccessToken [token]: Provide an OAuth2 bearer access token for inclusion in HTTP Authorization request header to the terminology server.`

david-simons:
this part is confusing for me @markiantorno - why is the R5Client being used for R4? (it does not have the `private void setAuth(HttpRequest httpget)` in Client.java (yet)

image

david-simons:
Overall this works!

david-simons:
was able to get the OAuth2 Bearer access token to be passed to TX, from CLI via parameter like-txAccessToken e5f81bb4-4cb8-47a7-9ee4-03b049dacbbb , on that branch/PR.

Now trying to overcome (customizing):

- missing api-version: 1 HTTP header
- missing CapabilityStatement from TX, upon metadata?_summary=true, and metadata?mode=terminology
- java.lang.IllegalStateException: Invalid use of BasicClientConnManager: connection still allocated. Make sure to release the connection before allocating another one in the R4 ClientUtils on the httpclient.execute(request). This is probably related to the required work on the HTTP client in the hl7validator.

david-simons:
as a note to self:
    `Source\Repos\org.hl7.fhir.core\org.hl7.fhir.convertors\src\main\java\org\hl7\fhir\convertors\txClient\TerminologyClientFactory.java
`
```
caseSTU3:returnnew TerminologyClientR3(checkEndsWith("/r3", url), accessToken);
caseR4:return new TerminologyClientR4(checkEndsWith("/r4", url), accessToken);
caseR5:return new TerminologyClientR5(checkEndsWith("/r4", url), accessToken);

```

This is the top-level choice where - depending on the FHIR version context - the appropriate TerminologyClientXX (and all underlying elements) are created.
    
My edits so far were on the DSTU2, STU3, and R4 clients. Only then I noticed that he R5 client has been refactored/improved already. Hence my changes (access token, api-version, conformance statement) are not yet applied to the TerminologyClientR5!

Also note that here I use the TermiologyClientR4, for FHIR R4, not the refactored R5-client.

The `java.lang.IllegalStateException: Invalid use of BasicClientConnManager:` occurs under the R4 case, deep down in the R4ClientUtils on the httpclient.execute(request).

grahamegrieve:
The use of /r4 is a temporary hack until the terminology server starts supporting R5 properly. Since no other terminology server supports R5, this doesn't affect anything else. I'll resolve this shortly. 

I don't understand the api-version thing. Nor the comment about CapabilityStatement

david-simons:
> The use of /r4 is a temporary hack until the terminology server starts supporting R5 properly. Since no other terminology server supports R5, this doesn't affect anything else. I'll resolve this shortly.

thank you for clarifying
 
> I don't understand the api-version thing. Nor the comment about CapabilityStatement
these are specific to the FHIR server we are using, to use as TX server, and connect the HL7Validator to.
It requires api-version: 1 as http header, and needs a workaround currently for the summary and terminology CapabilityStatement calls.
For the pull-request I will remove these. when back in the office ;)

In summary we're stuck on the `java.lang.IllegalStateException: Invalid use of BasicClientConnManager: connection still allocated. Make sure to release the connection before allocating another one in the R4 ClientUtils on the httpclient.execute(request)`. 
Have not been able to debug this enough to find the root cause.

org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/client/FHIRToolingClient.java
```diff
@@ -103,16 +105,18 @@
...
-  
-  public FHIRToolingClient(String baseServiceUrl, String username, String password) throws URISyntaxException {
+
+	public FHIRToolingClient(String baseServiceUrl) throws URISyntaxException {
+		preferredResourceFormat = ResourceFormat.RESOURCE_XML;
+		detectProxy();
+		initialize(baseServiceUrl);
+	  }
+	
+  public FHIRToolingClient(String baseServiceUrl, String accessToken, String username, String password) throws URISyntaxException {
```
markiantorno, line 115: It doesn't appear as though you ever use the username and password with the accessToken. Please create a constructor that takes in just the url and token instead of having to call the constructor over and over with null params. It makes its cleaner.

Also, add javadocs to details what is happening here.

org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/client/FHIRToolingClient.java
```diff
@@ -169,9 +173,18 @@ public void setMaximumRecordCount(int maxResultSetSize) {
 	}
 	
 	public TerminologyCapabilities getTerminologyCapabilities() {
-    return (TerminologyCapabilities) utils.issueGetResourceRequest(resourceAddress.resolveMetadataTxCaps(), getPreferredResourceFormat(), "TerminologyCapabilities", TIMEOUT_NORMAL).getReference();
+  	  try {
+		return (TerminologyCapabilities) utils.issueGetResourceRequest(resourceAddress.resolveMetadataTxCaps(), getPreferredResourceFormat(), "TerminologyCapabilities", TIMEOUT_NORMAL).getReference();
+	  } catch(Exception e) {
+		  // THIS IS A WORKAROUND FOR THE TERMINOLOGY SERVER RESPONDING WITH A 404 TO /metadata
```
markiantorno, line 179: If the call is returning a 404, we need to deal with the 404 and not put code in the `catch` block to execute. This is not good practice.

org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/client/FHIRToolingClient.java
```diff
@@ -188,7 +201,13 @@ public CapabilityStatement getCapabilitiesStatementQuick() throws EFhirClientExc
...
       capabilities = (CapabilityStatement)utils.issueGetResourceRequest(resourceAddress.resolveMetadataUri(true), getPreferredResourceFormat(), "CapabilitiesStatement-Quick", TIMEOUT_NORMAL).getReference();
     } catch(Exception e) {
-      handleException("An error has occurred while trying to fetch the server's conformance statement", e);
+        // THIS IS A WORKAROUND FOR THE TERMINOLOGY SERVER RESPONDING WITH A 404 TO /metadata
+		capabilities = new CapabilityStatement();
+		capabilities.setStatus(Enumerations.PublicationStatus.ACTIVE);
+		capabilities.setDateElement(DateTimeType.now());
+		capabilities.setKind(CapabilityStatement.CapabilityStatementKind.INSTANCE);
+		capabilities.setFhirVersion(Enumerations.FHIRVersion._4_0_1);
+		//handleException("An error has occurred while trying to fetch the server's conformance statement", e);
```
markiantorno, line 210: If the call is returning a 404, we need to deal with the 404 and not put code in the `catch` block to execute. This is not good practice.

org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/tests/NativeHostServiceTester.java
```diff
@@ -13,7 +13,7 @@ public void test() throws Exception {
     
     NativeHostServices svc = new NativeHostServices();
     svc.init("hl7.fhir.r4.core#4.0.1");
-    svc.connectToTxSvc("http://tx.fhir.org/r4", null);
+    svc.connectToTxSvc("http://tx.fhir.org/r4", null, null);
```
markiantorno, line 16: I really don't like this adding fields that we just expect the majority of users to null out. It's not clean at all.

We need to make other method signatures available for people to call if they don't want to pass a token.

markiantorno (changes requested):
A few things will need to change before this can merge. It's better to tackle these now, before you get too much further.
 Please:
1. Add documentation (javadocs) to describe what is going on
2. Update RELEASE_NOTES.md with a description of the test
3. Please add tests for this. We need to verify the changes work, and tests also serve as examples to clarify what is going on. You're making major changes here without any way of verifying they actually do the thing that they are supposed to.
4. Clean up the method calls so that we don't have methods where we're passing 2 or 3 nulls in because we aren't using fields. Do method overloading to solve this if you need to.

Once you've completed that, tag me, and I'll do a thorough check.

grahamegrieve (commented):
I don't like that apiversion header thing. It seems like a road to perdition, with everyone wanting to add support for some custom header. It seems like we need a single way to handle http headers. And the Tx server should just serve up a CapabilityStatement, so I don't see why we should do anything about that

grahamegrieve:
Looking at this, I don't like the overall design at all. What I think we should do is define an interface that provides HTTP headers for the actual request, and pass the interface to the underlying terminology clients. The host can implement the interface how it wants. So I would like to toast this PR, and start with a new PR that first of all defines the interface and sets it up the terminology clients. And then a separate PR for changes to the context. 

I need this functionality, btw, but I need basic auth with username/password, not a bearer token (even though it's actually a bearer token :-( )

david-simons:
> Looking at this, I don't like the overall design at all. What I think we should do is define an interface that provides HTTP headers for the actual request, and pass the interface to the underlying terminology clients. The host can implement the interface how it wants. So I would like to toast this PR, and start with a new PR that first of all defines the interface and sets it up the terminology clients. And then a separate PR for changes to the context.
> 
> I need this functionality, btw, but I need basic auth with username/password, not a bearer token (even though it's actually a bearer token :-( )

Thanks @grahamegrieve and @markiantorno for your feedback - in line with my thinking - let me briefly respond while still ouf-of-office:

Note that this PR is in Draft at this time - since for me it is a proof-of-concept, and to determine requirements/solution direction,  on both the HL7 Validator, as well as to the FHIR TX Server that we use - to ensure we can connect them. 
More development needs to happen indeed subsequently, and ideally the approach can be generalized.

For this PoC, I pretty much replicated what is already done on the master for the BasicAuth username/password parameters (for proxy auth?). Adding multiple optional string parameters is not the way too go (and hard to do with method overloading).

I was also about to propose replacing the existing username/password proxy parameters with a generic `-httpHeaders <file>` parameter, allowing anyone to ingest one or more custom HTTP parameters, be it basic auth, be it an bearer auth, be it an api-version, etc. 
These would be in the content of the `<file>`, on separate lines.
```
Authorication: Bearer TOKEN
Proxy-Authorization: Basic USER:PWD
api-version: 1
etc.

```

Totally fine if that is done on a separate branch and corresponding new PR.

markiantorno:
We're going to address this as part of the solution to https://github.com/hapifhir/org.hl7.fhir.core/issues/588.

I'll be opening a pull request with the changes for this.

Issue #557, opened 2021-07-09
FHIRPathEngine assumes that all extension elements declare a type


jduteau:
Line 5500 in FHIRPathEngine.java:
StructureDefinition exsd = worker.fetchResource(StructureDefinition.class, t.getType().get(0).getProfile().get(0).getValue());

I have a differential that doesn't declare the type in one of the extension element:
```
      {
        "id": "Claim.careTeam",
        "path": "Claim.careTeam",
        "slicing": {
          "discriminator": [
            {
              "type": "value",
              "path": "extension('http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-careTeamClaimScope').value"
            }
          ],
          "rules": "open",
          "description": "Slice based on whether the care team member belongs to the overall claim or to an individual claim item."
        },
        "mustSupport": true
      },
      {
        "id": "Claim.careTeam.extension",
        "path": "Claim.careTeam.extension",
        "slicing": {
          "discriminator": [
            {
              "type": "value",
              "path": "url"
            }
          ],
          "ordered": false,
          "rules": "open"
        },
        "min": 1
      },
      {
        "id": "Claim.careTeam.extension:careTeamClaimScope",
        "path": "Claim.careTeam.extension",
        "sliceName": "careTeamClaimScope",
        "min": 1,
        "max": "1",
        "type": [
          {
            "code": "Extension",
            "profile": [
              "http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-careTeamClaimScope"
            ]
          }
        ],
        "mustSupport": true
      },
...
      {
        "id": "Claim.careTeam:OverallClaimMember",
        "path": "Claim.careTeam",
        "sliceName": "OverallClaimMember",
        "min": 0,
        "max": "14"
      },
      {
        "id": "Claim.careTeam:OverallClaimMember.extension:careTeamClaimScope",
        "path": "Claim.careTeam.extension",
        "sliceName": "careTeamClaimScope"
      },
...
      {
        "id": "Claim.careTeam:ItemClaimMember",
        "path": "Claim.careTeam",
        "sliceName": "ItemClaimMember",
        "min": 0,
        "max": "10"
      },
      {
        "id": "Claim.careTeam:ItemClaimMember.extension:careTeamClaimScope",
        "path": "Claim.careTeam.extension",
        "sliceName": "careTeamClaimScope"
      },
```

When I run the publisher, that gets the following Java uncaught exception:
```
[...]

PR #571, opened 2021-08-09
FHIRPathEngine assumes that all extension elements declare a type

markiantorno:
https://github.com/hapifhir/org.hl7.fhir.core/issues/557


grahamegrieve (approved):
ok. But I do wonder whether there's an underlying issue somewhere - the snapshot doesn't have a type? why not? This feels like a workaround for that issue, and it should be investigated. The snapshot should always have a type

Issue #555, opened 2021-07-05
Handling minValue/maxValue in generate snapshot process


LouisRolland78:
**Describe the bug**
The fields minValue/maxValue are ignored when generating a snapshot from a differential (class ProfileUtilities, method updateFromDefinition).

For example for the field maxLength, we have this piece of code:
![image](https://user-images.githubusercontent.com/28785069/124455194-92516d00-dd89-11eb-9d65-0fd799ac3d29.png)

I think that we should have the same type of code for minValue and maxValue.

@grahamegrieve : Could you help me with this bug please? What is the Issues/Pull requests process?

Thank you.

LouisRolland78:
I tried to push the fix but I have 403 access denied. The piece of code to add is: (in all ProfileUtilities, just after the piece of code on derived.hasMaxLengthElement() (see image))


      if (derived.hasMaxValue()) {
        if (!Base.compareDeep(derived.getMaxValue(), base.getMaxValue(), false))
          base.setMaxValue(derived.getMaxValue().copy());
        else if (trimDifferential)
          derived.setMaxValue(null);
        else
          derived.getMaxValue().setUserData(DERIVATION_EQUALS, true);
      }
  
      if (derived.hasMinValue()) {
        if (!Base.compareDeep(derived.getMinValue(), base.getMinValue(), false))
          base.setMinValue(derived.getMinValue().copy());
        else if (trimDifferential)
          derived.setMinValue(null);
        else
          derived.getMinValue().setUserData(DERIVATION_EQUALS, true);
      }

markiantorno:
Please fork the repo into your own copy. Make the changes on the fork, and then open a PR from your fork to the main repo. 

I will review once you do this.

LouisRolland78:
Ok Thank you! I'm doing that.

PR #574, opened 2021-08-11
fix: #555 Handling minValue/maxValue in generate snapshot process

LouisRolland78:


markiantorno (changes requested):
@LouisRolland78 You need to add tests please. There's no way of verifying your changes work, or what they do without tests.

Also, please update the RELEASE_NOTES.md file with the change details.

Please add them, and then I will review.

grahamegrieve (commented):
You only need to add a test for R5

LouisRolland78:
I add two commits, one for unit tests ( I have add 2 tests and uncomment 1 test about cardonality, I have also removed the part with xml comparison because it is necessary to use the new XmlParser and I have no time for this refacto) and one for release notes. I hope it will be ok for you.
Sorry, i don't know how to update my pr with your current master branch to avoid conflict.

PR #578, opened 2021-08-17
Rendering enhancements

lmckenzi:
Changes to support new template enhancements, including:
- support for new extensions to manage FMM and standard status propagation
- ensure that codes returned from value set lookups and expansions include display names
- rendering support for codes, codings, codeable concepts and quantities

### DIFF
--- a/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
+++ b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
@@ -99,8 +99,8 @@ public class FilesystemPackageCacheManager extends BasePackageCacheManager imple
   private boolean progress = true;
   private List<NpmPackage> temporaryPackages = new ArrayList<>();
   private boolean buildLoaded = false;
-  private Map<String, String> ciList = new HashMap<String, String>();
   private JsonArray buildInfo;
+  private HashMap<String,String> packageRepos;
 
   /**
    * Constructor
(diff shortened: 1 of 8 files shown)


org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
```diff
@@ -521,26 +521,20 @@ private InputStream fetchFromUrlSpecific(String source, boolean optional) throws
 
   private InputStreamWithSrc loadFromCIBuild(String id) throws IOException {
     checkBuildLoaded();
-    if (ciList.containsKey(id)) {
-      InputStream stream = fetchFromUrlSpecific(Utilities.pathURL(ciList.get(id), "package.tgz"), false);
-      return new InputStreamWithSrc(stream, Utilities.pathURL(ciList.get(id), "package.tgz"), "current");
+    if (packageRepos.containsKey(id)) {
+      InputStream stream = fetchFromUrlSpecific(Utilities.pathURL((String)packageRepos.get(id), "package.tgz"), false);
```
markiantorno: Does this need a cast here?

markiantorno (changes requested):
You have two PRs open right now, and they appear to have overlap?

Can you either separate the changes so that they're in their own PRs or, just close one and have all the changes in the other please.

grahamegrieve (commented):
There's a lot going on here. Why the changes to ValidationResult? Why add rendering stuff to the context, when it belongs in the rendering classes? What's going on with the additional extensions?

lmckenzi:
Changes to ValidationResult is because when you do a code validation, the ValidationResult has long been supposed to return the Coding that was found to be valid.  However, it wasn't doing that consistently.  The changes are to ensure that it *does* do it consistently.

This rendering stuff isn't for narrative, it's to return information about the jurisdiction, usageContext, keywords and other metadata in a Conformance resource

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
lmckenzi: Nope.  Originally I hadn't properly typed the Hashmap.  Now that I have, these are redundant.  Updated and committed.

grahamegrieve:
I'm not going to approve this PR. We're not going to make these changes to IWorkerContext - they belong in the rendering classes (and overlap with code in them). Need to discuss these changes elsewhere, and figure out where all this stuff goes.

lmckenzi:
These changes have already been discussed in terms of "what should happen" in both Zulip and the IG authoring project.  We can certainly discuss where the code needs to go.  The key thing is that from within the code that spits out the JSON files for resources, we can grab the human-readable value for a CodeableConcept which will find the coding that corresponds to a value set and look up the display name if necessary and return it if there's no CodeableConcept.text present.  I wasn't aware the rendering classes handled terminology lookup?

Where do you want the discussion to take place?

grahamegrieve:
first replacement, https://github.com/hapifhir/org.hl7.fhir.core/pull/586

PR #584, opened 2021-08-23
Replace proprietary notnull with javax.annotation.Nonnull

tadgh:
* Prevents build issues upstream with projects that try to not rely on IDE-specific annotations.


grahamegrieve (commented):
if you made these changes without any others, why would we think that these are even used? these look like bad insertions by the Eclipse IDE

markiantorno:
The non-null is fine, we just need to not use the proprietary jetbrains ones and use the javax ones. The issue was that the dependency is being passed downstream and causing issues. Let me re-run the build and see if I can get it to pass.

Issue #513, opened 2021-05-25
ids aren't being properly checked when validating


lmckenzi:
The publisher was inserting ids that looked like this, but the validator never complained that they were invalid:
    "grouping": [
      {
        "id": "_dyn_capabilitystatement",
        "name": "Behavior: Capability Statements",
        "description": "The following artifacts define the specific capabilities that different types of systems are expected to have in order to comply with this implementation guide.  Systems conforming to this implementation guide are expected to declare conformance to one or more of the following capability statements."
      },

(note that underscores aren't a valid id character)

jmandel:
The validator also currently accepts `id`s > 1M chars, and >64 chars.

arscan:
This is also likely related to https://github.com/hapifhir/org.hl7.fhir.core/issues/476

PR #587, opened 2021-08-24
Issue 513

markiantorno:
ids aren't being properly checked when validating 

https://github.com/hapifhir/org.hl7.fhir.core/issues/513


grahamegrieve (approved):
This should have test cases in the validator tests to make sure other validators validate correctly

PR #596, opened 2021-09-03
Gg 202109 gg expand fix

grahamegrieve:


grahamegrieve:
@markiantorno please review. Particularly the package changes - we've talked about that before

PR #599, opened 2021-09-08
Terminology client headers

markiantorno:
@grahamegrieve please review this. If you're okay with how it is setup, I'm just going to copy it over to dstu2-r4 as well. Right now this is only for r5.


grahamegrieve (commented):
why is there multiple identical implementations of ClientHeaders?

grahamegrieve (commented):
And why are the headers only actually implemented for R5?

markiantorno:
> And why are the headers only actually implemented for R5?

I am unsure why the changes didn't get added (I must have run git add from a sub directory and not from the convertors project), but they are pushed up now

markiantorno:
> why is there multiple identical implementations of ClientHeaders?

I followed the existing convention where the network code, while similar between versions, was just copied from version to version. I can pull it out and put it in utilities in another PR if you'd like, but I'd like to get this one in as it is now, if possible.

grahamegrieve (approved):
ok

grahamegrieve:
I think it would be good to rationalise the code so code that only needs to be in one place is only in one place

markiantorno:
> I think it would be good to rationalise the code so code that only needs to be in one place is only in one place

Oh god yes. I think 95% of the networking code should be in it's own package and just return IBaseResources.

Issue #611, opened 2021-09-29
List profile url is incorrect.


markiantorno:
In r4, ListResource has a profile url of `profile="http://hl7.org/fhir/StructureDefinition/ListResource"`, however, touchstone validation throws an warning when validating the server metadata, if the ListResource profile url is not `profile="http://hl7.org/fhir/StructureDefinition/List"`.

Looking at R5, it appears this issue was resolved and the correct url was used.

Can we switch the R4 ListResource `@ResourceDef` tag to use the proper url.

markiantorno:
@katiesmilecdr once the next release of HAPI/Smile goes out, please confirm this solves the issue.

tanvishah9696:
Generally HL7 version bump updates only occur at a certain release time but this issue needs an update to be moved earlier.

PR #612, opened 2021-09-29
update ListResource profile in r4

katiesmilecdr:


markiantorno:
https://github.com/hapifhir/org.hl7.fhir.core/issues/611

PR #615, opened 2021-10-04
Fix GraphQLSchemaGenerator to produce valid GraphQL schemas

lschmierer:
The schemas generated by the R4 and R5 (https://build.fhir.org/fhir.schema.graphql.zip) `GraphQLSchemaGenerator`s are no valid GraphQL schemas.

This PR contains the following fixes:

- The `scalar` keyword generated for parameter types used to be capitalized (e.g. `Scalar number` instead of `scalar number`). Only `scalar` is a valid keyword in GraphQL.
- Two scalars with the name `date` were generated (one for primitives and one for parameters). As both use a JSON representation of type string, this fix replaces the second scalar with the comment `# Search Param date: already defined as Primitive with JSON Format: string `
- Various input types had properties of type `ElementBase` which is not an input type. Input types can only use other input types. A new input type `ElementBaseInput` was added. `ElementBase` was changed to `ElementBaseInput` for all input types.

As well as some minor enhancements:

- All scalars were already followed by a comment specifying the actual JSON type. However, for most of them a placeholder `??` was printed. This PR changes `??` to the JSON type as specified in https://www.hl7.org/fhir/datatypes.html.
  (e.g. `scalar date # JSON Format: string` instead of `scalar date # JSON Format: ??`)
- The comment `# import the types from 'types.graphql'` was changed to `# import * from 'types.graphql'`. This makes it easier to work with the schema using *graphql-tools* which is a very common GraphQL library in the NodeJS ecosystem (https://www.graphql-tools.com/docs/schema-loading). 

I am actually not sure how to really test this. There does not seem to be any GraphQLSchemaGenerator related tests in this repo. I tried building https://github.com/HL7/fhir with my local changes to https://github.com/hapifhir/org.hl7.fhir.core but were not succesful yet.
However, since the generated schemas were no valid GraphQL anyway, there is at least no way to break anything.

### DIFF
--- a/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/GraphQLSchemaGenerator.java
+++ b/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/GraphQLSchemaGenerator.java
@@ -43,6 +43,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


grahamegrieve (approved):
ok

PR #630, opened 2021-10-22
Validate MeasureReport groups only if Measure has groups

rob-reynolds:
Ref Zulip https://chat.fhir.org/#narrow/stream/179165-committers/topic/MeasureReport.2Egroup.20validation.

Tested locally that it addresses the issue.

Tested locally that issues that were detected prior to the update still are.
Pre update:
image

Post update:
image


grahamegrieve (approved):
ok

PR #637, opened 2021-11-02
Gg 202110 CVE 2021 42574

grahamegrieve:


grahamegrieve:
@markiantorno review please, and think this through - have I got the essence of the CVE covered, and do I need more test cases?

PR #639, opened 2021-11-02
changing JAVA_HOME to remove trailing slash

markiantorno:
Utilities test fails Java Home location because of trailing slash.


grahamegrieve (approved):
ok

PR #640, opened 2021-11-03
SHCParser file cleanup

dotasek:
On *nix systems, "c:\\temp\\payload.json" was being generated in the org.hl7.fhir.validation project when tests were being run. This led to accidental commits and difficulties in pulling code to Windows systems, which broke pipeline builds.

This replaces the file generation with java File.createTempFile and deletes explicitly. 

@grahamegrieve it doesn't actually look like the written file is read anywhere. Removing the line`JsonTrackingParser.write(jwt.payload, "c:\\temp\\payload.json");` doesn't appear to affect any tests included in the build either. Is this perhaps a leftover from a previous implementation? Could the output file writing be removed altogether?


grahamegrieve (commented):
accidental oops from me - purely a debugging thing I should have removed. So just remove it altogether (anything you ever see in c:\temp is pure debugging on my part and should never be committed)

markiantorno (approved):

PR #643, opened 2021-11-04
fix missing logging of requests

grahamegrieve:


markiantorno (changes requested):
I will do this with an interceptor.

grahamegrieve:
so I should merge this PR? or not?

Issue #644, opened 2021-11-07
Resolve references on disk


pieter-edelman-nictiz:
When validating a file on disk using the HL7 Validator, it's not possible to resolve a relative reference. However, sometimes it is required for proper validation to consider the referenced resource. I'm not sure if a "proper" mechanism exists, but it would seem logical to make the Validator treat a relative reference in the form of "ResourceType/id" as a relative reference _on disk_, so it would search for a file that represents this resource type and id.

Some considerations:
* This behavior could be activated using a switch command line switch.
* There could be rules regarding file names of referenced resources, like naming them according to the id or combination of resource type and id.
* Referenced resources could be required to be in the same folder, or they some command line option could be used to set the folder of referenced resources.

pieter-edelman-nictiz:
Just opened a draft PR as a demo of a simple approach.

grahamegrieve:
is this a good feature, is it? How would it work from a user point of view?

pieter-edelman-nictiz:
Suppose I have a file representing an Observation resource that I want to validate using the Validator. This Observation needs to reference a Patient resource (which also only exists as a file). It would seem logical if in the Observation I could just use:
```
<subject>
  <reference value="Patient/Everyman01"/>
  <display value="Adam Everyman"/>
</subject>
```

And that the Validator then would look for a file in the same directory called "Patient-Everyman01.xml" or "Everyman01.xml" and try to read them as the Patient resource. This would align with the way relative references work on a server.

grahamegrieve:
discussion amongst the committers: 

* we won't enable this by default (may have unexpected side effects in existing deployments)
* whatever we do, has to work for validator.fhir.org as well as the CLI validator
* there's lots of patterns that people use. For a survey see line 204 of org.hl7.fhir.igtools.publisher.SimpleFetcher
* so you'd have to provide a pattern, something like a parameter -resolve={type}/{id}.{fmt} which operates from the current directory

pieter-edelman-nictiz:
(link to line 204: https://github.com/HL7/fhir-ig-publisher/blob/be10fd8c5e22415fa5446d4757df905962ea947d/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/SimpleFetcher.java#L204)

Seeing the variety of file naming approaches, providing a pattern seems like a good approach. However, maybe there could be simpler alternatives:

[...]

PR #646, opened 2021-11-07
Issue #644

pieter-edelman-nictiz:
First attempt at resolving local references for files on disk.

The strategy is to simply look for files in the same folder where the file name matches the id (possibly combined with the resource type) and then check if id and resource type actully match.
Error handling, caching or documentation are still missing.


pieter-edelman-nictiz:
Related to issue #644

PR #653, opened 2021-11-10
Update maven config and tests to allow Java 8, 11 and 17 builds

dotasek:
* Updates several dependency versions.
* Conditionally sets up surefire tests with an optional argument that allows Java 8 surefire tests to run without encountering garbage collection errors.
* Updates the JAVA_HOME resolution for Mac in UtilityTests to use the resolved path instead of a string, which was failing on different Mac setups due to inconsistencies in using a trailing forward slash for the value.


dotasek:
Work in progress; will be adding Azure pipeline targeting this branch to test.

dotasek:
No longer Work in Progress; ready for review.

markiantorno (approved):
Good stuff.

PR #662, opened 2021-11-16
Gg 2021 misc issues

grahamegrieve:


pom.xml
```diff
@@ -19,7 +19,7 @@
 
     <properties>
         <hapi_fhir_version>5.1.0</hapi_fhir_version>
-        <validator_test_case_version>1.1.74-SNAPSHOT</validator_test_case_version>
+        <validator_test_case_version>1.1.73</validator_test_case_version>
```
dotasek, line 22: @grahamegrieve is this intentional? If I keep it as 1.1.74-SNAPSHOT, my local build works correctly; with this value it fails locally and in the pipeline.

grahamegrieve: well, it was intentional, but it is also wrong. I'll do a test case release and then update it to 1.1.74

grahamegrieve:
this was superceded by the next PR

PR #673, opened 2021-11-23
Add additional output for size tests

dotasek:
This test resulted in an intermittent failure last week. As the test uses randomly generated values and provided no output, we couldn't see the explicit cause for the failure.

This adds output for the test cases so next time we will at least get actionable feedback.


markiantorno (changes requested):
Please update RELEASE_NOTES.md with the change.

markiantorno (approved):

PR #683, opened 2021-12-06
Fix JAVA_HOME tests with trailing slashes

dotasek:


markiantorno (changes requested):
RELEASE_NOTES.md please

PR #690, opened 2021-12-16
fix for escaping json control characters in graphql

reasonableJP:
double escape, once for java and once for json

fixes https://github.com/hapifhir/hapi-fhir/issues/3069


grahamegrieve (approved):
ok

PR #704, opened 2022-01-03
Add map compile

lmckenzi:
Add ability to the validator to be able to convert a FHIR mapping language text file to an XML, JSON or RDF StructureMap instance.

Corresponding updates to the Using FHIR Mapping Language confluence page are available in 'draft' (not yet published).


org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/cli/services/ValidationService.java
```diff
@@ -253,6 +253,37 @@ public void transform(CliContext cliContext, ValidationEngine validator) throws
...
+            validator.getContext().generateSnapshot(sd, true);
+          } else {
+            validator.getContext().generateSnapshot(sd, false);
+          }
+        }
+      }
+      validator.setMapLog(cliContext.getMapLog());
+      cliContext.getMap();
+      StructureMap map = validator.compile(cliContext.getMap());
+      System.out.println(" ...success");
```
jvitrifork, line 279: Please don't use System.out ... it clutters the terminal

lmckenzi: I was just being consistent with existing behavior - which already uses System.out elsewhere.

dotasek: @jvitrifork is correct, but @lmckenzi is being consistent. Unfortunately System.out is used throughout to provide output for the command line interface.

dotasek (commented):
Is this PR in progress?

There are a few areas where I would like to see some work.

I've tested this by running it in what I believe is the correct manner:

```
java -jar org.hl7.fhir.validation.cli-5.6.22-SNAPSHOT.jar -compile http://hl7.org/fhir/StructureMap/supplyrequest-transform
```

I get the `...success` output, which is expected, but no other details. I can trace it in IntelliJ and see that `map` is populated by a valid StructureMap, so that works as I think it was intended.

That would be OK, except that when I execute it with parameters that don't make any sense, I also get `...success`.

```
java -jar org.hl7.fhir.validation.cli-5.6.22-SNAPSHOT.jar -compile garbagegarbagegarbage
```

The compile runs fine because the URL isn't found when context.getTransform(mapUri); is called, so `map` is populated by `null` and success is reported.

What would be an example URL to pass for a truly invalid StructureMap? Initially, I passed a local file full of garbage, but I got bitten by the 'gotcha' above.

lmckenzi:
@dotasek, I hadn't tested well for non-happy-path invocation.  I've modified the code to tighten things up.  It should now yell appropriately if you specify an invalid map, pass in a map-file it can't find, specify an invalid output location or fail to specify an output location.

dotasek:
@lmckenzi That's a definite improvement.

I can now run this and get a meaningful complaint:

```
java -jar org.hl7.fhir.validation.cli-5.6.24-SNAPSHOT.jar -compile garbagegarbagegarbage  -output /Users/david.otasek/IN/2022-01-07-PR-704-review/output.xml -version 4.0.1
```

And running this gives me what I am assuming is a legitimate output:

```
java -jar org.hl7.fhir.validation.cli-5.6.24-SNAPSHOT.jar -compile http://hl7.org/fhir/StructureMap/supplyrequest-transform -output /Users/david.otasek/IN/2022-01-07-PR-704-review/output.xml -version 4.0.1
```

dotasek:
@lmckenzi This branch needs a merge from master to pass the checks. I did one here that appears to pass on a local build, plus has some minor fixes I stumbled on while testing: https://github.com/lmckenzi/org.hl7.fhir.core/pull/1

lmckenzi:
Ok, merged.

PR #706, opened 2022-01-05
Changes relevant to improving the rendering of artifact intros.  Spec…

lmckenzi:
…ifically:

- ensure that when validating codes against value sets, the valid coding is actually returned
- strip out the new extensions we add for FMM
- improve the rendering of Quantity and Range.  (The former was horrid and the latter wasn't as nice as it could be.)

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/ProfileUtilities.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/ProfileUtilities.java
@@ -2087,6 +2087,8 @@ public class ProfileUtilities extends TranslatingUtilities {
 
   private void removeStatusExtensions(ElementDefinition outcome) {
     outcome.removeExtension(ToolingExtensions.EXT_FMM_LEVEL);
+    outcome.removeExtension(ToolingExtensions.EXT_FMM_SUPPORT);
+    outcome.removeExtension(ToolingExtensions.EXT_FMM_DERIVED);
     outcome.removeExtension(ToolingExtensions.EXT_STANDARDS_STATUS);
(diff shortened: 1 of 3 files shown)


grahamegrieve (commented):
what's going on with \u00A0?

lmckenzi:
Non-breaking space to prevent inappropriate wrapping.

grahamegrieve:
why follow the non-breaking space with a breaking space?

grahamegrieve (commented):
and I'm not happy with the changes to displayQuantity. Why? How is that related to artifact intros?

lmckenzi:
The space beside the non-breaking is an error.  Will fix.  The rendering of the Quantity is horrible.  It's not showing the value and it is showing the system (and showing a bunch of junk if there is no system).  For human display of a quantity you want the value, then the human-readable unit if there is one, or the code if there isn't a human-readable unit.  The code system is generally noise.  If you want to optionally add it as a 'detail' if the rendering asks for details and show it only if it actually exists and isn't null, then that could be ok.  The header renders all of the Resource.useContext information and some of that is quantities and ranges.

grahamegrieve:
" The code system is generally noise" - not if it's not ucum

lmckenzi:
Depends on the system.  In any case, having it be the first thing you see (and complaint that it doesn't exist if there isn't one) is absolutely not correct, as is not showing the value of the quantity at all.  What I could do is include the code system *if* the code is displayed (because there's no unit) and if the code system isn't UCUM.

E.g. `10 (unit=1234 from system abc)`
On the other hand, if there is a unit specified or a UCUM code, then it would just say `10 mg`

Would that be reasonable?

grahamegrieve:
yes. no need to display code system if it's ucum.

PR #713, opened 2022-01-14
Fix literal inclusion of Java constants in static text file in cli help

dotasek:
The help.txt file written to System.out contained Java constants that were written out literally. This change replaces them with placeholders that can be specified in Display.java


markiantorno (approved):
Please add RELEASE.md details, but good otherwise.

PR #714, opened 2022-01-14
Avoid unnecessary JsonParser/XmlParser class load

jamesagnew:
The R5 format JsonParser/XmlParser classes (not the elementmodel ones, the FHIR model ones) are massive classes at this point, to the point that they break code coverage tools..

Currently they get loaded if you use the InstanceValidator because that class uses `FormatUtilities` which in turn imports JsonParser and XmlParser even if they never get used for that use case. That appears to be the cause of this issue: https://github.com/hapifhir/hapi-fhir/issues/3268

This PR changes to using reflection for instantiation of these classes, which should avoid this issue.


jamesagnew:
Incidentally, I don't know if any other tooling (e.g. the IG Publisher) ever calls FormatUtilties#makeParser? If not, another solution to this issue would be to move these methods to another class, e.g. maybe as a static mathod on BaseParser?

PR #715, opened 2022-01-15
Use version utilities to resolve version for handleOutputToStream

dotasek:
This stems from executing a -compile option without an explicit version. "current" is selected by default, resulting in an exception.

In this case, current should resolve correctly.


markiantorno (approved):
Add RELEASE_NOTES.md, but fine otherwise.

PR #720, opened 2022-01-24
fixed detection of FHIR R5 from version string

glichtner:
The current version of the utilities does not correctly detect FHIR R5 (5.0.0-snapshot1) from version strings when running the validator using "-version 5.0.0" or "-version 5.0.0-snapshot1" parameter.

Although the appropriate functions to check for R5 were already implemented, they were not called in VersionUtilities.packageForVersion() and VersionUtilities.getCurrentVersion().
This PR adds the calls to these functions.


grahamegrieve (approved):
ok

PR #721, opened 2022-01-25
TX Server Caching for Tests

dotasek:
These changes do the following:

* Initialize validation engines and contexts in tests to use the included txCaches in the checked-in code
* Allow caching of errors in the case of testing
* Cache terminology capabilities to avoid hitting the server on initialization.
* JUnit tests for TerminologyCache and SimpleWorkerContext to avoid regression.


grahamegrieve (approved):
looks good

PR #727, opened 2022-01-30
Updates to GraphQL

jamesagnew:
This PR fixes a couple of issues with the GraphQL modules:

* The GraphQL engines for each version of FHIR did not fully account for `choice[x]` elements. Essentially a query like `{ObservationList() { effectiveDateTime }}` would fail if any observation was present with a `Period` value in `Observation.effective[x]`.
* The schema generator stored state in the StructureDefinition UserData and did not clean it up, meaning that if the same StructureDefinition object was passed to the schema generator more than once, the second time would result in an incomplete schema being generated.
* The schema generator generated two scalar definitions named `Uri` because of the datatype name conflicting with the search parameter name.
* The schema generator can now accept a `Writer` or an `OutputStream` as targets (previously only `OutputStream` was supported).
* The shcema generator uses the specified operations to determine whether to generate input or output types (or both) for both resources and types, in order to reduce the size of the schema


grahamegrieve (approved):
thx

PR #730, opened 2022-02-02
Refactor to use builder pattern in SimpleWorkerContext and ValidationEngine

dotasek:
This change primarily enforces the order of cache initialization and terminology client initialization, and is a step towards reducing the number of constructors required of both classes.


grahamegrieve (approved):
ok

PR #740, opened 2022-02-11
Ignore META-INF while unpacking dependencies

dotasek:
The unpack dependencies step in validator.cli appears to dump META-INF contents from every dependency into the root META-INF in target/classes. The resulting contents can then be misconstrued as those intented for validator.cli, and cause errors down the road.


grahamegrieve (approved):
ok

PR #750, opened 2022-02-24
Fix include resolution in LiquidRenderer

brynrhodes:
Include resolution doesn't work in the LiquidRenderer because the LiquidEngine's includeResolver isn't set. This PR adds a private imlementation of the IncludeResolver interface that uses the context's LiquidTemplateProvider.


grahamegrieve (approved):
ok

PR #751, opened 2022-02-25
Fix FHIRPathEngine funcReplace parameters

dotasek:
FHIRPathEngine funcReplace requires two parameters, but incorrectly throws an exception when two parameters are provided.


grahamegrieve (approved):
ok

PR #756, opened 2022-03-03
Fix intermittent failures in NarrativeGenerationTests

dotasek:
The WorkerContext used in the VocabTests was adding terminology that was then used by other tests, since the instance would persist in TestingUtilities. This primarily impacted NarrativeGenerationTests, causing intermittent failures if VocabTests was run prior to NarrativeGenerationTests.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,4 +4,4 @@
 
 ## Other code changes
 
-* no changes
+* Fix to r5 TestingUtilities to allow NarrativeGenerationTests to pass. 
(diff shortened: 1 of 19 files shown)


grahamegrieve (commented):
I don't understand this - what actually changed the fixes the tests?

dotasek:
@grahamegrieve

TestingUtilities keeps a map of IWorkerContext instances that can be re-used across multiple tests. I believe this is so that we can save time by not instantiating a new worker context for every test.

One of these tests, VocabTests, uses one of these contexts and loads additional terminology to it. The additional terminology contains SNOMET_CT international, which has the same URL and ID as regular SNOMET_CT. If VocabTests was run before NarrativeGenerationTests (test classes are ordered randomly), the context would contain SNOMED_CT international. NarrativeGenerationTests expects the (All versions) version of SNOMET_CT. This was the cause of our test failures.

The fix I implemented ensures that VocabTests uses its own instance of IWorkerContext, which won't be shared by other tests. We do a similar thing already for a few other tests by using TestingUtilities.getWorkerContext()

grahamegrieve (approved):
ok

grahamegrieve:
I didn't find where the alternative source was btw - where do the two sources come from?  I suspect there's a deeper underlying issue there

PR #758, opened 2022-03-04
Fix type conversion issue in converters.txClient.TerminologyClientR3

lmckenzi:
Hit this bug when handling resource expansions for an STU3 IG.  This code would never have worked, so apparently nothing called it before now.


grahamegrieve (commented):
what kind of testing did you do for this?

lmckenzi:
The original code blew up with a type conversion error when it ran.  Correcting the type declaration made the code run.  This was a clear typo in the original code - the converter takes in a Parameters and passes out a Parameters, so casting it to something else can't possibly work.  Plus, what's needed later in the code is a Parameters.

grahamegrieve (approved):
ok

PR #759, opened 2022-03-05
R4 b datatypes and xml cleanup

lmckenzi:
Cleaned up XML escaping, fixed problem with case on FHIR version and ensure new R4B data types don't show up in extensions, parameters, etc.


grahamegrieve (commented):
why these changes? (other than the xml one)

lmckenzi:
The 'toLowerCase()' on the version is because the enumerations are lower-case, but the ini value is all upper-case.  Without this change, the code does not run in Eclipse

Excluding CodeableReference and RatioRange from the extension types and other types is an approved change request (listed in the commit)

grahamegrieve:
I don't know why the case thing is a problem. I think your code is out of sync somewhere

lmckenzi:
I've traced it through.  The ini file definitely has upper-case.  The code definitely is looking for lower-case.  I'm current against both code bases.  (And this isn't a new problem - it's been consistent over the course of a month, which is when I first started messing with R4B.)

PR #767, opened 2022-03-15
Use getValueAsString for DateTimeType conversions

dotasek:
This changes all BaseDateTimeType extension conversions to use getValueAsString instead of getValue in order to preserve precision when converting.


grahamegrieve (approved):
ok looks good

Issue #771, opened 2022-03-21
Exception conversion R5 to R4 of StructureMap


oliveregger:
Converting a StructureMap Resource R5 to R4 returns the following exception:

```
Type mismatch: the type StringType was expected, but org.hl7.fhir.r5.model.IdType was encountered
	at org.hl7.fhir.r5.model.StructureMap$StructureMapGroupRuleTargetParameterComponent.getValueStringType([StructureMap.java:4480](http://structuremap.java:4480/))
	at org.hl7.fhir.convertors.conv40_50.resources40_50.StructureMap40_50.convertStructureMapGroupRuleDependentComponent(StructureMap40_[50.java:793](http://50.java:793/))
	at org.hl7.fhir.convertors.conv40_50.resources40_50.StructureMap40_50.convertStructureMapGroupRuleComponent(StructureMap40_[50.java:406](http://50.java:406/))
	at org.hl7.fhir.convertors.conv40_50.resources40_50.StructureMap40_50.convertStructureMapGroupComponent(StructureMap40_[50.java:256](http://50.java:256/))
	at org.hl7.fhir.convertors.conv40_50.resources40_50.StructureMap40_50.convertStructureMap(StructureMap40_[50.java:131](http://50.java:131/))
	at org.hl7.fhir.convertors.conv40_50.resources40_50.Resource40_50.convertResource(Resource40_[50.java:481](http://50.java:481/))
	at org.hl7.fhir.convertors.conv40_50.VersionConvertor_40_50.convertResource(VersionConvertor_40_[50.java:82](http://50.java:82/))
	at org.hl7.fhir.convertors.factory.VersionConvertorFactory_40_50.convertResource(VersionConvertorFactory_40_[50.java:24](http://50.java:24/))
	at org.hl7.fhir.convertors.factory.VersionConvertorFactory_40_50.convertResource(VersionConvertorFactory_40_[50.java:19](http://50.java:19/))
```

This is happening due to that in R5 https://build.fhir.org/structuremap.html#StructureMap.group.rule.target.parameter
dependent.parameter: parameter is a BackbonElement with Parameter as value[x]	
Σ	1..1		Parameter value - variable or literal
........ valueId			id	
........ valueString			string	
........ valueBoolean			boolean	
........ valueInteger			integer	
........ valueDecimal			decimal	
........ valueDate			date	
........ valueTime			time	
........ valueDateTime			dateTime

[...]

PR #772, opened 2022-03-21
fix exception R5 to R4 of StructureMap #771

oliveregger:
see issue #771 corrects transformation and adds a test case that transform is checked for this specific case


dotasek:
@oliveregger Thank you for the test case.

There are some details regarding IdType to StringType conversion that I'd like to add some additional code/tests for. I will create my own branch for comparison and notify here.

dotasek:
I opened up another draft PR to expand on what I found: https://github.com/hapifhir/org.hl7.fhir.core/pull/775

dotasek:
Changes to https://github.com/hapifhir/org.hl7.fhir.core/pull/775 are complete; I believe this should resolve the issue.

PR #773, opened 2022-03-22
Update BaseWorkerContext.java HashMap to TreeMap to avoid null keys

sanjan2:
HashMap accepts null keys, TreeMap doesn't. The map allResourcesById is used to store FHIR resources with unique ID keys (their unique URIs), a null key must therefore be wrong. 
Currently the HashMap masks the fact that a number of FHIR resources are parsed with null IDs and overwrite one another when being stored in the HashMap under a null key, f.e. there are 301 naming systems in the R4 core definitions on Simplifier but only 293 end up being loaded and available for the validator, 8 have null IDs and therefore null keys, so they have to be loaded from the server and parsed by instantiating the huge JsonParser for each FHIR instance to be validate because they cannot be found in the map.

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
@@ -43,6 +43,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.TreeMap;
 import java.util.Set;
 
 import lombok.Getter;
@@ -199,7 +200,7 @@ public abstract class BaseWorkerContext extends I18nBase implements IWorkerConte
   private int serverQueryCount = 0;
   private final Set<String> cached = new HashSet<>();
   
-  private Map<String, Map<String, ResourceProxy>> allResourcesById = new HashMap<String, Map<String, ResourceProxy>>();
+  private Map<String, Map<String, ResourceProxy>> allResourcesById = new TreeMap<String, Map<String, ResourceProxy>>();
   // all maps are to the full URI
   private CanonicalResourceManager<CodeSystem> codeSystems = new CanonicalResourceManager<CodeSystem>(false);
   private final Set<String> supportedCodeSystems = new HashSet<String>();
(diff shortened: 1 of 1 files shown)


grahamegrieve:
we'll investigate why there's resources with null ids - there shouldn't be. But TreeMap is slower than HashMap, so we won't accept this PR (at least as is)

grahamegrieve:
What has to loaded from the server? I'm not following that bit

sanjan2:
Hi Grahame,

thanks for looking into this!

Regarding the null IDs, I have seen them for Naming Systems, Structure Definitions, Value Sets, Code Systems and Concept Maps so far.

I stumbled across this while checking which children of an instance took the validator the longest time to validate in class InstanceValidator, method validateElement:
 // 5. inspect each child for validity
 for (ElementInfo ei : children) {...

I got the impression that for the naming systems the xml parser parses their uniqueID element but does not store the ID therein (tag 'value') in the field 'id' of the Resource superclass, so the call r.getID for filling the HashMap in BaseWorkerContext may return null for a naming system even though it contains a uniqueID element unless that naming system also contains an id tag (possibly not unique) outside the uniqueID element.

For the Concept Maps, Structure Definitions and CodeSystems profiles the url tag for storing a unique URI is 0..1, is that as intended?

As for the server bit, please ignore that, I think it will sort itself out with the null keys.

Please feel free to disregard the TreeMap and come up with something better ;)

Best of luck,
Sanjan

March 22, 2022 8:14 PM, "Grahame Grieve" ***@***.*** ***@***.******@***.***>)> wrote:
	we'll investigate why there's resources with null ids - there shouldn't be. But TreeMap is slower than HashMap, so we won't accept this PR (at least as is)

grahamegrieve:
how do you reproduce this?

sanjan2:
example instances:
https://update.kbv.de/ita-update/DigitaleMuster/ERP/Beispiele_eRP_V1.0.2.zip (https://update.kbv.de/ita-update/DigitaleMuster/ERP/Beispiele_eRP_V1.0.2.zip)

profiles to validate against:
https://update.kbv.de/ita-update/DigitaleMuster/ERP/KBV_FHIR_eRP_V1.0.2_zur_Validierung.zip (https://update.kbv.de/ita-update/DigitaleMuster/ERP/KBV_FHIR_eRP_V1.0.2_zur_Validierung.zip)

German ePrescription which a client wants implemented
April 5, 2022 5:59 AM, "Grahame Grieve" ***@***.*** ***@***.******@***.***>)> wrote:
	how do you reproduce this?

PR #775, opened 2022-03-25
Dotasek oe r4r5structuremapconversion

dotasek:
## Work in progress

This expands on work in #772 

The conversion uses the value type for the converted StructureMapGroupRuleTargetParameterComponent.

However, expanding the test cases revealed that the conversion from r4 to r5 isn't working as intended, producing only the default StringType instead of the expected IdType in this test (and other constant types). This will cause the test to fail until we fix this conversion.

@grahamegrieve would it be appropriate here to re-use the code from the FHIRLexer to convert StringType in r4 to the appropriate type in r5 (IdType, IntegerType, etc)?


grahamegrieve (approved):
ok. And no, I don't think we should reuse the FHIR lexer cdde

PR #783, opened 2022-04-06
Allow other designations

gabriel0316:
With returning the url it is possible to allow other designation.use than only the two specified SNOMED codes.

Otherwise, no other designation having a different designation.use.code and designation.use.system would be displayed in the IG.


grahamegrieve (approved):
ok

PR #784, opened 2022-04-08
Display all designations

gabriel0316:
With this commit all designations (incl. languages) will be displayed in the Logical Definition and the Expansion of a ValueSet.

Based on https://chat.fhir.org/#narrow/stream/179202-terminology/topic/in.20ValueSet.20definied.20designation.20not.20rendered.20in.20IG.20Publi.2E.2E.2E

Related pull request: https://github.com/hapifhir/org.hl7.fhir.core/pull/783


gabriel0316:
@grahamegrieve If necessary I would also like to add a test case (presumably here https://github.com/FHIR/fhir-test-cases). However, I would be glad if you helped me how I could add one there.

grahamegrieve (approved):
ok

PR #785, opened 2022-04-08
Allow Validator CLI to load igs using specific versions

dotasek:
This change allows a user to prefix an -ig argument with a supported FHIR version enclosed in square brackets.

Example: 

```
-ig [3.0.1]my-custom-ig.json
```


grahamegrieve (approved):
ok

PR #786, opened 2022-04-11
Fix Base64BinaryType instantiation 

dotasek:
All Base64Binary types were not setting PrimitiveType values when set via setValue()

This fixes the issue and adds tests to prevent regression


grahamegrieve (approved):
ok

PR #796, opened 2022-04-23
Fix problem rendering canonical URLs with | in them

rob-reynolds:
In the Data Requirements section of the narrative for a Library that references versioned ValueSets,
IS:
```
        <tr>
            <td>code</td>
            <td>In ValueSet <a
                    href=\"http://example/fhir/ValueSet/medications|2.19.1\">http://example/fhir/ValueSet/medications|2.19.1</a>
            </td>
        </tr>
```
Note the href contains a "|" (pipe) character, which is an "unwise" url character that causes an error in the publisher (Java).

Should be:
no "|" character

### DIFF
--- a/org.hl7.fhir.r4b/src/main/java/org/hl7/fhir/r4b/renderers/DataRenderer.java
+++ b/org.hl7.fhir.r4b/src/main/java/org/hl7/fhir/r4b/renderers/DataRenderer.java
@@ -544,7 +544,11 @@ public class DataRenderer extends Renderer {
     if (uri.getValue().startsWith("mailto:")) {
       x.ah(uri.getValue()).addText(uri.getValue().substring(7));
     } else if (Utilities.isAbsoluteUrlLinkable(uri.getValue()) && !(uri instanceof IdType)) {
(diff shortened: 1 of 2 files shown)


grahamegrieve (commented):
I don't think that this is the right change. I think that the render should escape the | if it doesn't know what else to do with it

grahamegrieve:
And nor do I understand why this code does these changes for the IdType, where | is an illegal character

rob-reynolds:
My initial implementation was to escape it.  I even had "escape the |" in my original description of the **Should Be.**

However, then I saw this and thought this was how that sort of thing was intended to be addressed and changed my implementation:
image

rob-reynolds:
I guess I don't understand the distinction between the render overloads well enough.

I'll just add an issue.

rob-reynolds:
Replaced by https://github.com/hapifhir/org.hl7.fhir.core/issues/797.

rob-reynolds:
> And nor do I understand why this code does these changes for the IdType, where | is an illegal character

Doesn't the `&& !(uri instanceof IdType) `expression mean this code does NOT do these changes for the IdType?

Issue #795, opened 2022-04-22
Review usage of Resource.getId()


dotasek:
This issue (#766) revealed that Resource.getId() returns version history as well as the logical ID.

For example: `1234/_history/1`

There may be cases where this is not the expected response, and instead, the logical part of the ID is actually needed (`1234` in the above example).

Existing uses of getId() should be evaluated to verify that they are in fact meant to use the whole ID, including history.

This review should cover all modules:

- [ ] org.hl7.fhir.convertors
- [ ] org.hl7.fhir.dstu2
- [ ] org.hl7.fhir.dstu3
- [ ] org.hl7.fhir.dstu2016may
- [ ] org.hl7.fhir.r4
- [ ] org.hl7.fhir.r4b
- [ ] org.hl7.fhir.r5
- [ ] org.hl7.fhir.utilities
- [ ] org.hl7.fhir.validation

dotasek:
## org.hl7.fhir.convertors

In ExtensionDefinitionGenerator we're using the ID of a StructureDefinition. I don't expect these will ever have history in the same manner as a FHIR resource:
  https://github.com/hapifhir/org.hl7.fhir.core/blob/master/org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/ExtensionDefinitionGenerator.java#L178

dotasek:
## org.hl7.fhir.convertors.misc.argonaut
    ### ArgonautConvertor
        * Self-contained
        * Uses getId() to generate file names.
        * Uses getId() to generate references (this should be OK. References can use history)
## org.hl7.fhir.convertors.misc.iso21090
    ### ISO21090Importer
        * Self-contained
        * Generates ValueSet and StructureDefinition using getId() to generate URL. Likely not something we want _history for.
## org.hl7.fhir.convertors.misc.searchparam
    ### SearchParameterProcessor
        * Self-contained
        * Generates ConceptMap using getId() to generate URL. Likely not something we want _history for.

dotasek:
## org.hl7.fhir.dstu2
###    model

    ExpressionNode
        * Used in evaluating Type... likely not _history
###   test
        FluentPathTests
            * Disabled. There's a TODO there.
            * Seems to only apply to a StructureDefinition. Likely not _history
###  utils
        BatchLoader
            * Uses getId() to compose target URL for PUT. Valid with _history
        FhirPathEngine
            * Uses getId() from StructureDefinition. Likely not _history
[...]

PR #799, opened 2022-04-27
Use Resource::getIdPart() appropriately in code

dotasek:
Addresses this issue:

https://github.com/hapifhir/org.hl7.fhir.core/issues/795

Note: This PR is in draft while the review in #795 is ongoing.


grahamegrieve (commented):
set setId() doesn't match getId() in functionality? I think we should document that clearly

PR #818, opened 2022-05-24
Update ValidationTests to JUnit5

dotasek:
This is a test of converting ValidationTests to JUnit 5. This can be merged if approved, but other JUnit 4 tests should be converted in separate pull requests as well.


dotasek:
This is actually the same behaviour that was rejected here:

https://stackoverflow.com/questions/45869932/unable-to-run-tests-with-junit5-console-launcher

Won't merge; will have to work around with a JUnit 4 test runner or other solution.

Issue #831, opened 2022-06-03
Validator CLI complains with java.lang.ClassCastException (CapabilityStatement -> TerminologyCapabilities)


imadeit:
I am using the validator_cli.jar released recently ([v5.6.47](https://github.com/hapifhir/org.hl7.fhir.core/releases/tag/5.6.47)) to validate my local resource.
Our local terminology server is only able to deal with JSON format of capability statement, so I posted another issue #830 

Then I tried to return an XML capability statement at first round of request for /metadata?_summary=true, then another exception was thrown as follows.
```
Loading
  Load FHIR v4.0 from hl7.fhir.r4.core#4.0.1 - 4575 resources (00:06.0497)
  Load hl7.terminology#3.1.0 - 4117 resources (00:01.0744)
  Terminology server http://localhost:8081Exception in thread "main" org.hl7.fhir.exceptions.FHIRException: Unable to connect to terminology server. Use parameter '-tx n/a' to run without using terminology services to validate LOINC, SNOMED, ICD-X etc. Error = Error fetching the server's terminology capabilities
	at org.hl7.fhir.r5.context.SimpleWorkerContext.connectToTSServer(SimpleWorkerContext.java:343)
	at org.hl7.fhir.validation.ValidationEngine.connectToTSServer(ValidationEngine.java:373)
	at org.hl7.fhir.validation.ValidationEngine.connectToTSServer(ValidationEngine.java:362)
	at org.hl7.fhir.validation.ValidationEngine.setTerminologyServer(ValidationEngine.java:783)
	at org.hl7.fhir.validation.cli.services.ValidationService.initializeValidator(ValidationService.java:333)
	at org.hl7.fhir.validation.cli.services.ValidationService.initializeValidator(ValidationService.java:314)
	at org.hl7.fhir.validation.ValidatorCli.doValidation(ValidatorCli.java:241)
	at org.hl7.fhir.validation.ValidatorCli.main(ValidatorCli.java:165)
Caused by: org.hl7.fhir.exceptions.FHIRException: Error fetching the server's terminology capabilities
	at org.hl7.fhir.r5.utils.client.FHIRToolingClient.getTerminologyCapabilities(FHIRToolingClient.java:144)
	at org.hl7.fhir.convertors.txClient.TerminologyClientR5.getTerminologyCapabilities(TerminologyClientR5.java:84)
[...]

Issue #830, opened 2022-06-02
Add support for application/fhir+json


imadeit:
I am using validator_cli.jar to validate my Terminology Server, which has only implemented application/fhir+json format.
But validator_cli.jar seems only request for application/fhir+xml. 
So when I use validator_cli.jar to validate my Terminology Server with -tx option, by requesting the capability statement, errors are always returned.
By reading the source code, it seems the validator throw an parse exception from JSON, which it is expecting a XML response.
So, would it be possible to specify the data format when requesting for capability statement?
like:
java -jar ./validator_cli.jar data.json -version 4.0.1 -tx http://<my-terminology-server> -contentType application/fhir+json

grahamegrieve:
This should be seamless and automatic - try json first, and if that fails, fall back to json. I thought it worked that way, but apparently not

imadeit:
@grahamegrieve 
Yes, I thought exactly as you do. But unfortunately, it does only request for XML only.
Maybe there is hidden option to set the preferredResourceFormat by calling setPreferredResourceFormat() or something, I was just not able to figure out.

grahamegrieve:
that was a note for when @dotasek looks at this, as he will

PR #832, opened 2022-06-06
#830,#831. support for JSON only Terminology server automatically and make timeout longer for slow terminology server 

kazuhiko-ohe:
1) for #830
added new switchResourceFormat() from line#164  and calling it from within getCapabilitiesStatementQuick() at line#186-187  of org.hl7.fhir.r5.utils.client.FHIRToolingClient

By this addition, json only terminology server will also work. A Terminology server (eg.  https://tx.jpfhir.jp/ ) that supports only JSON format was tested and worked by this addition.  
2) Since external terminology server may need long time out (for example, in the case of a terminology server that dynamically generates a terminology capabilityStatement from its own CodeSystem and ValueSet Resource.), add the sentence TIMEOUT_NORMAL_EXTERNAL_TX = TIMEOUT_OPERATION_EXPAND , and used at line#143 calling  (TerminologyCapabilities) client.issueGetResourceRequest(...

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/client/FHIRToolingClient.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/client/FHIRToolingClient.java
@@ -83,7 +83,8 @@ public class FHIRToolingClient {
   private static final int TIMEOUT_ENTRY = 500;
   private static final int TIMEOUT_OPERATION_LONG = 60000;
   private static final int TIMEOUT_OPERATION_EXPAND = 120000;
-
+  private static final int TIMEOUT_NORMAL_EXTERNAL_TX = TIMEOUT_OPERATION_EXPAND;
+  
   private String base;
   private ResourceAddress resourceAddress;
   private ResourceFormat preferredResourceFormat;
@@ -139,7 +140,7 @@ public class FHIRToolingClient {
         getPreferredResourceFormat(),
         generateHeaders(),
         "TerminologyCapabilities",
-        TIMEOUT_NORMAL).getReference();
+        TIMEOUT_NORMAL_EXTERNAL_TX).getReference();
     } catch (Exception e) {
       throw new FHIRException("Error fetching the server's terminology capabilities", e);
     }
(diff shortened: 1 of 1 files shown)


dotasek:
The failing tests appear to be impacting all PRs, not just this one. I will try to diagnose what's wrong.

grahamegrieve:
I don't understand the question about timeouts. Was the server overloaded at the time?

kazuhiko-ohe:
>

kazuhiko-ohe:
@dotasek  Thank you for your advice. 
> * There are ~ no new test case covering the new code.
 I understand.  I'll try. 

> * The catch you added will catch any exception; I believe this needs to catch the exception specific to XML processing, or else our exception reporting will lose the first round of exceptions from any server.
> * Related to point 2, I'd rather have very specific decisions made when switching the preferred resource format. I would suggest something like this, expressed as code
Yes, I also thought that this proposal needed a more specific approach. I'll try to make better approach.
We have not yet found the location of the process that is detecting that the request response is not xml nor allowed. Servers that only process JSON are expected to return a response code of 406 or 421 when the client requests XML, so  I  could  check this response code and try again in JSON. The FHIR specification doesn't seem to specify a procedure to negotiate XML or JSON when getting the first capabilityStatement. The capabilityStatement has a format element, but unfortunately it cannot be used unless the capabilityStatement in the resource format that can be processed is responded.

kazuhiko-ohe:
@dotasek 
I could not why so many failings occured only by changing TIMEOUT parameter.

dotasek:
@kazuhiko-ohe The reported test failures are actually unrelated to your code; I edited my earlier message, and left another indicating this. Apologies if you still spent time investigating them.

Grahame and I are attempting to fix these ourselves; once they are repaired, I will re-run the pipelines on this PR so at least those tests pass again.

dotasek:
@grahamegrieve ignore my initial timing question; this PR changed the timeout for server access, and the initial failing tests were coincidentally reporting test failures on timingTiming elements in JSON. They are actually unrelated.

grahamegrieve:
I don't like the name TIMEOUT_NORMAL_EXTERNAL_TX - what does it mean, that constant?

dotasek:
@kazuhiko-ohe Please check #838

The solution proposed there includes tests, covers and SHOULD report all failures and attempt all available formats.

I am not changing the timeout. As per the tests, I'd expect a server to fail fast on such a basic query.

PR #838, opened 2022-06-17
Attempt different preferredResourceFormat after failed initial attempt

### LINKED ISSUE #830: Add support for application/fhir+json

dotasek:
This change attempts to get capability statements with the preferredResourceFormat first, and then attempts all others.

If the preferredResourceFormat fails, the next successful format will be set as preferred.

If no successful connections are made, this will throw an exception, as with previous behaviour.

Fixes #830 #831


grahamegrieve (approved):
ok

PR #841, opened 2022-06-22
Better errors when connecting to server that doesn't support Terminology Capabilities

dotasek:
Improves errors reported in cases where server doesn't return expected TerminologyCapabilities data.

See #831


grahamegrieve (approved):
ok

PR #843, opened 2022-06-23
Fix presentDuration representation of millis

### LINKED ISSUE #823: CLI output: timings have erroneous 0 as first fractional seconds digit
When I ran a timing series for issue #815 I noticed that there always was an inexplicable jump between sub-second timings and timings of a second or more, and that the jump was bigger than expected by about an order of magnitude. Then I noticed that all the timings had a zero as the first of the four fractional seconds digits, and the timings suddenly started to make sense when I removed the erroneous zero digit with a bit of Excel magic.

I tried to track down the erroneous 0 in the source code but I had no luck given the limited search facilities here on GitHub.

Instructions for reproduction: find a file that takes 1 second to validate and invoke the validator with a dozen or so copies of the filename. Timings going below 1 second will be 0.09ish or lower instead of 0.9ish or lower.

This is no showstopper, but it doesn't exactly enhance the glory of the validator CLI either. ;-)

dotasek:
Millis are being represented by four digits instead of 3 in CLI output

Fixes #823

### DIFF
--- a/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/DurationUtil.java
+++ b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/DurationUtil.java
@@ -2,7 +2,7 @@ package org.hl7.fhir.utilities;
 
 import java.util.concurrent.TimeUnit;
 
-public class TimeUtil {
+public class DurationUtil {
   public static String presentDuration(long duration) {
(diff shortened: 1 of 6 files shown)


dotasek:
Once this is merged and the master build completes, the following downstream PR should also be merged: https://github.com/HL7/fhir-ig-publisher/pull/453

grahamegrieve (approved):
ok

PR #865, opened 2022-07-18
Gg 202207 r5 update

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/model/Observation.java
```diff
@@ -3208,1072 +3801,17 @@ public boolean equalsShallow(Base other_) {
...
- /**
-   * Search parameter: <b>category</b>
-   * <p>
-   * Description: <b>The classification of the type of observation</b><br>
-   * Type: <b>token</b><br>
-   * Path: <b>Observation.category</b><br>
-   * </p>
-   */
-  @SearchParamDefinition(name="category", path="Observation.category", description="The classification of the type of observation", type="token" )
-  public static final String SP_CATEGORY = "category";
```
dotasek, line 3256: What happened to all these search parameter constants? Do they exist elsewhere?

PR #870, opened 2022-07-21
Fix for #869 Use backbone element conversion

### LINKED ISSUE #869: Unhandled BackboneElement conversion
In the converter libraries, copyElement is used throughout, and ignores cases where Backbone Elements should be copied instead.

All converters should use a copyBackboneElement method where appropriate

Progress on this change will be tracked by version:

- [x] conv10_30
- [x] conv10_40
- [x] conv10_50
- [x] conv14_30
- [x] conv14_40
- [x] conv14_50
- [x] conv30_40
- [x] conv30_50
- [x] conv40_50

- [x] check above for instances of copyElement(src, tgt, var) as well
- [x] review test cases

dotasek:
Fixes #869


dotasek:
Merged as per @grahamegrieve's request.

PR #874, opened 2022-07-22
Rework precision checking in datetime

jamesagnew:
Avoid throwing an exception when parsing resources, in order to prevent people from being unable to parse known invalid resources.


dotasek (commented):
So when parsing resources, we're calling setValueAsString(theString) so we don't throw an exception. But when a programmer is creating resources through the constructor, it is checked, and we do throw an exception.

Two questions: 
1. The changes in these test cases basically remove an assertion that .setValueAsString throws an exception. Should there be an assertion in another test that .setValueAsString DOESN'T throw an exception when invalid?
2. The exception type changes from  DataFormatException to IllegalArgumentException in a few cases. Is that then going to be caught differently in the validator?

dotasek:
Merged per @jamesagnew request

PR #890, opened 2022-08-05
Temporary patch to restore Enumerations.FHIRVersion.toString()

dotasek:
This is a temporary patch to restore the fhir publisher build.


dotasek:
Rejecting; the explicit use of toCode() in kindling is preferred to restore the fhir publication build.

This PR implements that: https://github.com/HL7/kindling/pull/84

Issue #896, opened 2022-08-16
IG Publisher NP in PatientRenderer


oliveregger:
With IG Publisher version greater than 1.1.125 I get an exception with the IG Publisher and our ig:

ci-build: http://build.fhir.org/ig/hl7ch/ch-orf/branches/mapping_110/failure/build.log

java.lang.NullPointerException: Cannot invoke "org.hl7.fhir.r5.model.HumanName$NameUse.ordinal()" because "oldUse" is null
        at org.hl7.fhir.r5.renderers.PatientRenderer.isPreferred(PatientRenderer.java:104)
        at org.hl7.fhir.r5.renderers.PatientRenderer.chooseName(PatientRenderer.java:93)
        at org.hl7.fhir.r5.renderers.PatientRenderer.display(PatientRenderer.java:44)
        at org.hl7.fhir.r5.renderers.ResourceRenderer.describe(ResourceRenderer.java:102)
        at org.hl7.fhir.r5.renderers.PatientRenderer.render(PatientRenderer.java:31)
        at org.hl7.fhir.r5.renderers.BundleRenderer.renderDocument(BundleRenderer.java:207)
        at org.hl7.fhir.r5.renderers.BundleRenderer.render(BundleRenderer.java:281)
        at org.hl7.fhir.igtools.publisher.Publisher.getXhtml(Publisher.java:9725)
        at org.hl7.fhir.igtools.publisher.Publisher.saveDirectResourceOutputs(Publisher.java:8958)
        at org.hl7.fhir.igtools.publisher.Publisher.generateHtmlOutputs(Publisher.java:8275)
        at org.hl7.fhir.igtools.publisher.Publisher.generate(Publisher.java:6049)
        at org.hl7.fhir.igtools.publisher.Publisher.createIg(Publisher.java:1034)
        at org.hl7.fhir.igtools.publisher.Publisher.execute(Publisher.java:871)
        at org.hl7.fhir.igtools.publisher.Publisher.main([Publisher.java:10297](http://publisher.java:10297/))

will provide a PR with a possible fix

PR #897, opened 2022-08-16
Fix NP for PatientRenderer #896

oliveregger:
PR with avoiding the NP that the switch statement is not called and
returns true if newUse has NameUse.OFFICIAL, NameUse.USUAL (same
expression as value for NULL in switch statement).


grahamegrieve (approved):
ok

PR #902, opened 2022-08-24
Return lists for FHIRPath constants in r4

### LINKED ISSUE #881: R4 FHIRPathEngine resolveConstant of EvaluationContext doesn't support returning multiple values
In the context of a [conversation](https://chat.fhir.org/#narrow/stream/179266-fhirpath/topic/variables/near/290859425) with @grahamegrieve over zulip chat, 

R4 FHIRPathEngine [public Base resolveConstant(Object appContext, String name, boolean beforeContext) method](https://github.com/hapifhir/org.hl7.fhir.core/blob/master/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/FHIRPathEngine.java#L202) of EvaluationContext Interface doesn't support returning multiple values.

cc: @grahamegrieve

dotasek:
Fixes #881 using the same implementation used in r5/r4b


grahamegrieve (approved):
ok

Issue #901, opened 2022-08-24
BaseDateTimeType inconsistent setFieldValue


dotasek:
From a conversation with @JasonRoberts-smile . I believe this is using r4 models:

> I've found something weird in the core libraries, and I would like your opinion on it.
> 
> For context, I'm working with some legacy code from an open-source project that uses the FHIR model. It is converting a string to a DateTimeType by picking out substrings and making calls to setYear(), setMonth(), setDay(), etc.
> 
> In BaseDateTimeType.setFieldValue(), when the value of the object has not yet been initialized, it looks like we are creating a new GregorianCalendar(0, 0, 0) as the initial state. These zeroes are causing some strange outcomes.
> 
> When the first field we set is the year, the 0 in the dayOfMonth field means that the call to getTime() at the bottom of the method will return a Date object representing Dec 31 of the year before the one we specified.
> 
> When the first field we set is the day, the 0 in the year field results in a Date object representing that day in January, 1 BCE. Subsequent calls to setYear() can change the number of the year, but the epoch remains BCE. There doesn't seem to be any way to change the epoch through the API of the class (and why would there be, I suppose?).
> 
> When the first field we set is anything else, both these things happen. So we are getting a variety of wild results, depending on the order in which we call the setters, none of them correct.

> I'm working around it for now by initializing my DateTimeType instance to Jan 1, 1 CE via its constructor before I start setting the individual fields. But I wanted to bring this to your attention, because it doesn't seem like the intended behaviour.

dotasek:
I'm going to add a test case for this soon.

Until then, @grahamegrieve I think this would be in the generated model code. Am I correct in this?

dotasek:
Failing test: 

```java
  @Test
  public void testFieldSet() {
    final int YEAR = 1979;
    final int MONTH = 1;
    final int DAY = 23;
    final DateTimeType dateTimeYearFirst = new DateTimeType();
    dateTimeYearFirst.setPrecision(TemporalPrecisionEnum.DAY);
    dateTimeYearFirst.setYear(YEAR);
    dateTimeYearFirst.setDay(DAY);
    dateTimeYearFirst.setMonth(MONTH);

    final DateTimeType dateTimeDayFirst = new DateTimeType();
    dateTimeDayFirst.setPrecision(TemporalPrecisionEnum.DAY);
    dateTimeDayFirst.setDay(DAY);
    dateTimeDayFirst.setYear(YEAR);
[...]

PR #903, opened 2022-08-24
Fix DateTimeType inconsistencies when using field setting

dotasek:
Fix for #901 

When setting the fields of a no-arg constructed DateTimeType, it was using setFieldValue(), which in turn used new GregorianCalendar(0, 0, 0);

Switching to GregorianCalendar() fixes this issue.


dotasek:
If this change passes tests across all OS/JVM combos, the remaining models (dstu3, dstu2016may, dstu2) should also be updated.

dotasek:
dstu2016may and dstu3 have been added. dstu2 does not have setFieldValue(), so is omitted. This is ready for review.

grahamegrieve (approved):
ok

PR #919, opened 2022-09-08
WIP Add new param to BaseAdvisor_40_50

dotasek:
This adds a new field to BaseAdvisor_40_50, `produceIllegalParameters`, which manages the conversion of 'illegal' parameters 'Illegal' parameters are those whose code is not in the following list: 

```java
"apply", "path-resource", "path-pages", "path-tx-cache", "expansion-parameter", "rule-broken-links", "generate-xml", "generate-json", "generate-turtle", "html-template"
```

By default, any parameters not in that list will be added as an extension. However, if the flag `produceIllegalParameters` is set to true, the parameter will be added as is.


grahamegrieve (approved):
ok

PR #926, opened 2022-09-19
Show display name of concepts which are not mapped

gabriel0316:
For concepts that have not corresponding target their display name should be displayed if present.


KevinDougan-SmileCDR:
@gabriel0316   Can you please also add a Unit Test which verifies the expected outcome of this change in behaviour?

grahamegrieve:
see https://github.com/FHIR/fhir-test-cases/r5/narrative

grahamegrieve:
new entries is OK if there's unmapped concepts

gabriel0316:
I guess the the html-files along with the json have to be adapted in order to have everything set for the tests.

gabriel0316:
Just created https://github.com/FHIR/fhir-test-cases/pull/118 - hope its correct

grahamegrieve (approved):
ok

PR #928, opened 2022-09-22
Enhanced testing for character encoding issues

dotasek:
Occasionally CLI users or developers will execute on Java environments that use different character encodings. For example: Eclipse Temurin, on Windows, uses Cp1252 instead of UTF-8, which can cause mystery errors).

This PR adds a Linux job that switches the character encoding explicitly. The Windows image as it stands would also find this issue, BUT Azure can swap out the JVM used in its images as it updates. It could be that the next update results in the use of a JVM that doesn't use that encoding anymore. Linux and Java 17 are chosen for this task because they appear to run quickly in comparison to Mac and Windows builds.

This PR will also fix any bugs caught by the new Cp1252 job.


grahamegrieve (approved):
ok

PR #933, opened 2022-09-28
Warn CLI users if file.encoding is not UTF-8

dotasek:
Since having a file.encoding other that UTF-8 can occasionally result in hard-to-diagnose errors, the Validator CLI will now check the default encoding, and print a warning if it is anything else.


grahamegrieve (approved):
ok

PR #937, opened 2022-09-30
Fix intermittent failure of SnapshotGenerationTests

dotasek:
DO NOT MERGE YET

There may be more places where IWorkerContext re-use is being mutated.


dotasek:
This PR is being replaced with https://github.com/hapifhir/org.hl7.fhir.core/pull/947 

There were recent refactors of IWorkerContext that made it easier to apply this fix than to do a merge.

PR #939, opened 2022-10-04
When we load StructureMaps from mapping files, make the mapping langu…

lmckenzi:
…age the narrative.

Right now the narrative is empty, which is useless.  Retaining the map file information is the most human readable mechanism we have for StructureMaps


grahamegrieve (approved):
ok

PR #947, opened 2022-10-11
Fix intermittent failure of SnapshotGenerationTests

dotasek:
DO NOT MERGE YET. There may still be instances of test classes that are still mutating a shared context.

Uses copies of IWorkerContext to better separate test classes.


grahamegrieve:
Not necessary anymore

PR #956, opened 2022-10-18
Use a copy of worker context from TestingUtilities instead of an actual instance

dotasek:
Currently when TestingUtilities provides a IWorkerContext, it provides a reference to an instance. This can cause problems down the line if a test loads a package that interferes with the results of another test, as tests are run in random order.

This change makes TestingUtilities provide a copy of a standard worker context instead, which means that individual tests will load packages to a copy of a context, instead of a single shared instance, preventing the issues above.


grahamegrieve:
So doing this turned out way more complicated than I realised, but I got it working in another PR. I'd close this, but I didn't make changes to the test cases - do we need to make those changes?

PR #964, opened 2022-10-27
Internationalization Pluralization

dotasek:
WIP

This adds plural management to Internationalization


grahamegrieve (commented):
why does FHIRPathEngine get involved in the plural rules? Shouldn't that be handled somewhere else?

dotasek:
@grahamegrieve FHIRPathEngine isn't involved in any of the plural rules directly; it defines makeExceptionPlural, which calls a formatMessagePlural and composes a full error message using it. 

It looks like in the majority of these cases, it's saying something like:

> Expected one value, but 23 values were found. 

We can skip defining a singular phrase in this case, and only need to define the _other (plural) form in English. The same applies to de and nl. 

Where calling formatMessagePlural would come in handy is for Polish and other languages with multiple plural forms (one, few, many, other, etc.).

grahamegrieve (approved):
ok

PR #971, opened 2022-10-30
Fix problem with SVG rendering

lmckenzi:
The SVG namespace was getting stripped from included SVG files - and hyperlinks were not working when using img instead of include.

### DIFF
--- a/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/XhtmlParser.java
+++ b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/XhtmlParser.java
@@ -483,7 +483,6 @@ public class XhtmlParser {
     for (String an : node.getAttributes().keySet()) {
       if (an.equals("xmlns")) {
         result.def(node.getAttribute(an));
-        nsattrs.add(an);
       }
       if (an.startsWith("xmlns:")) {
         result.ns(an.substring(6), node.getAttribute(an));


dotasek:
This should have a test; will work on putting one together.

lmckenzi:
You can see an example of the manifestation of the problem here: https://build.fhir.org/exchanging.html#decision-tree

Included SVG *should* look like it does here: http://build.fhir.org/ig/HL7/davinci-ehrx/exchanging.html#decision-points

You can see in the source how the namespace has been stripped

grahamegrieve:
Well, I for one don't understand how the proposed fix solves the problem, and if it does it's probably an accident. We need an XHTML parsing test case, not a comparison between two different IGs

lmckenzi:
The current code strips the xmlns= attribute.  My change stops that from happening.  I really don't know why the original code was doing that.

grahamegrieve:
yes but why does your change do that? looks to me like it works by increasing the bugs

lmckenzi:
My change removes the xmlns= attribute from the list of attributes to be stripped.

grahamegrieve:
That doesn't make sense; the xmlns attribute will be added back in. So it gets duplicated? Or if it's not being added in, why is not being added in? I'm not going to approve adding a bug on top of a bug

lmckenzi:
The namespace is *not* added back.  It's just stripped.  I don't know what the current code is trying to do or why it's trying to do it.  Having a namespace declared explicitly on a node isn't a problem.  Replacing it with a prefix (if that's the intention, though it's not happening) doesn't seem like a necessary step.  However, if that's what's desired, the code doesn't work, and I'm not sure how to make it work.

grahamegrieve:
Handled differently in a different PR

PR #972, opened 2022-10-31
Fix pipes in terminologyCache files

dotasek:
http://terminology.hl7.org/CodeSystem/ urls containing pipe characters would result in filenames containing those pipes. This is an illegal character for filenames in Windows.

This PR replaces those characters with an X character, which is consistent with how we replace other special path chars in the cache.


grahamegrieve (approved):
ok

PR #984, opened 2022-11-09
Remove excluded tests

dotasek:
There is an exclusion for tests in /validation/cli, which results in tests that can be executed in an IDE, but that aren't actually run by our builds. This removes that exclusion to allow those tests to run.


dotasek:
At @markiantorno's suggestion, I ran a deploy locally using this branch, and other maven projects appear to be able to pick up the deployed artifact. I'm going to merge this in on Monday on the off chance that it has other impacts.

PR #985, opened 2022-11-10
Do not delete previously undefined xmlns attributes when processing Xhtml nodes

dotasek:
By default, xmlns attributes are stripped from an XHTML node.

Previously, only the xmlns attribute was added back when:

1. It was a root node.
2. It was a non-root node and its namespace was different than the default.

In the case of SVG nodes, this resulted in necessary namespaces like xlink being stripped.

This change adds back any namespaces that haven't been defined in the parent node.

### DIFF
--- a/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/XhtmlParser.java
+++ b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/XhtmlParser.java
@@ -43,6 +43,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.hl7.fhir.exceptions.FHIRException;
 import org.hl7.fhir.exceptions.FHIRFormatError;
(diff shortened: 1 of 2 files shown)


grahamegrieve (commented):
how can a default namespace have a prefix?

dotasek:
That was done to manage this case:

```xml
<xhtml:div xmlns:xhtml="http://www.w3.org/1999/xhtml">
   <xhtml:img src="http://pbs.twimg.com/profile_images/544507893991485440/r_vo3uj2_bigger.png" alt="Twitter Avatar"/> 
   @fhirabend
</xhtml:div>
```

Which should transform to:

```xml
<div xmlns="http://www.w3.org/1999/xhtml">
   <img src="http://pbs.twimg.com/profile_images/544507893991485440/r_vo3uj2_bigger.png" alt="Twitter Avatar"/>
   @fhirabend
</div>
```

Reviewing it now, that naming and logic isn't at all obvious. It might well be the right, but I wouldn't know because the correct result could be a side-effect for all I know...

dotasek:
This requires https://github.com/FHIR/fhir-test-cases/pull/123 to be merged before tests will pass.

dotasek:
Renamed some of the logic. I think this is more coherent now.

PR #989, opened 2022-11-14
Fix missing space in slicing validation failure message

vadi2:


org.hl7.fhir.utilities/src/main/resources/Messages.properties
```diff
@@ -380,8 +380,8 @@ Unable_to_resolve_name_reference__at_path_ = Unable to resolve name reference {0
 Details_for__matching_against_Profile_ = Details for {0} matching against profile {1}
 Does_not_match_slice_ = Does not match slice ''{0}'' (discriminator: {1})
 Profile__does_not_match_for__because_of_the_following_profile_issues__ = Profile {0} does not match for {1} because of the following profile issues: {2}
-This_element_does_not_match_any_known_slice_ = This element does not match any known slice{0}
-defined_in_the_profile = defined in the profile 
+This_element_does_not_match_any_known_slice_ = This element does not match any known slice {0}
```
vadi2, line 383: This is the only real change

The rest were whitespace fixes made by github.dev

grahamegrieve:
This will require changes to the validator test cases too

dotasek:
This looks like it will change the expected message by deleting `defined_in_the_profile`

See the usage of that property here:

https://github.com/hapifhir/org.hl7.fhir.core/blob/cf7ef7218ad573abdd15ea56070e754ef380ea28/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java#L5790

Otherwise, the space fix looks correct.

vadi2:
`defined_in_the_profile` is still there, vscode just deleted the space at the end of that line

Are the test cases in https://github.com/FHIR/fhir-test-cases ? I had a look but didn't find this exact error message I'm fixing.

grahamegrieve:
"This element does not match any known slice" appears a number of times in the validation results

vadi2:
It does, but not the one where a space was missing

grahamegrieve:
Wrong. There are several examples where the tests fail after this

dotasek:
@vadi2 apologies, the code snippet above was hiding that portion. It was clearly what you said in the full diff.

PR #1001, opened 2022-11-22
fix np and index out of bound exception

oliveregger:
while switching to the lastest org.hl7.fhir.core we noticed in our tests a nullpointer exception and index out of bound exceptions happened:

1.  ParserBase excepts a path, but this is not set when within root
2. The expression is in the first getParameter element if the size is only one (should apply same logic as in analyzeTransform https://github.com/hapifhir/org.hl7.fhir.core/blame/master/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/structuremap/StructureMapUtilities.java#L2352)


grahamegrieve (approved):
ok

Issue #408, opened 2020-12-23
Option locale without effect


mmreith:
The option locale has no function

-locale en

does not change the output

grahamegrieve:
what language did you try?

mmreith:
just english (en). Language is set to german by default through windows command, but i need the display in english for debugging purposes.

markiantorno:
I'm assuming this is regarding the output of the validator cli @mmreith ?

mmreith:
@markiantorno correct. Do you need more information?

markiantorno:
@mmreith can you send me the exact steps to reproduce please?

Thank you.

mmreith:
@markiantorno My call was, but you could use whatever you like: java -jar validator_cli.jar -ig aws-zertifizierung\Sollprofile\PF20_all_Mussfelder -ig aws-zertifizierung\Sollprofile\PF24 -ig aws-zertifizierung\Sollprofile\PF22_24_Mussfelder -ig KBV_FHIR_AW\KBV-Basis -ig KBV_FHIR_AW\Terminologie -ig KBV_FHIR_AW\Application -ig KBV_FHIR_AW\Extensions-KOPIE -ig de.basisprofil.r4#0.9.11 -version 4.0.1 -recurse -no-extensible-binding-warnings **-locale en** Mustermann_Martina_5faa0942-a96c-43c7-ad25-0854829bd083_AW.xml

but i did not get an output in english. If you want to reproduce that, you might have to set **-locale de**, since this is an available translation,

mmreith:
@markiantorno do you need more information?

DarthGizka:
I'm having the same problem and now I find that it has already been reported here.

F7 confirms that the parameter gets parsed correctly and that `CliContext.locale` (which defaults to English anyway) gets set accordingly. Couldn't find where the translation occurs because it doesn't happen when I run the jar under the IntelliJ debugger (had `-locale nl` in order to make it differ from both the system locale and defaultish locales like C or en).

One problem definitely lies in `CliContext::setLocale(Locale)`, which receives a valid Locale object but stores only its localised display language name. In `CliContext::getLocale()` there is a half-hearted attempt at getting a Locale object for this stored string, but that must fail because a localised display language name is not a valid BCP 47 language tag as expected by [`Locale.forLanguageTag(String)`](https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#forLanguageTag-java.lang.String-).

[...]

PR #1002, opened 2022-11-23
Fixed -locale parameter for validatorcli

Hankych:
Implements #408


dotasek (approved):

grahamegrieve:
so does this change the output language? Can we add an end-to-end test of choosing output language?

PR #1010, opened 2022-11-26
Update to Java 11 Minimum

dotasek:
This change drops Java 8 support, and requires a Java 11 build.


dotasek:
Still doesn't work on Windows. For reasons. Do not merge.

Issue #1008, opened 2022-11-25
Immmunization r4 to dstu3 conversion handles notGiven incorrectly


dotasek:
In the R3 -> R4 and R4 -> R3 conversion maps (https://www.hl7.org/fhir/R4/immunization-version-maps.html) the notGiven value does not appear to be correctly handled in either direction for the `"notGiven": true` and ` "status": "not-done"` cases in R3 and R4 respectively. 

The expected conversion is outlined in the additional test cases in this branch: https://github.com/hapifhir/org.hl7.fhir.core/tree/do-immunization30_40-notgiven

Relevant test and test resources:

https://github.com/hapifhir/org.hl7.fhir.core/blob/do-immunization30_40-notgiven/org.hl7.fhir.convertors/src/test/java/org/hl7/fhir/convertors/conv30_40/Immunization30_40Test.java
https://github.com/hapifhir/org.hl7.fhir.core/blob/do-immunization30_40-notgiven/org.hl7.fhir.convertors/src/test/resources/immunization_30.json
https://github.com/hapifhir/org.hl7.fhir.core/blob/do-immunization30_40-notgiven/org.hl7.fhir.convertors/src/test/resources/immunization_40.json

dotasek:
Additional test cases may need to be added to ensure that all r4 statuses `completed | entered-in-error | not-done` are converted appropriately.

PR #1016, opened 2022-11-29
Fix handling of Immunization notGiven and status

dotasek:
Fixes https://github.com/hapifhir/org.hl7.fhir.core/issues/1008


org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/conv30_40/resources30_40/Immunization30_40.java
```diff
@@ -10,10 +10,17 @@
...
+
 import org.hl7.fhir.exceptions.FHIRException;
+import org.hl7.fhir.r4.model.Extension;
+import org.hl7.fhir.r4.model.Immunization;
+
+
 
 public class Immunization30_40 {
 
+  public static final String NOT_GIVEN_EXTENSION_URL = "http://hl7.org/fhir/3.0/StructureDefinition/extension-Immunization.notGiven";
```
lukedegruchy, line 22: FYI:  Opening this URL gives me a page with

image

lukedegruchy (approved):

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/conv30_40/resources30_40/Immunization30_40.java
dotasek: Those URLs don't need to resolve (ideally, the do, but they are not required).

It should match this from the R3->R4 conversion map:

```
src.notGiven as v ->  tgt.extension as vt,  vt.url = 'http://hl7.org/fhir/3.0/StructureDefinition/extension-Immunization.notGiven',  vt.value = v;
```

dotasek:
From a conversion with @grahamegrieve: we should remove the extension on the way back from R4 to R3. There is a mechanism for doing this built in, so it shouldn't be a chunk of custom code.

grahamegrieve (approved):
ok

PR #1018, opened 2022-11-29
Correct GraphQL Schema Validation Errors

jamesagnew:
The current GraphQL schema generator generates schemas that fail validation. The issue is that inheritance in GraphQL has to be done with interfaces (ie you can't extend a type, only an interface) so fragments like the following don't work:

(Ignore the line numbers, they are a fragment of HAPI's logging)

```
52: type Age implements Quantity {
53:   id: String
54:   extension: [Extension]
55:   value: decimal  _value: ElementBase
56:   comparator: code  _comparator: ElementBase
57:   unit: String  _unit: ElementBase
58:   system: uri  _system: ElementBase
59:   code: code  _code: ElementBase
60: }
```

```
743: type Quantity implements Element {
744:   id: String
745:   extension: [Extension]
746:   value: decimal  _value: ElementBase
747:   comparator: code  _comparator: ElementBase
748:   unit: String  _unit: ElementBase
749:   system: uri  _system: ElementBase
750:   code: code  _code: ElementBase
751: }
```

As it turns out, inheritance in GraphQL is pretty bizarre:
* Only interfaces can really do any kind of true inheritance
* Interface inheritance is transitive but is required to be explicit (e.g. if `Dog` implements `Animal` and `Animal` implements `LivingThing`, then it's an error for `Dog` to not also explcitly implement `LivingThing`)
* Elements defined in an interface have to be repeated in the type definition that implenents that interface (ie you don't get them for free through the inheritance)

This PR resolves these issues so that the schema fully validates.

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/GraphQLSchemaGenerator.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/GraphQLSchemaGenerator.java
@@ -298,12 +298,12 @@ public class GraphQLSchemaGenerator {
       List<StringBuilder> list = new ArrayList<>();
       StringBuilder b = new StringBuilder();
       list.add(b);
-      b.append("type ");
-      b.append(sd.getName());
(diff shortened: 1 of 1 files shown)


grahamegrieve (approved):
ok

Issue #1019, opened 2022-11-29
FHIR Mapping Language: cast to positiveInt not yet supported


oliveregger:
Using a cast with positiveInt leads to a message that it is not yet supported. [Testcase](https://github.com/FHIR/fhir-test-cases/pull/124) and PR for functionality will be provided.

oliveregger:
https://github.com/hapifhir/org.hl7.fhir.core/pull/1020

PR #1020, opened 2022-11-29
FHIR Mapping Language: cast support for more types #1019

oliveregger:
adds support for additional primitive types that maps can use or example:

```
ext.value = cast(value,'positiveInt')
```
testcase added with https://github.com/FHIR/fhir-test-cases/pull/124


grahamegrieve (approved):
ok

Issue #1022, opened 2022-11-30
BUG: BaseWorkerContext.processValidationResult expects wrong type for validate-code results


lawley:
This method incorrectly casts Paramters values to StringType for the OUT results `system` and `code`.
As per the FHIR Spec, these should be `uri` and `code` respectively.

Note that tx.fhir.org incorrectly returns these values as StringType.

For robustness suggest casting to `PrimitiveType<?>` and using `asStringValue()`

PR #1023, opened 2022-11-30
Fix handling of validate-code response for #1022

lawley:
To accommodate tx.fhir.org behaviour, cater for `valueString` as well as correctly typed response values


dotasek:
What specific behaviour/response requires this change?

Is there a clear test that can illustrate the error?

grahamegrieve (approved):
ok

lawley:
This change allows conferment terminology servers to be used with -tx option in the ig publisher (e.g., currently this fails with ontoserver because code relies on bad response from tx.fhir.org).
I've now managed to add a unit test.

PR #1027, opened 2022-12-05
Profile Utilities refactor

dotasek:
This change moves the path processing code for Profile Utilities to its own class.

This will allow later refactors to split the process into smaller logical chunks.


grahamegrieve (approved):
ok

PR #1038, opened 2022-12-13
Fix/improve OSGi bundle wiring

bdenton:
This PR makes a couple of changes to the OSGi bundle wiring

1. Remove versioning on hapi-fhir bundle imports. This is needed because, otherwise, OSGi will use whatever version the org.hl7.* projects were built against. The hapi-fhir bundles have a version dependence on org.hl7.*... not the other way around
2. Make a number of bundle imports optional. For instance, org.hl7.validation does not need to depend on all of the Fhir structure versions. It should just get wired to the version being processed. There were other dependencies needed for testing that were getting loaded unnecessarily in live deployments.


bdenton:
The build for this PR will work once `org.hl7.fhir.r5.test.NarrativeGeneratorTests` is fixed. This test failure is not related to this PR but seems to be caught up in the changes to `ProfileUtilities`

grahamegrieve:
@bdenton I don't understand this PR, and maybe it's because I don't understand OSGI, but:

>  This is needed because, otherwise, OSGi will use whatever version the org.hl7.* projects were built against.

well, yeah, that's the point? what other version would give reliable outcomes?

> Make a number of bundle imports optional. For instance, org.hl7.validation does not need to depend on all of the Fhir structure versions. It should just get wired to the version being processed. 

but it does depend on all the versions through the convertor package?

bdenton:
@grahamegrieve Thanks for the feedback.. here are my thoughts..

1. "Remove versioning on hapi-fhir bundle imports."... this is the result of the two projects depending on each other. When I created this PR, I was trying to build an OSGi-based Fhir server based on the latest builds of both of the projects. The problem was that, although the bundles from hapi-fhir v6.x correctly resolved its imports at runtime, the bundles from this project could not resolve the imports from hapi-fhir v5.x. (I may have the versions a bit mixed up.. it's been a while).
So, I figured that hapi-fhir was where the real dependency needed to be so I removed the import versions in this project. This way, this project can be a dependency to both hapi-fhir v5.x and v6.x... I suppose another solution would be to change the import version to be the hapi-fhir version built against **and any subsequent version**

2. "Make a number of bundle imports optional".. this is also the result of trying to build an OSGi-based server for a specific version of FHIR. For instance, if I am building a server that will only be dealing with DSTU3 resources, then I shouldn't need to drag in all of the R4, R5, etc. models and converters at runtime as well... making the imports optional makes this possible.
Another solution would have been to restructure the packaging of this project to eliminate or reduce the need to reference multiple FHIR versions in the same bundle. Also, further separating functionality needed by a FHIR server from functionality used by non-interactive components (e.g. CLI) would help.

grahamegrieve:
@dotasek I'll defer to you on #1. @bdenton with regard to #2, if you're just using DSTU3, then you use utilities and the DSTU3 package, and you don't get any other dependencies?

bdenton:
@grahamegrieve Consider the following changes in this PR..
The utilities bundle has non-optional Package-Imports for a bunch of test bundles (Jupiter, bytebuddy, lombok, etc) that are not ever used in "normal" operation. This causes all of those packages to get loaded by the OSGi framework at runtime even if they are never used in most situations. My change is to make these Package-Imports optional so that they will only resolve if the test bundles are also being loaded based on OSGi framework configs.
The same thing is also happening in the validation bundle... it has non-optional Package-Imports for all of the FHIR model versions. If I want to build an OSGi-based server for, say, R4, then I shouldn't need to drag along all of the other models

grahamegrieve:
ok, so I follow the utilities / tests things. But the validator is hard-coded to use R5 internally, so that's not optional, it's only the other versions that could be optional. Or at least, it's not useful for it to be optional. Or have I mis-understood? If OSGi only loads what's needed, why isn't everything optional?

bdenton:
Just a note on OSGi behaviour..
When each bundle is loaded, the framework works to resolve all "dependencies" as enumerated in the "Import-Package" manifest statement. The framework will not "start" the bundle until all dependencies are resolved. Obviously, the framework does not which of these "Import-Packages" will actually get called. 
So, for the validation bundle, the current POM will cause the OSGi framework to load *all* of the FHIR model bundles and will fail if they are not all resolvable even if only 1 or 2 are actually going to get called at runtime.

dotasek:
Two things here: version-wise, I agree with Grahame. The version of org.hl7.fhir.validator is bound to the version of org.hl7.fhir.r5, org.hl7.fhir.r4b, etc. and trying to use different version will likely result in very broken behaviour. If there's a way to enforce matching versions we should be doing it.

The relationship between this and the HAPI parent pom is something I have to look into more deeply. It causes confusion to the HAPI devs as well. If there were a way to constrain the version stripping to just that parent POM, that would be much more ideal.

One thing I did notice here is that this change will break two things:

1. The validator CLI will cease to function, as the dependencies it needs are no longer there.
2. Downstream projects like fhir-ig-publisher and kindling will also cease to function for the same reason.

I've mirrored this PR with the changes necessary to the CLI pom here: https://github.com/hapifhir/org.hl7.fhir.core/tree/do-20230502-osgi-bundle-wiring

I seem to have gotten fhir-ig-publisher running again with similar changes, and will try the same for kindling.

dotasek:
@bdenton Have you had a chance to look at the changes I did on my branch?

dotasek:
As per discussion with @grahamegrieve, since there is no further response, we are closing this. Please re-open if this is not the case.

PR #1044, opened 2022-12-19
Gg 202212 profiled element

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/profilemodel/ProfiledElement.java
```diff
@@ -0,0 +1,159 @@
...
+   * @return the definition of the element in the base specification
+   */
+  public ElementDefinition baseDefinition() {
+    throw new NotImplementedException("Not done yet");
+  }
+  
+  /**
+   * @return the short documentation of the definition (shown in the profile table view)
+   */
+  public String shortDoco() {
```
dotasek: I think we can use the extra characters for shortDocumentation

dotasek:
I think it would be cleaner and more testable for the API to be encapsulated outside of IWorkerContext, possibly in something like a ProfileElementBuilder returned by IWorkerContext.

I'd like the dataMode to be a bit clearer too. Maybe something explicitly following the commentary, like `hasResourceBase` or `basedOnResource`, or even a class extending ProfiledElement called ResourceBasedProfiledElement.

grahamegrieve:
I intended to have something like ProfileElementBuilder in the background actually doing the work. How would it make the API better?

grahamegrieve:
I considered something like ResourceBasedProfiledElement but then you'd either have to overload getInstance() or typecast everywhere

dotasek:
> I intended to have something like ProfileElementBuilder in the background actually doing the work. How would it make the API better?

The less in each class, the easier it is to digest for developers. We're already at 60+ methods for IWorkerContext, so it seems like a good opportunity to keep it from getting hairier. 

Things like debugging and testing are easier if there are smaller classes to work with. We know the intent of this handful of methods now, but months down the road, that might be a bit fuzzier since they live with dozens of other methods. If they are in a distinct interface from IWorkerContext, devs will immediately know they work together. 

It seems pretty simplistic, but it's clearer for me if I see 4/5 methods are tested instead of figuring out which of the missed methods in 43/66 tested methods are relevant.

grahamegrieve:
I don't follow. I think you're imagining that the worker context as a getProfileBuilder() operation, and that ProfileBuilder has the same operations that are drafted in IWorkerContext now, and that this is better for users to always have an extra method call to make?

grahamegrieve:
I can live with that. So like as I updated?

dotasek:
Yes, that looks good.

PR #1047, opened 2022-12-22
Profile Utilities Refactor 2

dotasek:
This refactor further breaks down path processing into distinct blocks and reduces the parameters being passed from method to method.

It also moves all profile classes into a dedicated package.


grahamegrieve (approved):
ok

PR #1048, opened 2022-12-23
Fix appending text in ProfileUtilities

qligier:
If the base text was null, appending a text in the derived element (using "...") would show a "null" before the text. This commit adds the method Utilities.appendDerivedTextToBase() to implement the correct behavior.

Example with:
```json
{
  "id": "MedicationStatement.implicitRules",
  "path": "MedicationStatement.implicitRules",
  "comment": "...All modifiers SHALL be documented in the profile",
  "max": "0"
}
```
2022-12-23 18 54 53 build fhir org ef2ff575624c


grahamegrieve (approved):
ok

Issue #1046, opened 2022-12-21
isBase64 is not available on Android (Use java.util.Base64)


vitorpamplona:
The new check for Base64 strings is not available on Android: https://github.com/hapifhir/org.hl7.fhir.core/blob/b67d9176fd8bb7e1eccec2cb5246ea5ebc3d4b54/org.hl7.fhir.dstu2/src/main/java/org/hl7/fhir/dstu2/model/Base64BinaryType.java#L102 

This is similar to https://github.com/hapifhir/hapi-fhir/issues/1441

Android's boot classpath overrides Apache Commons Codec with an older version than the [1.15](https://mvnrepository.com/artifact/commons-codec/commons-codec/1.15) referenced by this project. Google doesn't seem to be [enticed to fix this](https://issuetracker.google.com/issues/37033030). The available class is [here](https://android.googlesource.com/platform/external/apache-http.git/+/android-live-tv-l-mr1/src/org/apache/commons/codec/binary/Base64.java)

There are two possible solutions: 
1. Use `java.util.Base64` (added in Java 8 to replace the need for Apache Commons Codec)
2. Use the old `Base64.isArrayByteBase64(toCheck.getBytes())` method

Which one is preferred?

JPercival:
@jamesagnew @brynrhodes - This goes hand-in-hand with some hapi-fhir changes to support Android:
https://github.com/hapifhir/hapi-fhir/pull/4070

PR #1051, opened 2022-12-23
Swaps isBase64 checks into isArrayByteBase64 for Android compatibility

vitorpamplona:
Fixes: https://github.com/hapifhir/org.hl7.fhir.core/issues/1046


dotasek (changes requested):
After some discussion with @jamesagnew, there's an alternative implementation here: [Swaps isBase64 checks into isArrayByteBase64 for Android compatibility](https://github.com/hapifhir/org.hl7.fhir.core/pull/1079)

One is a touch more code than the other, but doesn't use a deprecated method.

I can be convinced either way, and would like an outside opinion on which to pick.

dotasek:
Fixed in https://github.com/hapifhir/org.hl7.fhir.core/pull/1079

Closing this PR.

PR #1059, opened 2023-01-06
Fix StructureDefinition40_50Test on Windows pipeline

dotasek:
On the pipelines, and possibly on some developer machines, git may be configured to introduce CR characters into downloaded files. This results in errors when a byte-by-byte compare is made in tests.

In one of these breaking tests, the default system lineSeparator is replaced by the single java newline `/n` character to prevent this from impacting the test.


grahamegrieve (approved):
ok

PR #1075, opened 2023-01-16
Test showing that extensions are not carried over

jkiddo:


grahamegrieve:
carumba, this one's a pig to deal with. I'm not sure it's a bug, in fact, it's a long standing design decision.

jkiddo:
Ehhm.... what?
This happens for enumeration elements only. Do it with eg. name and the extensions are carried over just fine.

grahamegrieve:
I think, from looking at the code, that it happens with primitive types

jkiddo:
Nope - works fine for e.g. booleans. Its a problem for all the castTo* called in the setProperty which do a `hasPrimitiveValue()` - and that seems ... well ... a bit inconsistent

grahamegrieve:
see https://chat.fhir.org/#narrow/stream/179167-hapi/topic/SetProperty.20.26.20primitives

grahamegrieve:
why close it? Might as well keep the test

PR #1078, opened 2023-01-17
FTPClient: upload file path creation, logging, and improved default path

dotasek:
This adds several improvements to the FTPClient:

1. When the path for a file to be uploaded doesn't exist, it is created.
2. Logging has been introduced to better diagnose problems
3. The path the client was created with is handled more robustly. FTP is sensitive to relative and absolute paths, particularly when changing the current working directory (CWD). This attempts to create a consistent reset to the CWD when a method might change it during execution.


dotasek:
Merging, after @grahamegrieve verified that it's working on Zulip.

PR #1092, opened 2023-01-25
Enhancements to StructureMap transform to support logical models as source and value set lookup

nkarip:
1. Implements org.hl7.fhir.r5.utils.structuremap.FFHIRPathHostServices.resolveValueSet (taken from AHDIS Matchbox).
2. Sets logical structure definition in input parser when input source is a logical model.

Current behavior for JSON inputs that are instances of a logical model is to look for resourceType, which will cause an exception if not present.

Enhancement extracts the input source type from the first group in the structure map, finds the Structure Definition by URI and sets ParserBase logical property to that Structure Definition if kind is logical. This bypasses the later check for resourceType during the transform execution.

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/structuremap/FFHIRPathHostServices.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/structuremap/FFHIRPathHostServices.java
@@ -96,7 +96,7 @@ public class FFHIRPathHostServices implements FHIRPathEngine.IEvaluationContext 
 
   @Override
   public ValueSet resolveValueSet(Object appContext, String url) {
-    throw new Error("Not Implemented Yet");
+	return structureMapUtilities.getWorker().fetchResource(ValueSet.class, url);
   }
 
 }
--- a/org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/utils/structuremap/FFHIRPathHostServicesTest.java
+++ b/org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/utils/structuremap/FFHIRPathHostServicesTest.java
@@ -1,0 +1,31 @@
+package org.hl7.fhir.r5.utils.structuremap;
+
+import java.io.IOException;
+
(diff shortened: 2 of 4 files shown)


dotasek (changes requested):
Is there a minimal example of the transformation this would enable? 

There appears to be some test coverage for this method already: src/test/java/org/hl7/fhir/r5/test/StructureMappingTests.java

Those tests appear to cover most of the code you added, but this portion of the code you added doesn't get covered:

```java
if (sourceSD.getKind() == StructureDefinition.StructureDefinitionKind.LOGICAL) {
    	parser.setLogical(sourceSD);
    }
```

There are also a places where an exception is thrown where it would be good to give an example of what could trigger such an exception.

```java
if (inp.getMode() == StructureMap.StructureMapInputMode.SOURCE)
	    if (type != null)
	      throw new DefinitionException("This engine does not support multiple source inputs");
```

dotasek:
In addition, a simple test for org.hl7.fhir.r5.utils.structuremap.FFHIRPathHostServices.resolveValueSet would be good as well.

nkarip:
Added tests - built the inputs locally rather than in fhir-test-cases since they aren't exactly meaningful examples, just designed to exercise the specific lines flagged above.

nkarip:
Sorry, I accidentally clicked close button, please ignore - this is still open.

dotasek:
I can see those tests hitting those lines, and it looks to me like it's correctly loading http://hl7.org/fhir/cdatest/StructureDefinition/ClinicalDocument as a logical model.

I do wonder if it's possible to expand testTransformLogicalSource to test something other than "{}" and include it in `r5/structureMapping` fhir-test-cases, though.

nkarip:
I would argue adding additional content defeats the purpose of a unit test. Right now the test targets the specific changes made. Adding additional content to the source exercises the entire map stack. I guess it depends on whether the intention is to isolate future failures or run integration tests.

grahamegrieve (approved):
ok

dotasek (approved):
Your last comment makes sense.

I did a little experimentation with the test you have to verify that changing the behaviour of ValidationEngine with regards to setLogical(sourceSD) will in fact cause a test failure, so it looks good to me.

PR #1096, opened 2023-01-30
Rework Package API to allow for authentication on package servers

grahamegrieve:


grahamegrieve:
@jamesagnew please review - I think this is a breaking change for HAPI, but it establishes the capacity for NPM package servers to require authentication - this is a request to support non-public infrastructure

PR #1118, opened 2023-02-16
Fix https for terminology server

dotasek:
Connecting to the terminology server via https was failing due to the processing in TerminologyClient.checkEndsWith

This adds tests for expected terminology servers hosted on tx.fhir.org and fixes the https issue.


grahamegrieve (approved):
ok

PR #1121, opened 2023-02-17
StructureDef: remove example label when no label is given

costateixeira:
Pending validation. Note that example.label is still  to be 1..1 but if a label is not present, the failure should not be on the renderer.


costateixeira:
@grahamegrieve is this ok?

grahamegrieve (approved):
ok

PR #1127, opened 2023-02-21
Add maven options to prevent closed connections

dotasek:
We are frequently experiencing closed connections during our builds, which means many fail inconsistently. 

This may be an instance of https://issues.apache.org/jira/browse/WAGON-486

This PR adds the recommended fix; if all pipelines cleanly pass, this is likely an appropriate fix.


dotasek:
Note that this option is present in the HAPI build also: 

https://github.com/hapifhir/hapi-fhir/blob/7827f65f2dcbc68f0d65739a1749762a2357d01a/release-pipeline.yml#L173

dotasek:
This PR was an attempt to fix the PR pipeline. https://github.com/hapifhir/org.hl7.fhir.core/pull/1133 succeeded at this, so this PR is no longer needed.

PR #1128, opened 2023-02-21
Break into build and test steps

dotasek:
We are frequently experiencing closed connections during our builds, which means many fail inconsistently.

This may be caused by the http connections being kept open over a long build process.

This PR splits the build and tests; if all pipelines cleanly pass, this is likely an appropriate fix.


dotasek:
This PR was an attempt to fix the PR pipeline. https://github.com/hapifhir/org.hl7.fhir.core/pull/1133 succeeded at this, so this PR is no longer needed.

PR #1132, opened 2023-02-22
Fix r4b FhirVersionEnum

dotasek:
Updates to the correct FHIR version.


grahamegrieve:
wow looks like a good idea to me

grahamegrieve (approved):
ok

PR #1135, opened 2023-02-24
Fix codecov reporting for PR pipeline

dotasek:
The PR pipeline is no longer reporting codecov code coverage.

This PR should fix that and reports should become available when it is ready to merge.


dotasek:
This was fixed in https://github.com/hapifhir/org.hl7.fhir.core/pull/1137

PR #1110, opened 2023-02-08
Implement R5 RDF format changes

balhoff:
Changes to Turtle examples as well as Turtle profile files to align with the proposed FHIR RDF spec changes.

PR #1141, opened 2023-02-27
Output literal datatypes for anyURI types in RDF examples

balhoff:
This is a small correction to the previous PR #1110 related to Jira ticket https://jira.hl7.org/browse/FHIR-39075.


dbooth-boston (approved):
Looks good to me.  Thanks for fixing this!

PR #1108, opened 2023-02-07
FML lexer source element delimiter support

nkarip:
FML mapping.g4 grammar indicates source element can be alphanumeric IDENTIFIER or DELIMITEDIDENTIFIER surrounded by doublequote or backtick.

Current behavior: src."-quote" in FML map file becomes "element": "\\"-quote\\"" in Structure Map, which causes input source key mismatch due to expected doublequotes.

Implemented fix: src."-quote" in map becomes "element": "-quote" (without escaping) in Structure Map, which allows input source JSON data {"-quote": "value"}.

PR #1173, opened 2023-03-16
FML lexer source element delimiter support (update)

dotasek:
Duplicates https://github.com/hapifhir/org.hl7.fhir.core/pull/1108, but with current code base and gentle renaming.

From original PR:

> FML mapping.g4 grammar indicates source element can be alphanumeric IDENTIFIER or DELIMITEDIDENTIFIER surrounded by doublequote or backtick.

> Current behavior: src."-quote" in FML map file becomes "element": "\"-quote\"" in Structure Map, which causes input source key mismatch due to expected doublequotes.

> Implemented fix: src."-quote" in map becomes "element": "-quote" (without escaping) in Structure Map, which allows input source JSON data {"-quote": "value"}.


dotasek:
This is partially a review and update of #1108

If this now passes tests, the code seems reasonable to me.

@grahamegrieve : does this change correctly follow the spec for Structure Mapping?

nkarip:
I think the naming change makes it more confusing. In the grammar "Identifier" and "Constant" are different tokens. There is already a "readConstant" method. Having a separate "readAsStringOrProcessedConstant" makes it seem like another method to read a "Constant" rather than the method to read an "Identifier" token. Other than that looks fine.

dotasek:
Merging after offline discussion with @grahamegrieve

PR #1193, opened 2023-03-28
Get IPackageProvider from test resource properties file

dotasek:
This gets a properties file from the resource folder of the test src and uses it to instantiate the IPackageProvider used in FilesystemPackageCacheManager.

Practically speaking, this means that if the `hl7.core.utilities.properties` file is present in the `src/test/resources/` folder, all tests will run with the FilesystemPackageCacheManager using whatever IPackageProvider is indicated in that file.

### DIFF
--- a/.github/workflows/bidi-checker.yml
+++ b/.github/workflows/bidi-checker.yml
@@ -23,6 +23,6 @@ jobs:
       # Run the check for bidi characters.
       - name: Check for bidi characters
         id: bidi_check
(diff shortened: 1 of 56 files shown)


dotasek:
Closing, as merging master into this PR has become too difficult for automatic merge.

Re-opened as: https://github.com/hapifhir/org.hl7.fhir.core/pull/1211 using a manual merge.

PR #1199, opened 2023-03-31
Make XhtmlNode validation case insensitive

jamesagnew:
Currently the XhtmlNode#validate() method expects tag names to be lowercase, and all attributes to be lowercase except `ID` which has to be upper case. This PR makes this all case insensitive.

It also uses sets instead of lists for better performance.

### DIFF
--- a/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/XhtmlNode.java
+++ b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/XhtmlNode.java
@@ -1,5 +1,6 @@
 package org.hl7.fhir.utilities.xhtml;
 
+import static org.apache.commons.lang3.StringUtils.defaultString;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
 /*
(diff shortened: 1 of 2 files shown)


grahamegrieve:
Why is this OK? The rules for xhtml are that the names are case sensitive

jamesagnew:
@grahamegrieve you make a good point. Withdrawn.

grahamegrieve:
I did think we'd have some discussion about it - why, for instance? And if we sets are faster than lists, we should still do that part?

jamesagnew:
Ah fair enough - The reason I abandoned this was that I decided not to use this method and take another approach for the thing I wanted to use it for.

I'm adding the rendered narrative to the HTML representation that HAPI spits out when you hit a FHIR resource directly in the browser. For example, see the narrative being displayed here: https://hapi.fhir.org/baseR4/Patient/591805

I'm worried about people abusing this though by putting sketchy HTML into resources they upload. My original plan was to use the XhtmlNode.validate method to make sure that no illegal tags were present. This came with two disadvantages though:
* It catches unacceptable tags and attributes, but doesn't catch people using CSS tricks to load external content, etc
* It choked on people with invalid casing on attributes. I found in testing this that lots of narratives people had uploaded to hapi.fhir.org used different casing for the `id` attribute for example - And I think it makes sense for a validator to reject that but it's less helpful when your use case is just security validation.

FWIW I ended up swapping this for using the [OWASP Java HTML Sanitizer](https://github.com/OWASP/java-html-sanitizer/) instead. It just strips anything that could be abused including links, dangerous CSS attributes, etc. so it's much better for this use case.

I could resubmit this PR without the case insensitivity..

grahamegrieve:
it would be good to get the PR without case sensitivity, but maybe it would be good to have it as an option anyway? 

Also, validate() is written to the serve the validator, which enforces the FHIR rules, and they don't have rules about CSS tricks to load external content. Is there something we missed there?

PR #1203, opened 2023-04-04
Add -fhir-settings param to Validator CLI

dotasek:
~~Add a parameter to load -api-key-file for use in the ValidationEngine~~

Add a centralized `FhirSettings` utlities for use in getting global settings like the following:

* Authorization keys
* Tool paths (Ruby, merge, etc)
* Temp directories

This can be set using several methods:

1. In the validator CLI: using the param `-fhir-settings`
2. From a Java system property called `fhir.settings.path`

`-fhir-settings` takes precedence over `fhir.settings.path`, and if neither are set, the default location `user.home/.fhir/ fhir-settings.json` is used.

Command-line executions MUST call the following method BEFORE any calls that might use FhirSettings:

FhirSettings.setExplicitFilePath(String path)

Otherwise, Utilities.path(...). which is used throughout the code may be inconsistent with regards to [tmp] substitutions. At present, if setExplicitFilePath is called after FhirSettings.getInstance() is called, it will throw an exception.

FHIRSettings will be able to be re-used for ig-publisher and fhir publisher. See the following related PRs:

* https://github.com/HL7/fhir-ig-publisher/pull/671
* https://github.com/HL7/kindling/pull/122


vadi2:
What are the API keys intended to be used for?

grahamegrieve (commented):
it seems to me that the .getInstance() call is private business to the settings object, and no something that the classes users should have to get involved with. Or when would they do something else?

dotasek:
> it seems to me that the .getInstance() call is private business to the settings object, and no something that the classes users should have to get involved with. Or when would they do something else?

getInstance() follows the singleton pattern with lazy instantiation.

We need it to be lazy, because CLI tools need to set the source file BEFORE it gets instantiated, and places like Utilities.path(...) could instantiate with the default setup, resulting in weird mixes of Utilities.path(...) behaviour. 

Were it not for that, there could just be a static final field instead of lazy instantiation. But the requirement is to be able to set it from CLI params, which can't happen statically.

grahamegrieve:
ok but the question wasn't about lazy instantiation, which makes sense, but about how it impacts on external consumers. It's not their business it's lazy?

dotasek:
I'm struggling a bit to see how we could hide that. The only thing I can think of is making the 'fields' statically accessible, and hiding the instantiation there:

```java

public static String getApiKey(String key) {
   FhirSettingsPOJO fhirSettings = getInstance().fhirSettings;
   return fhirSettings.getApiKeys() != null
      ? fhirSettings.getApiKeys().get(key)
      : null;
}
```

So the end user would be able to just call FhirSettings.getApiKey("myKey") instead of FhirSettings.getInstance().getApiKey(). 

Is that what you had in mind?

grahamegrieve:
Yes I think that is - a set of facade methods to hide the details

dotasek:
> Yes I think that is - a set of facade methods to hide the details

Updated. Let me know if this looks correct now.

grahamegrieve (approved):
yes I like it now! thx.

Issue #1206, opened 2023-04-06
HumanName.getNameAsSingleString() returns an empty string as opposed to returning the text element


amir-kt:
https://github.com/hapifhir/org.hl7.fhir.core/blob/c4f8b65ba448bfdb594d0ff6db57260f6c5b84ce/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/model/HumanName.java#L681

HumanName.getNameAsSingleString() calls the get function for the different parts of the name and adds them to the list. This is fine for all the name parts that are Collections because nothing is added to the nameParts list for an empty collection.
But for the getFamilyElement() a new StringType is created and added to the nameParts list, hence the function never actually has an empty nameParts and will return "" instead of the text element's value.

amir-kt:
opened a PR: https://github.com/hapifhir/org.hl7.fhir.core/pull/1207

PR #1207, opened 2023-04-06
Fix HumanName getNameAsSingleString never returning text element

amir-kt:
fixing https://github.com/hapifhir/org.hl7.fhir.core/issues/1206

### DIFF
--- a/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/model/HumanName.java
+++ b/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/model/HumanName.java
@@ -678,7 +678,9 @@ public class HumanName extends Type implements ICompositeType {
     List<StringType> nameParts = new ArrayList<StringType>();
     nameParts.addAll(getPrefix());
     nameParts.addAll(getGiven());
-    nameParts.add(getFamilyElement());
+    if (hasFamilyElement()) {
+      nameParts.add(getFamilyElement());
+    }
     nameParts.addAll(getSuffix());
     if (nameParts.size() > 0) {
       return joinStringsSpaceSeparated(nameParts);


dotasek (changes requested):
This fix looks fine. 

Please review the additional changes in [this branch](https://github.com/hapifhir/org.hl7.fhir.core/tree/amir-kt/master), confirm that they are correct, and merge them into this PR.

amir-kt:
merged [your branch](https://github.com/hapifhir/org.hl7.fhir.core/tree/amir-kt/master) with this PR. I removed a line in the unit test that defined an unused variable. 

I also fixed the same issue in r5 and added your unit tests to it too. I'm not sure what the other environments are or whether you'd like to implement the fix there too.

dotasek:
Thank you @amir-kt for merging that test. It was also a very good idea to check r5; that's something I should have noticed. I will look at the other FHIR versions and update them accordingly.

dotasek (changes requested):
@amir-kt I checked the previous versions and did the updates where applicable:

https://github.com/hapifhir/org.hl7.fhir.core/tree/amir-kt/master

Could you review these changes, merge them into this request, and then we can merge this into master and close.

dstu2016may passed without any alterations because that field is automatically instantiated in getFamily() in that version.

dstu2 has no getNameAsSingleString method.

amir-kt:
Awesome, reviewed and merged. Thanks @dotasek!

dotasek:
@amir-kt I'm going to merge this PR once we repair our pipeline issues, but in reviewing it with @grahamegrieve, he pointed out that this construction is not necessarily ideal.

We will follow up by modifying the logic as described in this issue: https://github.com/hapifhir/org.hl7.fhir.core/issues/1214

Please take a look at that issue and comment, as it will change the behaviour we defined in this PR.

dotasek:
I've run this without the tests that are failing on Master on a Windows machine. I'm going ahead with the merge.

PR #1210, opened 2023-04-10
Bump JaCoCo Version

dotasek:
Bumps the JaCoCo version (maybe fixing our STDOUT errors in pipelines?) and fixes the reports pom.xml, which improperly set JaCoCo to a version out of sync with the rest of the project.


dotasek:
Superseded by https://github.com/hapifhir/org.hl7.fhir.core/pull/1213

PR #1211, opened 2023-04-10
Get IPackageProvider from test resource properties file

dotasek:
This gets a properties file from the resource folder of the test src and uses it to instantiate the IPackageProvider used in FilesystemPackageCacheManager.

Practically speaking, this means that if the hl7.core.utilities.properties file is present in the src/test/resources/ folder, all tests will run with the FilesystemPackageCacheManager using whatever IPackageProvider is indicated in that file.

This replaces https://github.com/hapifhir/org.hl7.fhir.core/pull/1193, which had become too difficult to merge automatically.


dotasek:
This PR will no longer be needed, as TestingUtilities.injectCorePackageLoader() will be removed in Grahame's recent work.

PR #1218, opened 2023-04-14
Bump logback

dotasek:
Bumping logback in hopes that maybe STDOUT errors on Windows Java 11 pipeline will finally disappear.


dotasek:
This is passing all tests, but timing out on Windows instead of failing due to STDOUT. I'm merging this to get PRs to at least not fail due to errors.

PR #1223, opened 2023-04-17
Return ID part when FhirPath expression result is converted to string

jamesagnew:


grahamegrieve:
@jamesagnew so this is the only case of this kind of behaviour?

jamesagnew:
@grahamegrieve Well, it's definitely not the only spot we need to do this, but it's the only spot I could find that still needed it. Here's an example of pretty much the same issue from the GraphQL engine: https://github.com/hapifhir/org.hl7.fhir.core/blob/master/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/GraphQLEngine.java#L587

grahamegrieve:
ok. you interpreted that differently to what I meant. I'm glad there's nowhere else that we need to make that id change, but what I was asking is, there's no other fields/types where this issue arises?

jamesagnew:
@grahamegrieve ah, got it. No, thanksfully IdType is the only type with the weird overloaded behaviour. 

The Meta fields were also like this in the old UHN Dstu1/2 structures but you (thankfully) convinced me that was bad when we moved to the hl7.org structures.

grahamegrieve (approved):
thank, James. Approved

PR #1225, opened 2023-04-18
Update fhir-path-tests.xml

jkiddo:
Added consecutive FHIRpath tests

### DIFF
--- a/org.hl7.fhir.r4/src/test/resources/patch/fhir-path-tests.xml
+++ b/org.hl7.fhir.r4/src/test/resources/patch/fhir-path-tests.xml
@@ -1719,5 +1719,83 @@
         </managingOrganization>
       </Patient>
     </output>
-  </case>  
+  </case>
+  <case name="Consecutive operations" mode="both">
+    <input>
+        <Patient xmlns='http://hl7.org/fhir'>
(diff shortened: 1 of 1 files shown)


dotasek:
Is this code dead?

Where are these tests actually executed? The only reference I can find to them is in testcases.ini in the FHIR spec code (https://github.com/HL7/fhir/blob/master/tests/testcases.ini), which in turn seems to be referenced nowhere. 

The Git blame in these is quite old, some of it dating back to 2017, and there are commits around it with wording like "[Remove IG publisher (it's gone to Maven)](https://github.com/HL7/fhir/commit/56ceb0df08dd72381a4f98f1ea4ac5ea6d139a8b)"

grahamegrieve:
yes - these changes should be made to https://github.com/FHIR/fhir-test-cases/blob/master/r4/patch/fhir-path-tests.xml

jkiddo:
Oversight from my side - I will provide a new PR. Why not delete the old file?

grahamegrieve:
I will

jkiddo:
@grahamegrieve -> https://github.com/FHIR/fhir-test-cases/pull/133

Issue #1212, opened 2023-04-11
SubscriptionTopic and SubscriptionStatus converters for VersionConvertorFactory_43_50


dotasek:
There are no SubscriptionTopic or SubscriptionStatus converter available for VersionConvertorFactory_43_50. These should be added.

This was requested by @fil512 to make life easier in HAPI-FHIR

fil512:
Examples:

```
{
  "resourceType": "SubscriptionTopic",
  "id": "1",
  "meta": {
    "versionId": "1",
    "lastUpdated": "2023-04-13T19:11:45.601-04:00"
  },
  "url": "http://example.com/topic/test",
  "status": "active",
  "resourceTrigger": [ {
    "resource": "Encounter",
    "supportedInteraction": [ "create" ],
    "queryCriteria": {
      "previous": "Encounter?status=planned",
      "current": "Encounter?status=finished",
      "requireBoth": true
    }
  } ]
}
```
{
  "resourceType": "SubscriptionStatus",
  "status": "active",
  "type": "event-notification",
  "eventsSinceSubscriptionStart": "1",
  "notificationEvent": [ {
    "eventNumber": "1",
    "focus": {
      "reference": "Encounter/3"
    }
  } ],
  "subscription": {
    "reference": "Subscription/2"
  },
  "topic": "http://example.com/topic/test"
}
```

```

fil512:
Those are both R4B if it matters

fil512:
I guess you can tell from the eventNumber

dotasek:
Implemented in: 
https://github.com/hapifhir/org.hl7.fhir.core/pull/1226
https://github.com/hapifhir/org.hl7.fhir.core/pull/1243

PR #1226, opened 2023-04-18
Add SubscriptionTopic and SubscriptionStatus conversions for R4B <-> R5

dotasek:
Adds feature for #1212

Note that this conversion only covers the minimal examples provided by the R5 spec, and by @fil512. 

This conversion will also be inherently lossy where fields are not 1-to-1 matches (as in SubscriptionSearchModifier).

### DIFF
--- a/org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/advisors/impl/BaseAdvisor_43_50.java
+++ b/org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/advisors/impl/BaseAdvisor_43_50.java
@@ -31,7 +31,10 @@ public class BaseAdvisor_43_50 extends BaseAdvisor50<org.hl7.fhir.r4b.model.Exte
       return true;
     } else if (lastPath.equals("Basic") && url.startsWith("http://hl7.org/fhir/5.0/StructureDefinition/extension-Requirements.")) {
       return true;
-    } else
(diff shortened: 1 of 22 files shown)


grahamegrieve (commented):
this:

if (src.hasEventsSinceSubscriptionStart()) {
       org.hl7.fhir.r4b.model.StringType value = new org.hl7.fhir.r4b.model.StringType();
       value.setValue(src.getEventsSinceSubscriptionStartElement().getValueAsString());
       tgt.setEventsSinceSubscriptionStartElement(value);
     }

why does isn't follow the normal pattern? This will drop extensions, and so it will eventually need to be redone. This PR should be redone using the normal patterns (and there's a spelling mistake - covert)

dotasek:
> why does isn't follow the normal pattern? 

What is the normal pattern for this? 

EventsSinceSubscription is not an extension. All that appears to differ is the type (String in one, integer64 in the other). Is there a place in the existing code following the pattern you'd like?

Is this the only place in this conversion that this is incorrect?

grahamegrieve:
oh? that's unfortunate that's the first I looked at. We do it like this: org.hl7.fhir.convertors.conv40_50.datatypes40_50.primitive40_50.String40_50.convertStringToMarkdown(StringType) - pick one of the two types, and add a conversion there. In this case, I'd add a convertStringToInteger to String since the type Integer64 doesn't exist at this time

dotasek:
One @grahamegrieve's recommendation, I'll be expanding this conversion to cover the full resources, not just what is required of the examples.

grahamegrieve (approved):
ok

PR #1227, opened 2023-04-20
Add appropriate SimpleContext for StructureMapUtilities instances

dotasek:
Possible fix for https://github.com/hapifhir/org.hl7.fhir.core/issues/1219

PR #1230, opened 2023-04-23
Fix IG Loader problems where map conversion no longer works with a 'n…

lmckenzi:
…ull' context

### DIFF
--- a/org.hl7.fhir.dstu3/src/main/java/org/hl7/fhir/dstu3/context/SimpleWorkerContext.java
+++ b/org.hl7.fhir.dstu3/src/main/java/org/hl7/fhir/dstu3/context/SimpleWorkerContext.java
@@ -140,6 +140,11 @@ public class SimpleWorkerContext extends BaseWorkerContext implements IWorkerCon
     return res;
   }
 
+  public static SimpleWorkerContext fromNothing() throws FileNotFoundException, IOException, FHIRException {
+    SimpleWorkerContext res = new SimpleWorkerContext();
+    return res;
+  }
+
   public static SimpleWorkerContext fromPackage(NpmPackage pi, boolean allowDuplicates) throws FileNotFoundException, IOException, FHIRException {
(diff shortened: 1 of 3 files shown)


dotasek:
This is essentially the same fix as this WIP: https://github.com/hapifhir/org.hl7.fhir.core/pull/1227

So, this also applies (from @grahamegrieve):

> But the context is empty. So it just kicks the can down the road?

Exploring the code a bit to see how SimpleWorkerContext is initialized for r5 (the default for ValidationEngine), it looks like what's missing can be found here: https://github.com/hapifhir/org.hl7.fhir.core/blob/256964cfc2d2117b8fd88835c0684c7e6fe42789/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ValidationEngine.java#L318 

It might be that we have to create a per-version builder that follows this, otherwise the context will be missing some necessary setup.

lmckenzi:
Context doesn't seem to be necessary for the mapping transformation to run.

dotasek:
@lmckenzi I did notice that. 

When I did my parallel fix, I ran through [KMCGit](https://github.com/KMCGit)'s example, but since 5.6.81 no longer works due to new packages, I had nothing to compare the result to. 

However, the step we're editing is common to all validation tasks: loadIgsAndExtensions() is called whenever the ValidationEngine is instantiated. So if we make a change here, it's not limited to mapping transformation.

I'm not against adding this fix in, because it fixes the immediate break, but I feel we should investigate what else we're impacting.

grahamegrieve:
Context is totally necessary to run a transformation

lmckenzi:
Execute a map, yes.  Convert an FML file to a StructureMap, no.

grahamegrieve (approved):
I'm going to approve this, but it doesn't make sense to me - the context must be needed for something, and I very much suspect that something isn't being processed correctly with an empty context

PR #1242, opened 2023-04-28
Fixed typoes in Messages.properties and updated Messages_nl.properties

ahenket:
As per DM Zulip discussion with Grahame


grahamegrieve (approved):
awesome

PR #1256, opened 2023-05-08
Refactor extension ignore in conversion

dotasek:
In the present conversion code, when an extension is encountered during a copyDomainResource method call, it goes to the BaseAdvisor class for the relevant versions (BaseAdvisor_30_40, BaseAdvisor_14_40, etc) and then decides whether or not to ignore it. 

This means it's not obvious which extensions are being ignored within a particular resource convertor . CapabilityStatement30_40 for example, handles an extension for `acceptUnknown`, but to prevent that conversion from copying the extension to the extensions field after converting it to an acceptUnknown field, you must know to look in BaseAdvisor_30_40. 

This PR applies the same pattern used in the copyElement method, which includes a list of URLs to ignore.

It also includes additional tests to cover resource conversions that were not yet covered.

Note: In 10_30 and 14_30 conversions for CapabilityStatement/Conformance, the ignore statements were unnecessary; dstu2, dstu2016may, and dstu3 all have acceptUnknown elements, thus the extension would not be present in any of these resources.


grahamegrieve (approved):
love it

PR #1259, opened 2023-05-12
Fix problem when profiling a base type that supports both Reference a…

lmckenzi:
…nd canonical

See the https://github.com/chn100/icd-breaker-PlanDef-check project for a simple example that creates the QA problem.

What was happening is that when checking the Reference type, 'ok' was getting set to true.  Then when checking the third data type (canonical), because ok was true, the code was checking for a type match on *canonical* even though the data type was Reference, and it wasn't finding a match.  The new code separates out "ok" for an overall match from whether there is a type match and fixes the problem.


grahamegrieve (approved):
ok

grahamegrieve (approved):
Is there a test case for this?

lmckenzi:
I included a reference to a simple IG that creates the problem above.

grahamegrieve:
For a change like this, we need a test case

lmckenzi:
Test case created: https://github.com/FHIR/fhir-test-cases/pull/137

Issue #990, opened 2022-11-16
German translations of plural phrases required


dotasek:
There are a number of phrases in Messages_de.properties that need to be corrected in order to manage plural cases.

At the moment, each of these is tagged with #REPARIERE MICH PLURAL, and the existing phrase has simply been copied. An appropriate singular phrase (`_one`) and plural phrase (`_other`) should be put in place.

```
#REPARIERE MICH PLURAL
Bundle_BUNDLE_Entry_Type3_one=Der type "{1}" ist nicht g\u00fcltig - muss einer von {2} sein
Bundle_BUNDLE_Entry_Type3_other=Der type "{1}" ist nicht g\u00fcltig - muss einer von {2} sein
```

https://github.com/hapifhir/org.hl7.fhir.core/blob/dbef06873089dfc1f3bbf4e97a8f7577c77925de/org.hl7.fhir.utilities/src/main/resources/Messages_de.properties#L18

Note that there are a number of phrases  where the singular case (`_one`) is left intentionally blank. The comment explaining why is in English. It would be more appropriate if these comments were in German as well.

```
#The following error cannot occur for a single item. _one case left intentionally blank.
#Profile_based_discriminators_must_have_only_one_type__in_profile_one=
Profile_based_discriminators_must_have_only_one_type__in_profile_other=Profilbasierte Diskriminatoren d\u00fcrfen nur einen Typ ({1} im Profil {2}) haben.
```

patrick-werner:
currently on it. Will provide translations for all open  phrases

PR #1263, opened 2023-05-15
Pw/translations de

patrick-werner:
added missing german translations


patrick-werner:
#closes #990

dotasek:
This is actually quite difficult to review, as many keys and values have changed.

To make this easier, I am working on developing some unit tests to do several things:

1. To report how many message definitions from I18nConstants are covered in each language, so we can track whether coverage has increased/decreased in a particular commit.
2. To confirm that plural phrases are completely constructed in each language.

Please be patient as I build this test; I will use this PR to help refine it.

dotasek:
I added a test that reports any missing property keys, and also verifies that all plural phrases are present.

There were a few fixes that needed to be done in other languages, but the original updates in this PR all seem valid.

patrick-werner:
> I added a test that reports any missing property keys, and also verifies that all plural phrases are present.
> 
> There were a few fixes that needed to be done in other languages, but the original updates in this PR all seem valid.

thanks for the quick review, adding automated translation coverage checks seems to be great.
I noticed that some keys are not used anymore, or some phrases only exist in e.g. spanish. This might be another great area to automatically test.

grahamegrieve (approved):
ok

Issue #1264, opened 2023-05-18
Add R5 TestPlan conversions to R3, R4 and R4B


rettema:
In support of the fhir-ig-publisher issue https://github.com/HL7/fhir-ig-publisher/issues/618, we need conversion classes for the new R5 TestPlan conformance resource to R3, R4 and R4B.

rettema:
PR submitted - https://github.com/hapifhir/org.hl7.fhir.core/pull/1265

rettema:
PR #1265 merged. Closing this issue.

PR #1265, opened 2023-05-18
#1264 add R5 TestPlan conversion classes to R3, R4 and R4B

rettema:
R5 TestPlan conversion classes added for R3, R4 and R4B
R5 TestPlan renderer class added
R5 Requirements renderer class updated with minor enhancements


grahamegrieve (commented):
What's with the % coverage changes in the Requirements renderer - how does that work?

rettema:
Trying to show a meaningful connection between the statements and how many of them have corresponding artifacts that "satisfy" (provide coverage of) those statements. I realize that it's extra calculated information not actually stored as part of the Requirements resource instance and can remove it if you feel it's not appropriate for the renderer.

grahamegrieve:
I think we should remove it because it's misleading - most requirements will be satisfied elsewhere. It's an unusual case for the requirement to post-documentation (which is when you fill out the satisfiedBy element)

rettema:
Understood. Anything else before I commit that change?

PR #1272, opened 2023-05-23
Bundle.getLink(String) comparing string to enum

### LINKED ISSUE #1266: R5 Bundle.getLink(String) does not work because String is compared to an enum
The R5 version of [`Bundle.getLink(String)`](https://github.com/hapifhir/org.hl7.fhir.core/blob/master/org.hl7.fhir.core.generator/configuration/Bundle.java#L16) always returns null. It looks like the problem is `BundleLinkComponent.getRelation()` used to return a `String`, but now returns a `Bundle.LinkRelationTypes` enum value, so the string comparison will always fail. I believe the comparison needs to be made against `next.getRelation().toCode()` as is done in `BundleEntryComponent`.

pnewhook:
Compare the string value of a LinkRelationTypes enum in Bundle.getLink(String)

Fixes #1266


grahamegrieve (approved):
ok

PR #1279, opened 2023-05-26
copy Element children to prevent value modification when validating

glichtner:
# Problem (short)
Running the IG Publisher with validation (implemented in org.hl7.fhir.core) results in unintended modifications of resource instance values in the resulting implementation guide.

# Problem (long)
During validation, when validating invariants for the Range datatype, the values of range.low and range.high fields are changed by the [Utilities.lowBoundaryForDecimal()](https://github.com/hapifhir/org.hl7.fhir.core/blob/233d0ecf86f7a93cc2be2d8b8c7f4056ba0c9ce3/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/Utilities.java#L1550) / [Utilities.highBoundaryForDecimal()](https://github.com/hapifhir/org.hl7.fhir.core/blob/233d0ecf86f7a93cc2be2d8b8c7f4056ba0c9ce3/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/Utilities.java#L1672) functions. To prevent changes of the original resource, the elements are [copy()'d before](https://github.com/hapifhir/org.hl7.fhir.core/blob/233d0ecf86f7a93cc2be2d8b8c7f4056ba0c9ce3/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/FHIRPathEngine.java#L4018), but this function [does not create copies of the Element.children ArrayList entries](https://github.com/hapifhir/org.hl7.fhir.core/blob/233d0ecf86f7a93cc2be2d8b8c7f4056ba0c9ce3/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/Element.java#L1371), which include the range.low.value field. Therefore, the value changes by Utilities.lowBoundaryForDecimal() are carried over in the resource instances in the actual implementation guide.

# Reproduce
Create a sushi project, change `fhirVersion` to 5.0.0 in `sushi-config.yaml` and create an `input/fsh/example.fsh` file with the following content:
```
Instance: ProcedureExample
InstanceOf: Procedure
Description: "Example procedure"
* status = #unknown
* subject.display = "subject"
* occurrenceRange
  * low = 40 's'
  * high = 41 's'
```

## Expected output in the IG after running the ig publisher
```json
...
  "occurrenceRange" : {
    "low" : {
      "value" : 40,
      "system" : "http://unitsofmeasure.org",
      "code" : "s"
    },
...
```
Note: This is the output after running sushi (i.e. in `fsh-generated/resources`) , but NOT after running the ig publisher (in `output/`), see below.

## Actual output in the IG
```json
...
  "occurrenceRange" : {
    "low" : {
      "value" : 39.50000000,
      "system" : "http://unitsofmeasure.org",
      "code" : "s"
    },
...
```

## Solution
Copy the actual elements in element.children.

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/Element.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/Element.java
@@ -1368,7 +1368,9 @@ public class Element extends Base {
     dest.value = value;
     if (children != null) {
       dest.children = new ArrayList<>();
-      dest.children.addAll(children);
+      for (Element child : children) {
+        dest.children.add((Element) child.copy());
+      }
     } else {
       dest.children = null;
     }


dotasek (changes requested):
This change looks reasonable to me. 

Is there a way to add a test to this so it doesn't regress later?

grahamegrieve:
I don't understand this error report. looking at the mentioned routines in Utilities, they don't have any access to the elements in question, so the discussion completely doesn't justify the rather significant change being proposed. Something is missing from the report

glichtner:
The Utilities.lowBoundaryForDecimal function is called by FHIRPathEngine.funcLowBoundary which is called by FHIRPathEngine.evaluateFunction for the validation of the rng-2 constraint of the [Range datatype](http://hl7.org/fhir/R5/datatypes.html#Range) (`low.value.empty() or high.value.empty() or low.lowBoundary().comparable(high.highBoundary()).not() or (low.lowBoundary() <= high.highBoundary())`). The FHIRPathEngine.funcLowBoundary [copies the element and then sets the value (for low) to the return value of Utilities.lowBoundaryForDecimals: ](https://github.com/hapifhir/org.hl7.fhir.core/blob/233d0ecf86f7a93cc2be2d8b8c7f4056ba0c9ce3/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/FHIRPathEngine.java#L4019), which is where the original child element's value is overwritten (and later written to instances in the implementation guide as shown in my example).

grahamegrieve:
Thank you that makes more sense. @dotasek this is not really something that's easy to test in a unit test. This code:

```java
      Base v = base.copy();
      v.setProperty("value", new DecimalType(Utilities.lowBoundaryForDecimal(value, precision == 0 ? 8 : precision)));
      result.add(v);
```

certainly anticipates that the copy works the way that @glichtner's PR makes it work. 

Eclipse finds 6500 references to this method, and I'm not going to review them all, but a quick scan makes me wonder how many other bugs are caused by this. I'm going to accept this PR

Issue #1288, opened 2023-06-03
Validator CLI: 'Can't find %s profile %s but trying to go on' even though profile is actually found and applied


DarthGizka:
Consider the following tiny resource (boiled down to its essence from a much bigger production resource):

```xml
<Identifier xmlns="http://hl7.org/fhir">
	<system value="http://fhir.de/sid/arge-ik/iknr"/>
	<!-- this is supposed to be a string consisting of 8 or 9 digits -->
	<value value="w0rgn"/>
</Identifier>
```
Validating this results in a bunch of mysterious messages:

```
FHIR Validation tool Version 6.0.11 (Git# 44db9547439f). Built 2023-06-02T14:25:08.722Z (25 hours old)
  Java:   17.0.7 from C:\Program Files\Java\jdk-17 on amd64 (64bit). 8120MB available
  Paths:  Current = D:\dev\FHIR\Beispieldaten\Identifier, Package Cache = C:\Users\KnightOwl\.fhir\packages
  Params: -version 4.0.1 -ig de.basisprofil.r4#1.3.2 -profile http://fhir.de/StructureDefinition/identifier-iknr|1.3.2 Identifier_sid_arge_ik_iknr.xml
  Locale: United States/US
  Jurisdiction: United States of America
Loading
  Load FHIR v4.0 from hl7.fhir.r4.core#4.0.1 - 4576 resources (00:03.955)
  Load hl7.terminology#5.1.0 - 4180 resources (00:01.240)
  Load hl7.terminology.r5#5.0.0 - 4174 resources (00:00.808)
  Load hl7.fhir.uv.extensions#1.0.0 - 1328 resources (00:00.016)
  Terminology server http://tx.fhir.org - Version 2.1.0 (00:01.074)
  Load de.basisprofil.r4#1.3.2 - 177 resources (00:00.162)
  Package Summary: [hl7.fhir.r4.core#4.0.1, hl7.fhir.xver-extensions#0.0.12, hl7.terminology#5.1.0, hl7.terminology.r5#5.0.0, hl7.fhir.uv.extensions#1.0.0, de.basisprofil.r4#1.3.2]
  Get set... Can't find Identifier profile http://fhir.de/StructureDefinition/identifier-iknr but trying to go on
Can't find Identifier profile http://fhir.de/StructureDefinition/identifier-kvid-10 but trying to go on
[...]

PR #1291, opened 2023-06-05
Refactor Validator CLI params

dotasek:
This PR refactors the the processing of Validator CLI Params in order to make the addition of future command line tasks easier.

This work encapsulates the validator's modes of operation into distinct `Task` classes.

Each Task class defines:

1. The code to be executed by the task.

2. The conditions required for the validator to use the task (normally, is there a param like '-compile' present to trigger a compile task). ValidatorCLI iterates across all available tasks and picks the first that returns true via shouldExecuteTask(), or the default task (ValidateTask)

3.  The command-line help to provide instructions for the user. This is split from the original (very long) document. Tasks that are not presently documented are marked as 'hidden' and not displayed

Tasks are divided into ValidationEngineTasks, which require a ValidationEngine to be built from CliContext, and SimpleTasks, which only require CliContext and the cli params[] object.


DarthGizka:
The 'let's ask all tasks until we find one who likes the current set of parameters' scheme seems to suffer a bit from big-ball-of-mud syndrome. It might be wiser to analyse both the current state of affairs in `Params.loadCliContext()` and `ValidatorCli.doValidation()` on one hand and 50+ years of successful industry-wide practice on the other, in order to arrive at a draft list of requirements to drive the design.

One feature of successful CLIs is the clear separation between 'commands' (or operation modes) on one hand and a big pool of switches on the other. Not all of the switches apply to all commands but many switches apply equally to many commands. I.e. whenever there is FHIRy output we can choose between JSON and XML, minified or prettified output, output to named specific file or to filename derived by appending a suffix, and so on, regardless of whether we're validating to `OperationOutcome` or snapshotting a `StructureDefinition`. This orthogonality is important in order to reduce complexity. Of command switches there can only be one per invocation; they are mutually exclusive. No command switch means 'main mode of operation', which is 'validate'.

The current set of commands/switches has grown to its current state over time with a series of hacks and patchings; it is very irregular and it has probably never been analysed in toto (at least not by people within the project), with an eye towards what it *should* look like, both for ease of understanding/use and ease of maintenance. 

An example for lack of orthogonality is `-ig`. When validating I can pass a filename for a .tgz package instead of a package id (package#version) but with `-install` this bombs. Never mind that `-install` doesn't do anything useful ATM; it is just an example.

There are two easy roads to success. The one taken by Torinox is a folder full of class files where each class represents a command and is pulled into the infrastructure by way of an annotation. This is well worthy of study - e.g. drop `~/.dotnet/tools/.store/firely.terminal/3.1.0/firely.terminal/3.1.0/tools/net6.0/any/Firely.Terminal.dll` on LINQPad and then select [Open all in ILSpy], then navigate to namespace `Firely.Terminal.Commands` and behold beauty. However, while the dispatch infrastructure is beautiful, the parameter processing is simply awful. A successful CLI *must* allow switches to appear in any order; if it doesn't then it becomes a bloody pain to use for the varied, complicated, multi-switch things the validator CLI is built for. Firely's cowboy style of operation - don't check for errors, so you never have to handle and report any - is also very much *not* recommended. My latest pet peeve is that they started snapshotting/expanding everything that they install in the machine-wide package cache *without a switch to turn that nonsense off*. Maybe they should have tested their snapshot feature before prescribing it for all and sundry. ATM they cannot even install a core package like R4 without breaking it, let alone the weird stuff that this normative for electronic prescriptions in Teutonia.

The other easy road to success is the opposite of the Firely free-for-all: a curated Java enum as the central pillar (or rather two: one for commands, one for switches). This is one of very few things that Java does better than all other languages/platforms, so why not capitalise on it? ;-) Each definition in the central command enum can hold the name of the switch, perhaps a one-letter abbreviation for busy people, and a shortish description (not more than one or two lines) that gets dumped out on `-help`. This means that the project lead can see what's what by pulling up a single source file or at most two, instead of having to write programs to gather the information or invoke the program with `--help`.

A fully curated scheme might also have lists of applicable switches (which are in another enum) for each command and perhaps a separate list for 'universal' switches without clear ties to any commands (i.e. log verbosity level, debug switch and so on).

Very useful in practice is to have `--help`, `-h` and `-?` punch through any assortment of parameters and switches, so that one can stop typing a command at the 100th character, append ' -?' and refresh one's memory on switch or command syntax without having to scrounge up a separate console or terminal, recall the command line, the back-erase the ` -?' and continue crafting whatever spell one was working on.

The individual commands will then have comparatively few specific parameters (both switchy and plain). 

*Note: I've analysed the command/parameter structure of the CLI several times over the years because I wanted to stick the validator behind an HTTP API while retaining the full power of the CLI and full compatibility with it (including the ability to simply pass a CLI command line). Other features were loading CLI jars dynamically on demand (by specifiying a version like 6.0.15) and intelligent, multithread-safe caching of profile packages and validation contexts. I never had the time to finish that thing for its original purpose but the bulk of the project got recycled for wrapping the ABDA/DAV reference validator; this has been in production in the central cloud for our cooperative (i.e. for all participating companies) for over a year now. I haven't given up on the CLI jar HTTP wrapper thing yet, though, because I want to use `validator_cli.jar` as master validator for all my tests (like double-checking ABDA/DAV validator tests and vetting all profile packages that are normative in our field) and for that I need both the throughput and the decoupling/flexibility that the HTTP API provides.*

Just imagine for a moment that you wanted to stick the validator behind an HTTP API, or behind a web frontend, or behind a slick GUI frontend. Even if you don't actually want to do that - or at least not now - the mental exercise can help you see more clearly the lie of the land. Command/parameter passing is only one facet; another is that a lot more goes to the console than is reflected in the OperationOutcome; this information can be quite important but it can only be captured in completely unstructured form via `stderr` and/or `stdout`. For some unfathomable reason it seems fashionable in Java-land for everybody and their aunt to instantiate a logger - and then stick that logger in a static class field. It is *impossible* to do worse than that. How would you separate the output for one request/thread/client from that for another? The whole validation logic/infrastructure must be thread-confined anyway in order to prevent breakage, and so it will only work on one request at any one time. Even if the output is not destined for the client, it must retain its association with a given request (via a correlation id like global HTTP request number) even for display on an operator console or output to a log file. Otherwise you'd have output from 50 different threads interspersed and no hope of understanding what's going on. The CLI will likely remain single-threaded (or single-request) but classifying/tagging/codifying the output can help make it machine-processable. Think test suites. You'd want to know if the validator suddenly started emitting unexpected message like in issue #1288, wouldn't you?

dotasek:
@DarthGizka there's a lot to process in that comment, and I certainly appreciate the feedback.

I enthusiastically agree with the suggestion of moving to distinct 'commands' as opposed to the weird overlap of params that currently define the CLI's behaviour.

However, this particular refactor was intended to be a quick one, without breaking changes. It was meant to reduce the complexity of adding new functionality, at  @grahamegrieve's request. As you've no doubt observed, in our previous state, the process of doing so was spread out in numerous places, and hard to grasp. The PR includes tests to ensure that we're not breaking existing behaviours, even if those behaviours have non-intuitive logic, and moves this logic (for now) into more clearly defined places.

I propose this:

1. I will go through this PR and make some nomenclature changes (switching from 'Task' to 'Command' is something that comes to mind) to indicate the direction we intend to go. If you can think of any other changes that would be prudent, please comment.
2. We should open an issue for a thorough refactor following @DarthGizka's suggestions. Changes to the CLI will require notifying the user community. With the suggested changes, people's workflows will break, and we should give them warning. 

@DarthGizka @grahamegrieve does this sound reasonable?

DarthGizka:
That certainly sounds perfectly reasonable. ;-)  I mostly wanted my remarks to be a reminder that any extra effort put into some change (i.e. beyond what is immediately necessary and absolutely indispensable) should be commensurate to the degree that the resulting structure fits into your where you project is headed and can support the envisioned future structure. The reason is that there are never enough resources for project like this anyway, and so any extra effort you invest should only go towards the future and not towards the past. If you can't see your proposed structure as a pillar of future structure or at least as a stepping stone towards it (e.g. enabler for future refactorings), then it's a stop-gap and any extra effort put into will likely not result in any appreciable reduction of entropy.

grahamegrieve (approved):
ok

PR #1298, opened 2023-06-09
CodeQL coverage

dotasek:


vadi2:
CodeQL might take a while to run - in my experience running it on commits to master only is enough since it reports all the issues it finds to the security tab. If you run it on every pr,  it'll slow down the time to merge one.

Issue #1301, opened 2023-06-12
Local conversion package build failure on SubscriptionTopic40_50Test.java due to hardcoded output stream paths


rettema:
A local maven build of the conversion package / project fails on the SubscriptionTopic40_50Test.java class due to hard coded output stream paths.

```
    org.hl7.fhir.r5.formats.JsonParser r5_parser = new org.hl7.fhir.r5.formats.JsonParser();
    r5_parser.setOutputStyle(OutputStyle.PRETTY).compose(new FileOutputStream("/tmp/source.json"), r5_actual);
    r5_parser.setOutputStyle(OutputStyle.PRETTY).compose(new FileOutputStream("/tmp/output.json"), r5_conv);
```

This appears to be additional, local logic not found in any of the other conversion package test classes and should be removed.

dotasek:
Fixed here: https://github.com/hapifhir/org.hl7.fhir.core/commit/d2c873336c97d0cadf97e58cb964e8e8a662ef50

PR #1305, opened 2023-06-12
Remove hard coded output stream paths from SubscriptionTopic40_50Test class

rettema:
Hard coded output stream paths in the SubscriptionTopic40_50Test class are preventing successful builds of the **org.hl7.fhir.convertors** project in a local Windows environment.

Addresses hapifhir issue #1301


grahamegrieve (approved):
ok

PR #1306, opened 2023-06-12
Get rid of too trusting SSLCertTruster

dotasek:
We shouldn't be trusting all SSL certs, and this code has comments indicating it was a workaround.


grahamegrieve (approved):
ok

PR #1309, opened 2023-06-15
Add lenient-display-validation param

dotasek:
Validation from the remote server doesn't create a `lenient-display-validation` param, meaning that when the Validator CLI is run with  `-display-issues-are-warnings`, this isn't passed on to the server, and all display errors remain errors.

This adds that param.


grahamegrieve (approved):
ok

Issue #1302, opened 2023-06-12
org.hl7.fhir.r4b project should not have Eclipse IDE specific configuration files committed


rettema:
The **org.hl7.fhir.r4b** project contains the Eclipse IDE specific configuration files **.classpath** and **.project**. No other core library project directories contain these files. I believe they have been accidentally committed and should be removed.

PR #1314, opened 2023-06-19
Remove eclipse specific files

dotasek:
Fixes https://github.com/hapifhir/org.hl7.fhir.core/issues/1302


dotasek:
I did the removal and verified that tests run, and that validation can still be run using Eclipse. If this change somehow breaks a developers Eclipse build, the place for a fix would be elsewhere in the project.

Issue #1282, opened 2023-05-29
Allow way to get datetime string with provided precision


TipzCM:
Currently, DateTime objects only provide the value as string in the precision of the defined class.

It would be nice if there was a separate method `getValueAsString(Precision thePrecision)` so as to allow lesser precision for dates as needed.

Would be even nicer if that could be on a non-versioned api, but that might be too much to ask

PR #1315, opened 2023-06-19
New getValueAsString(TemporalPrecisionEnum) method for DateTimeType

dotasek:
Adds a convenience method to get a DateTimeType value as a string with a specific precision that isn't the instances set precision.

Implements https://github.com/hapifhir/org.hl7.fhir.core/issues/1282


grahamegrieve (approved):
ok

PR #1321, opened 2023-06-22
ValidationEngine::validate(): print SourceFile::getRef() instead of Object::toString()

DarthGizka:
`ValidationEngine::validate()`  currently prints things like
```
  Validate org.hl7.fhir.validation.ValidatorUtils$SourceFile@5382184b
```
instead of the name of the source file.

The reason is that `ValidationUtils.SourceFile` does not override Java's default implementation of `toString()`. The proposed fix is to print `getRef()` instead.


grahamegrieve (approved):
ok

Issue #1328, opened 2023-06-28
Validator CLI: typo in `Params::readInteger` prevents use of `-watch-scan-delay` and `-watch-settle-time`


DarthGizka:
[`Params::readInteger`](https://github.com/hapifhir/org.hl7.fhir.core/blob/345151b13219e7f5919ad9235e6302c3b2dc80bf/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/cli/utils/Params.java#L430-L436C1) currently tries to parse the name of the parameter instead of its value.

I propose to rename `n` and `v` to `name` and `value` respectively. This should make it more apparent which spot needs to be fixed. :wink:

In the meantime, pundits can enjoy the new release by modifying a local copy of `Params.java` (simply reference `validator_cli.jar`as a library), compiling it and replacing the class file in the jar. This is much simpler and quicker than trying to build this ginormous project locally.

DarthGizka:
fixed with release 6.0.21

PR #1333, opened 2023-06-30
Fix typo in ValidatorUtils::readInteger(), fix accident-prone file list loop & exempt missing files from processing

DarthGizka:
This patch fixes a typo in `ValidatorUtils::readInteger()` that currently prevents any use of the `watch-mode-delay` and `watch-settle-time` parameters. This solves issue #1328.

It fixes a file list processing loop that misbehaves if the contents of the directory it is processing can change concurrently. This was discussed in the fhir.org chat [watch mode topic](https://chat.fhir.org/#narrow/stream/179239-tooling/topic/Validator.20CLI.20watch.20mode/near/369582987).

The patch also contains two small, surgical changes that keep disappeared files from being processed if their disappearance is already known. This significantly improves stability in the face of disappearing/deleted files.


grahamegrieve (approved):
ok

PR #1360, opened 2023-07-24
Do 20230727 fix example scenario conv

dotasek:
DO NOT MERGE


dotasek:
No longer needed. Deleting.

PR #1373, opened 2023-07-31
Fix error creating new manifest file 2

dotasek:
When validation tests fail, they attempt to write output to a new file in a directory that doesn't necessarily exist.

This creates that directory in that case.

(this is a copy with merged changes from a passing PR)


dotasek:
Closing; the correct fix for the breaking tests is in master, and should be merged into the original PR, not this one.

PR #1376, opened 2023-08-01
Example scenario rendering

lmckenzi:
Add ability to render ExampleScenarios, including diagrams


grahamegrieve:
This POM change brings GPL code into the library, and we can't accept it at all.

grahamegrieve:
looking at the code, we really think that you should use a StringBuilder at line 117 of the ExampleScenarioRenderer.java

lmckenzi:
> looking at the code, we really think that you should use a StringBuilder at line 117 of the ExampleScenarioRenderer.java

Updated to use StringBuilder as suggested.

grahamegrieve (approved):
ok

PR #1381, opened 2023-08-03
TerminologyCache - only trim whitespace outside of the json keys

dotasek:
The terminology cache is stripping whitespace from the entire JSON object when generating hashes, but isn't limited to JSON formatting. This will result in inconsistencies when JSON field values rely on those spaces.

Maybe later, this can be made more JSON aware for efficiency, but for now, reducing the normalization to whitespace trimming outside the JSON allows matches that pass existing tests.


grahamegrieve (approved):
ok

PR #1394, opened 2023-08-15
Bump okhttp dependency, exclude vulnerable transitives

dotasek:


grahamegrieve:
This failed because of an issue in the terminology server

PR #1396, opened 2023-08-17
2023 08 gg tx tests externals

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/test/utils/CompareUtilities.java
```diff
@@ -28,21 +30,52 @@
...
+
+  private String presentExpected(String expected) {
+    if (expected.startsWith("$") && expected.endsWith("$")) {
+      if (expected.startsWith("$choice:")) {
+        return "Contains one of "+readChoices(8, expected).toString();
+      } else if (expected.startsWith("$fragments:")) {
+        List<String> fragments = readChoices(11, expected);
+        return "Contains all of "+fragments.toString();
+      } else if (expected.startsWith("$external:")) {
+        String[] cmd = expected.substring(1, expected.length() - 1).split("\\:");
```
lawley: Should this be split(":") -- I don't think the \\ is wanted here?

This looks like a copy-paste bug from code that splits on | which does need escaping since it is a regex special char.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/test/utils/CompareUtilities.java
```diff
@@ -361,6 +397,14 @@ private static boolean matches(String actualJsonString, String expectedJsonStrin
           }
         }
         return true;
+      } else if (expectedJsonString.startsWith("$external:")) {
+        String[] cmd = expectedJsonString.substring(1, expectedJsonString.length() - 1).split("\\:");
```
lawley, line 423: Same issue as above, ":" shouldn't be escaped

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/test/utils/CompareUtilities.java
```diff
@@ -361,6 +397,14 @@ private static boolean matches(String actualJsonString, String expectedJsonStrin
...
+        if (externals != null) {
+          String s = externals.asString(cmd[1]);
+          return actualJsonString.equals(s);
+        } else {
+          return actualJsonString.contains(cmd[2]);
```
lawley: Should this work like `$fragments:...$` and allow for multiple matches (where each is `|`-separated)?

lawley (changes requested):

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/test/utils/CompareUtilities.java
grahamegrieve: Does it matter? I just generally escape special characters so I don't have to think about it

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/test/utils/CompareUtilities.java
grahamegrieve: same response !

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/test/utils/CompareUtilities.java
grahamegrieve: If there a use case?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/test/utils/CompareUtilities.java
lawley: colons aren't special though
I guess it's benign if `\:` is treated identically to `:`

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/test/utils/CompareUtilities.java
lawley: Wanting the message to mention both the code and the CodeSystem/ValueSet URI

grahamegrieve: ok

PR #1409, opened 2023-08-25
rewrite the way language works in value sets

grahamegrieve:


grahamegrieve:
for an explanation of what's going with these changes, see [test-cases]/tx/languages.md

vadi2:
For anyone else looking for the file - https://github.com/FHIR/fhir-test-cases/blob/29ebeedcb8e8a879770608331e1a75f883c00180/tx/languages.md

Issue #1412, opened 2023-08-29
Out Of Memory issue during resource validation with local reference check enabled 


volodymyr-korzh:
During the validation of the resource with local reference check enabled slow memory leak is happening. 
This causes an Out Of Memory error. 

Analyzing heap dump showed a reference loop between objects `FhirPathEngine` `ValidatorHostServices` `InstanceValidator`:
`FhirConstant` is inner class of `FhirPathEngine` and holds implicit reference to `FhirPathEngine`
`ValidatorHostServices` is inner class of  `InstanceValidator`  and holds implicit reference to `InstanceValidator`
This caused an infinite reference loop `FhirConstant` -> `FhirPathEngine` -> `ValidatorHostServices` -> `InstanceValidator` -> `FhirPathEngine `-> `ValidatorHostServices` etc

 This is an _Inner Classes That Reference Outer Classes_ memory issue which can be fixed by making inner classes static.
As inner class object implicitly holds a reference to the outer class object, thereby making it an invalid candidate for garbage collection.

dotasek:
Confirmed closed in HAPI-FHIR by @volodymyr-korzh

PR #1413, opened 2023-08-29
Fixed Out Of Memory issue during resource validation with local reference check enabled

volodymyr-korzh:
During the validation of the resource with local reference check enabled slow memory leak is happening.
This causes an Out Of Memory error.

This is an Inner Classes That Reference Outer Classes memory issue which can be fixed by making inner classes static.
As inner class object implicitly holds a reference to the outer class object, thereby making it an invalid candidate for garbage collection.

closes issue #1412


grahamegrieve (commented):
@volodymyr-korzh I want to go further, see PR 1414

dotasek:

PR 1414 will also be merged after this.

PR #1419, opened 2023-08-30
Update to cqframework 2.11 and remove unused common-beanutils declaration

Thopap:
hapi fhir validation introduce the usage of commons-beanutils with 05134dc because a transitive outdated was received from cqframework 1.5. Commons-beanutils was never needed at runtime by cqframework and only during buildtime. cqframework 1.5 wrongly declare jaxb2-basics as compile dependency, while only jaxb2-basics-runtime is needed. This was fixed with cqfamework 2.x.
Since 2.x seems to work properly what i could see, an update seems to run without any issue in test execution and the commons-beanutils was removed.


dotasek (changes requested):
This looks good to me. But needs the following lines added to `.github/workflows/license-check/license-special-cases.txt` to allow the 'new' license.

```
# License string includes nested brackets, causing parser breakage, but is a valid BSD license.
(BSD 3-Clause "New" or "Revised" License (BSD-3-Clause)) abego TreeLayout Core (org.abego.treelayout:org.abego.treelayout.core:1.0.3 - http://treelayout.sourceforge.net)
```

Aside from this, I checked IG publisher, and this doesn't impact the results on my test IGs, and all tests appear to pass.

Thopap (commented):
done

dotasek (approved):

PR #1427, opened 2023-09-05
Fix system cache directory

dotasek:
The system cache directory was being built using a relative path resulting in:

`/Users/david.otasek/IdeaFhirCoreLibraries/org.hl7.fhir.core/org.hl7.fhir.r5/var/lib/.fhir/packages/hl7.fhir.r4.core#4.0.1/package/package.json`

instead of:

`/var/lib/.fhir/packages/hl7.fhir.r4.core#4.0.1/package/package.json`


grahamegrieve (commented):
it looks to me like there's a bug here that this makes worse on windows. The correct location of the system cache is documented here: https://confluence.hl7.org/display/FHIR/FHIR+Package+Cache

where does it end up on windows with this change? c:/var/...?

dotasek:
@grahamegrieve oh, yeah, this PR isn't right, then. I'll fix it and push changes.

grahamegrieve (commented):
shouldn't it be C:\\Program Data not C:\\ProgramData?

DarthGizka:
It's a bit more complicated than that. `C:\Program Data` is the location in a Windows installation for an English locale or a multi-language installation; however, even in such installations it can be redirected elsewhere (e.g. when space on the system SSD gets tight). On the other hand, `C:\ProgramData` - without a blank between `Program` and `Data` - will not exist unless a user (or a clueless program) creates it.

The only reasonably reliable way of obtaining the correct location in Java would be `System.getenv("ProgramData")` (no blank in the name of the environment variable!).

dotasek:
@grahamegrieve @DarthGizka you are both correct.

I copied directly from https://confluence.hl7.org/display/FHIR/FHIR+Package+Cache, but @DarthGizka has the correct approach.

BTW, the confluence page should be updated to that effect.

grahamegrieve (approved):
ok

PR #1440, opened 2023-09-19
2023 09 gg type performance

grahamegrieve:


org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackageIndexBuilder.java
```diff
@@ -57,7 +57,7 @@ public void start(String filename) {
             "Content        nvarchar NULL,\r\n"+
             "PRIMARY KEY (FileName))\r\n");
 
-        psql = conn.prepareStatement("Insert into index (FileName, ResourceType, Id, Url, Version, Kind, Type, Supplements, Content) values (?, ?, ?, ?, ?, ?, ?, ?, ?)");
+        psql = conn.prepareStatement("Insert into ResourceList (FileName, ResourceType, Id, Url, Version, Kind, Type, Supplements, Content) values (?, ?, ?, ?, ?, ?, ?, ?, ?)");
       } catch (Exception e) {
         conn = null;
```
dotasek, line 70: This looks like it may leave a hanging connection when exceptions are caught. Don't know if it could cause hangups on trying to delete or rewrite.

Consider:

```java
if (conn != null) { conn.close(); }
```

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackageIndexBuilder.java
```diff
@@ -45,7 +45,7 @@ public void start(String filename) {
         new File(filename).delete();
         conn = DriverManager.getConnection("jdbc:sqlite:"+filename);
```
dotasek, line 46: So, now that I get a longer look at this; are we not locking ourselves into an implementation with this? Are we then forcing ourselves to use sqlite forever, or until we need to break things by switching?

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackageIndexBuilder.java
grahamegrieve: ok

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackageIndexBuilder.java
grahamegrieve: what else could we use? For this, I don't know of anything, so we leave this as it is, I think.

PR #1441, opened 2023-09-20
Add okio dependency for running vsac

dotasek:
Adds the okio dependency required for running VSACImporter


reubendaniels (approved):
I have reviewed this PR and can confirm that it addresses the issue.

PR #1466, opened 2023-10-19
Initial support (R4B) to handle multiple source rules in a group

brianpos:
Add support for rules that have multiple source variables
(as noted in the spec here: https://www.hl7.org/fhir/mapping-language.html#7.8.0.8.1 )
e.g.
```
item.linkId as itemId, itemAnswer.value as coding
     -> condition.id = ('condition_' + %itemId.substring(1) + '_'+ %coding.code
```
The unit test requires some test files that are in the fhir-test-cases repo here:
https://github.com/FHIR/fhir-test-cases/compare/master...brianpos:fhir-test-cases:master

This is only draft as I've only ported the change to the R4B branch, once conceptually approved, I will copy this change (and test) into R4 and R5.

### DIFF
--- a/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/StructureMapUtilities.java
+++ b/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/StructureMapUtilities.java
@@ -1320,6 +1320,13 @@ public class StructureMapUtilities {
   }
 
   public class Variables {
+    public Variables() {
+    }
+    public Variables(Variables parent) {
+      _parent = parent;
+    }
+  
+    private Variables _parent;
     private List<Variable> list = new ArrayList<Variable>();
 
     public void add(VariableMode mode, String name, Base object) {
@@ -1335,6 +1342,7 @@ public class StructureMapUtilities {
     public Variables copy() {
       Variables result = new Variables();
       result.list.addAll(list);
+      result._parent = _parent;
       return result;
     }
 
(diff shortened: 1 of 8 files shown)


brianpos:
Note that the test `testMultipleSourceVariables` passes and works.
However the very similar `testMultipleSourceVariablesCopyId` doesn't and gets a weird can't cast a StringType to an Id message: this is the rule that has the issue
```
item.linkId as itemId, itemAnswer.value as coding
   -> condition.id = ('condition_' + %itemId.substring(1) + '_'+ %coding.code)
```

bvdh:
> Note that the test `testMultipleSourceVariables` passes and works. However the very similar `testMultipleSourceVariablesCopyId` doesn't and gets a weird can't cast a StringType to an Id message: this is the rule that has the issue
> 
> ```
> item.linkId as itemId, itemAnswer.value as coding
>    -> condition.id = ('condition_' + %itemId.substring(1) + '_'+ %coding.code)
> ```

This might be caused by an issue in org.hl7.fhir.r5.model.TypeConvertor.
This class does not accept primitive types for many data classes. Updating `castToId` to 
```
public static IdType castToId(Base b) throws FHIRException {
    if (b == null) {
      return null;
    }

    if (b instanceof IdType)
      return (IdType) b;
    else if ( b instanceof StringType )
      return ( new IdType( b.primitiveValue() ));
    else if (b.hasPrimitiveValue())
      return new IdType( b.primitiveValue() );
    else
      throw new FHIRException("Unable to convert a "+b.fhirType()+"("+b.getClass().getName()+") to a Id");
  }

```
might fix this.

PR #1479, opened 2023-11-01
Update CDA validation targets

benjaminflessner:
Since CDA is in ballot now, it would be great to update the default CDA version to the ballot (especially since this is the basis for the "Common Validation Options" on validator.fhir.org).

Additionally, update C-CDA to use #current, at least until we have a ballot package ready.

@dotasek


benjaminflessner:
Apparently since I can't request reviewers here.....  @grahamegrieve  ping!

grahamegrieve (approved):
ok

PR #1493, opened 2023-11-22
Include time tracking for ValidationResponse

dotasek:
Adds time tracking to ValidationResponse.

This is required for the validator website to be able to report time tracking.

As some validations might break their validation into multiple validation fragments, time tracking is included as a map, with the original file name as the key to each entry.


grahamegrieve (approved):
ok

PR #1497, opened 2023-11-24
Remove fhir-test-cases from Validator CLI JAR

dotasek:
This removes the dependency that increased the Validator CLI JAR size and fixes some tests that were found to be broken.


dotasek:

PR #1503, opened 2023-11-29
Remove basic auth (unneeded) for AU connectathon test

dotasek:
IPSBuilderTest is failing because basic auth is no longer needed (the test will work correctly without auth)


dotasek:

PR #1508, opened 2023-12-01
CLI Task to preload package cache

dotasek:
This PR adds a new task that uses PackageVisitor to visit and load all available packages from the secondary packages server into the filesystem cache.

This PR also includes a refactor of FilesystemPackageCacheManager to use a Builder pattern to configure servers and cache directory locations.


dotasek:
Merging following discussion with @grahamegrieve on Zulip

PR #1510, opened 2023-12-04
Ensure Address.line[] element data is copied over

### LINKED ISSUE #1509: Address.line[] element is not converted in some versions, instead only the String value is
In `org.hl7.fhir.convertors.conv30_40.datatypes30_40.complextypes30_40.Address30_40` the line's StringType value is copied without any extensions. 
In `org.hl7.fhir.convertors.conv40_50.datatypes40_50.general40_50.Address40_50` the whole StringType element is copied over. 

I think the expected behavior is that line elements are copied over in the same way between all versions.

vilaug:
Fixes #1509, I had an issue with this in the 30_40 conversion.


grahamegrieve:
Please fix other affected inter-version conversions

vilaug:
@grahamegrieve what do you mean? I didn't find any other instances of this happening in the Address convertors. The convertors 40_50, 43_50 do not have this issue.

vilaug:
> @grahamegrieve what do you mean? I didn't find any other instances of this happening in the Address convertors. The convertors 40_50, 43_50 do not have this issue.

I see some more primitive types not invoking their convertors, I'll push a commit soon

grahamegrieve (approved):
great, thanks

Issue #1474, opened 2023-10-30
Check that http header names are case insensitive


grahamegrieve:
HTTP says that HTTP header names are case insensitive, and http/2 mandates lowercase. Check that header names are being used in a case-insensitive fashion

dotasek:
The investigation so far suggests that we are using headers in a case insensitive manner.

For dstu2 and dstu2016may, we are using org.apache.http.HttpResponse. Internally, this uses case insensitive checks for equality: https://github.com/apache/httpcomponents-core/blob/d1c77bc7f7697548cd83d04ba87a1ec7105da308/httpcore5/src/main/java/org/apache/hc/core5/http/message/HeaderGroup.java#L213

In dstu3, r4, r4b, and r5, we're using OkHttp3, which also uses case insensitive checks for equality: https://github.com/square/okhttp/blob/3afe55cdcdb8dbaa7bd5e3a914e6412cf587d9a1/okhttp/src/commonMain/kotlin/okhttp3/internal/-HeadersCommon.kt#L68

In utilities, we're using MessageHeader, which is dependent on the Java implementation. So far, Temurin 17 indicates it is also case insensitive, and I believe it will likely be the case for all JVMs. 

(from decompiled code)
```java
 for (int i = nkeys; --i >= 0;) {
                if (k.equalsIgnoreCase(keys[i]))
                    return values[i];
            }
```

What we didn't do in a number of places is stick to lowercase header names in our code. I'm going through those now and changing them all to lowercase, so we are consistent.

grahamegrieve:
so have we sorted this now?

dotasek:
It looks like our libraries are taking care of this for us. I may still try to put in some unit tests to verify their behaviour, but I think it's unlikely they'll regress with regards to this.

I would like to keep the issue open as a reminder until I get some of those tests in, but it won't be a high priority.

PR #1518, opened 2023-12-12
Check that http header names are case insensitive

dotasek:
In response to #1474 

This PR will remain in draft until the tests are updated to check for case insensitivity.


grahamegrieve (approved):
ok

PR #1528, opened 2023-12-31
Add TurtleGeneratorTests

tmprd:
* See https://github.com/w3c/hcls-fhir-rdf/issues/136
* This provides a way to test the generation of RDF resources from XML resources without needing to build the entire FHIR specification.
* Please note that current RDF tests only either parse some sample Turtle files (which may be outdated) or generated ShEx files. In contrast, this generates the very same Turtle files which go in the specification.  
* Some "round-trip" tests may not pass with certain resources, due to existing issues, demonstrating the need for tests. For example, the round-trip parsing of generated Turtle resources currently fails for those containing URIs with vertical bars. See https://github.com/w3c/hcls-fhir-rdf/issues/120#issuecomment-1849044884
* Additional tests will need to be added, but this should be an improvement.


grahamegrieve (approved):
ok

PR #1541, opened 2024-01-12
Enable FIXME check for checkstyle

dotasek:


dotasek:
I will open a fresh version of this from the master branch.

PR #1550, opened 2024-01-19
Fix file type inference in ValidationService

dotasek:
When ValidationService encounters a file that it cannot infer a type for, it sets the file type to null, which isn't correctly handled.

This change includes a fatal error indicating that the file type was not recognized.


grahamegrieve (approved):
ok

PR #1552, opened 2024-01-24
Fix NullPointer exception when params are null

dotasek:
The params in question can be explicitly set to null in some cases, which results in failures in this method, as well as some downstream method calls.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,4 +4,5 @@
 
 ## Other code changes
 
-* no changes
+* Fix null pointer on explicitly null params in DSTU3 terminology requests
+* Fix text creation on wrong xhtml node type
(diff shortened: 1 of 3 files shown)


grahamegrieve:
Indeed they are null, and what should happen is that the parameters in question should be removed from the code entirely Sent from my iPhoneOn 25 Jan 2024, at 6:05 am, dotasek ***@***.***> wrote:﻿The params in question can be explicitly set to null in some cases, which results in failures in this method, as well as some downstream method calls.



You can view, comment on, or merge this pull request online at:
  https://github.com/hapifhir/org.hl7.fhir.core/pull/1552

Commit Summary

  f7ef5bd  Handle case of null params
  5477227  Use new test cases


File Changes  (2 files)

  
    M
    org.hl7.fhir.dstu3/src/main/java/org/hl7/fhir/dstu3/utils/client/FHIRToolingClient.java
    (8)
  
  
    M
    pom.xml
    (2)
  


Patch Links:

  https://github.com/hapifhir/org.hl7.fhir.core/pull/1552.patch
  https://github.com/hapifhir/org.hl7.fhir.core/pull/1552.diff

PR #1553, opened 2024-01-24
Fix self-closing tag logic in XhtmlComposer

qligier:
Fixes https://github.com/HL7/fhir-ig-publisher/issues/848


grahamegrieve:
I finally ran into this problem, but the fix proposed here wasn't complete because of the who empty/expanded thing. I copied this code into trunk, and fixed it up. Thanks

PR #1557, opened 2024-01-29
Restore and fix failing SimpleWorkerContextTests

dotasek:
@grahamegrieve these are the test fixes from the cases you had to disable. What follows is why I think we need them.

These tests verify that worker context is using the appropriate mechanism (cache or client) depending on whether or not the caches contain the necessary objects. The mocks here prevent us from having to fully load a context and have these tests run for way longer. 

I believe these were originally added as part of a refactor of terminology caching to ensure nothing broke. They should continue to indicate that it's performing as intended (recognizing what is in the cache, using the cache or the client, etc).


grahamegrieve:
Wel, maybe, but in practice, I changed the rules for this, and updated the tests. I guess that maybe the updated tests make sure they didn't change by accident so I'm ok with having them

dotasek:
> I guess that maybe the updated tests make sure they didn't change by accident so I'm ok with having them

That's not an ideal test, but I think it's a good first step. I'm thinking of ways to refactor the cache so that tests are easier than this, but my pre-requisite for being able to refactor at all is having something like these to make sure nothing breaks.

Merging.

PR #1577, opened 2024-03-19
Loader for rfc5646 languages file

dotasek:
Adds a class for a language subtag registry, classes for the Language, Script, Region, Variant and ExtLang subtags, as well as a loader to populate the registry from an RFC5646 formatted file.

This registry and the loader do not implement grandfathered and redundant subtags.

The loader allows for granular loading of specific subtag types.


grahamegrieve (approved):
ok

PR #1578, opened 2024-03-19
Refactor terminology service tests for reuse outside core

dotasek:
This moves the logic for testing codes in the local IWorkerContext to the utilities package so that it can be re-used by FHIR implementations such as HAPI.

Further refactoring should likely be done, but this at least exposes the code where it can be accessed via a dependency on org.hl7.fhir.validation.


grahamegrieve (approved):
looks good to me

PR #1593, opened 2024-04-08
2024 04 gg i18n wip2

grahamegrieve:


org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/DateTimeUtil.java
dotasek, line 1: There's a difference in behaviour between the toHumanDisplay methods. In the original, if the precision is DAY, theValueAsString is returned. In the new version, it's dateFormat.format(value). Should theValueAsString be removed from the original method to follow the new implementation, or is the old method destined to be dropped?

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/i18n/I18nBase.java
dotasek, line 1: There is a gotcha lurking in here. Because we set the default locale to en_US, developers running tests in other regions will probably see some failing tests.

The solution here is to add the following to the maven-surefire-config so that tests are run under a consistent default language+region.

```xml
 <argLine>@{argLine} -Xmx5632m -Duser.language=en -Duser.region=US</argLine>                
```

Eclipse users may be impacted by this if they are running tests directly in Eclipse, as the Eclipse JUnit plugin won't pick up surefire args.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/utils/RenderingContext.java
```diff
@@ -176,6 +209,12 @@ public boolean isXml() {
     }
   }
 
+  public enum MultiLanguagePolicy {
+    NONE, // ONLY render the language in the locale
```
dotasek, line 213: The only usage of `MultiLanguagePolicy.NONE`, seems to deviate from the comment: `// ONLY render the language in the locale`. It looks to me like in CodeSystemRenderer, it's rendering according to the default definition instead of requesting any locale.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/ToolingExtensions.java
```diff
@@ -802,6 +811,20 @@ public static String getLanguageTranslation(Element element, String lang) {
     return null;
   }
 
+  public static StringType getLanguageTranslationElement(Element element, String lang) {
+    for (Extension e : element.getExtension()) {
+      if (e.getUrl().equals(EXT_TRANSLATION)) {
+        Extension e1 = ExtensionHelper.getExtension(e, "lang");
+
+        if (e1 != null && e1.getValue() != null && e1.getValue() instanceof CodeType && ((CodeType) e1.getValue()).getValue().equals(lang)) {
```
dotasek, line 819: Are we only accepting exact matches for lang? If for example, the extension has `en` as the lang, it looks to me like this will not even attempt to translate `en_US`

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/DateTimeUtil.java
grahamegrieve: That's a good question. I think the original method is wrong, but changing it... we should consult James

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/i18n/I18nBase.java
grahamegrieve: I think we should look at the actual tests before deciding what to do. I don't run the tests in US region, so I expect to see any issues immediately

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/utils/RenderingContext.java
grahamegrieve: that's fixed in my next commit

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/ToolingExtensions.java
grahamegrieve: Good point, and one that's on my todo list

grahamegrieve:
no failing tests in AU region

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/i18n/I18nBase.java
dotasek: In testing the above, I got mine to fail by switching to `de-DE`, which breaks any tests that string match based on message bundles. `en-AU` will still give you translation phrases in English. I can run a whole build with this set, but set to fail-at-end and try to get a list of all failures.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/ToolingExtensions.java
dotasek: I think there's some logic here that may be of use: https://devdocs.io/openjdk~17/java.base/java/util/locale#filter(java.util.List,java.util.Collection)

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/DateTimeUtil.java
dotasek: I suspected it might be wrong. 

I can check in HAPI for usages, either directly, or through `toHumanDisplay()` and see what the impact might be.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/i18n/I18nBase.java
grahamegrieve: what did you switch? You changed the code, right? Then that's expected

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/DateTimeUtil.java
grahamegrieve: yes please

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/i18n/I18nBase.java
dotasek: Yes, but that's because I'm emulating what a developer in a different region would experience. A dev in Germany would have a `de-DE` locale, which would be picked up as the test runner default. Their tests would fail like mine did unless the explicitly switch to `en-US`. We can make this unnecessary with the argline addition I mentioned.

grahamegrieve: but as far as I can see, none of the tests I changed here run in the default locale. That's what I'm not understanding - which tests?

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/DateTimeUtil.java
dotasek: It looks like FastDateFormat may not work as expected. I switched `ourHumanDateFormat` to public visibility, and ran this test:

```java
@Test
  public void testDateTimeUtilHumanDateFormat() {
    DateTimeType dayPrecision = new DateTimeType("2019-01-02");
    Assertions.assertEquals(dayPrecision.toHumanDisplay(),DateTimeUtil.ourHumanDateFormat.format(dayPrecision.getValue()));
    DateTimeType monthPrecision = new DateTimeType("2019-01");
    Assertions.assertEquals(monthPrecision.toHumanDisplay(), DateTimeUtil.ourHumanDateFormat.format(monthPrecision.getValue()));
    DateTimeType yearPrecision = new DateTimeType("2019");
    Assertions.assertEquals(yearPrecision.toHumanDisplay(), DateTimeUtil.ourHumanDateFormat.format(yearPrecision.getValue()));
  }
```

The actual format it's using is different than the one we expect:

```
Expected :2019-01-02
Actual   :2-Jan-2019
```

Switching would no doubt break some tests for us and HAPI. Even switching to FastDateFormat.getDateInstance(FastDateFormat.SHORT) gives us the format `02/01/19`. 

I'd argue that since toHumanDisplay claims it "Returns a human readable version of this date/time using the system local format.", using this format is correct, and we should make the switch. It will break a handful of tests for HAPI, and for us, but would be correct according to our method documentation.

dotasek: Oh, and precision goes right out in using FastDateFormat. It's always full length:

```
Expected :2019-01
Actual   :01/01/19
```

PR #1594, opened 2024-04-10
Update semver regex

gabriel0316:
https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Semver.20validation

### DIFF
--- a/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/VersionUtilities.java
+++ b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/VersionUtilities.java
@@ -328,7 +328,7 @@ public class VersionUtilities {
     if (Utilities.noString(version)) {
       return false;
     }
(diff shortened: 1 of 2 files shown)


grahamegrieve:
I don’t think this is enough by a long shot. If we’re going to allow + in server versions, it needs a lot of testing so that we know it works across the whole ecosystem. Does the validator support this? Do the package servers support it? Does the publisher support it?

And in the code, does versionutilities support it?

gabriel0316:
I see - yes, that of course makes sense. 

But this change may be a good starting point. And, the current version (in production) of the regex already allowed the following: `1.0.0-asdf+asdf+asfd`

grahamegrieve:
@dotasek for us to discuss next week

grahamegrieve:
see https://chat.fhir.org/#narrow/stream/179239-tooling/topic/NPM.20.26.20SemVer

grahamegrieve:
see https://chat.fhir.org/#narrow/stream/179239-tooling/topic/NPM.20.26.20SemVer: it's good to go, and we need to do this across the board

grahamegrieve:
to test:

* support for + in FilesystemBasedPackageCache
* support for + on packages.fhir.org
* support in the package client
* validation in FHIRVersion elements
* support in go-publish

PR #1596, opened 2024-04-11
Make sure JUnit 4 and 5 dependencies are optional in org.hl7.fhir.utilities module

### LINKED ISSUE #1561: OSGI metadata of org.hl7.fhir.utilities introduces hard junit dependency
The OSGI metadata of org.hl7.fhir.utilities introduces a hard junit dependency. In the Maven POM, JUnit is registered as a provided dependency.

Please update the OSGI metadata of the module in such a way that the junit packages are imported only optionally to align the OSGI metadata semantics with the Maven POM semantics.

cmark:
Also removed duplicate managed versions of JUnit 4.13.2 and mockwebserver 4.11.0.

Fixes #1561

### DIFF
--- a/org.hl7.fhir.utilities/pom.xml
+++ b/org.hl7.fhir.utilities/pom.xml
@@ -89,7 +89,6 @@
             <optional>true</optional>
         </dependency>
 
-
         <dependency>
             <groupId>com.ibm.icu</groupId>
             <artifactId>icu4j</artifactId>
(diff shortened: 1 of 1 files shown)


reckart:
Checked OSGI metadata in utilities JAR - JUnit is now optional - looks good to me :)

dotasek:
I have no objections the intent here. 

There is one case I would like to check before I approve this: the Validator CLI supports executing JUnit tests so that FHIR contributors can run them against a local fhir-test-cases directory without having to use a JRE. It's possible the dependencies will have to be added in the validator-cli module now that they are removed from utilities.

cmark:
Hi @dotasek, I think all the necessary dependencies are declared in the validation.cli module but if you find something that's missing I can add it. Let me know and I will update the PR.

reckart:
@dotasek the dependencies have been [`scope provided`](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-scope) already, meaning that they were already not transitive before.

dotasek (approved):
Can confirm that compiled CLI performs as intended with these changes.

cmark:
Thanks @dotasek!

PR #1609, opened 2024-05-01
Use extension class on Exception to force try-catch blocks on checkCodeOnServer calls

dotasek:
checkCodeOnServer calls in InstanceValidator can produce general Exceptions. By switching to a wrapper exception, we can force handling of any internal exceptions in InstanceValidator code.


dotasek:
Merging, after discussion with @grahamegrieve

PR #1612, opened 2024-05-06
Move i18n coverage png table to job in master pipeline

dotasek:


dotasek:
Since the master pipeline is failing, this will also fail. The intent of this PR is to make the master branch appear as the only pipeline in commit history, rather than have the i18n pipeline report success on a failing master build.

PR #1614, opened 2024-05-07
Fix/canonical split base validator

Boereck:


grahamegrieve:
well, you added a test case, but it doesn't fail?

Boereck:
I added two commits, one with the fix and one with the test case. I can squash them into one. Or I can first only include the test case, so we see the test fail in CI, and later add the commit with the fix to the branch, so we can see the test case works.

grahamegrieve:
where's the test case? All I saw was a supplement?

Boereck:
I actually forgot to test for the absence of errors. I updated the respective test case. Now in [ResourceValidationTests](https://github.com/hapifhir/org.hl7.fhir.core/pull/1614/files#diff-4d8c209ac7af79c4d436e2ee3cd8ea5dbd22613a9bc67823564f073e119280f1R131-R153) I check if the file `codesystem-example-supplement-version.xml` from `fhir-test-cases` validates without errors, which it was not before the fix in commit https://github.com/hapifhir/org.hl7.fhir.core/pull/1614/commits/81f02da5c2c0b18fcf7cb06652fe111640012873 (which is part of this PR).

Let me know if anything needs to be changed to be accepted.

grahamegrieve (approved):
ok

PR #1632, opened 2024-05-22
Fix handling of hintAboutNonMustSupport to actually be useful.  Speci…

lmckenzi:
…fically:

- don't yell about missing mustSupport if differential element is setting max to 0, is just declaring slicing, or has a fixed or pattern value
- make clear *what* element doesn't have a mustSupport element in a profile that should


grahamegrieve (commented):
is there a test case for these changes?

lmckenzi:
The fixes are about the rendering.  Do we have test cases for the actual rendering?  If so, where?

grahamegrieve (approved):
ok

PR #1636, opened 2024-05-26
Corrected ObligationsRenderer to handle multiple actors and multiple …

lmckenzi:
…codes.  Also got obligations with elements to render properly (which means knowing whether you're on a table page or definitions page, what tab you're on, and whether the element for the obligation is in-scope for that tab (so you know whether to hyperlink or not).  Had to make links on the tables point to definitions because table anchors are not unique.


grahamegrieve (commented):
why the changes to the old versions? Does this pass the test cases?

lmckenzi:
IntelliJ said the code wasn't used anywhere, and it seemed better to remove the modules than update them.  Why would the test cases use them if nothing else does?

grahamegrieve:
> IntelliJ said the code wasn't used anywhere, and it seemed better to remove the modules than update them.

They're public, so you don't know whether they're being used. How are they relevant?

> Why would the test cases use them if nothing else does?

The test cases question was about your actual code

lmckenzi:
See https://github.com/FHIR/fhir-test-cases/pull/173

grahamegrieve:
I still can't accept this until the removed code is restored

grahamegrieve (approved):
ok. will fix up afterwards

PR #1640, opened 2024-05-28
Maintain a map of pre-built ValidationEngine instances 

dotasek:
This adds a map of ValidationEngine instances to ValidationService.

When ValidationService receives a validation request in which the CliContext specifies a baseEngine (a key for the map), the new engine is generated from the pre-existing one, a faster operation than loading all packages.

The only caveat for generation is that the language and locale for the new engine are still set from the CliContext after the engine is copied.


grahamegrieve (approved):
ok

PR #1642, opened 2024-05-30
Log validation session counts

dotasek:
Adds additional logging indicating the number of sessions currently in the session cache. Useful for evaluating current ValidatorService resource usage in the wild.


grahamegrieve (approved):
ok

PR #1644, opened 2024-05-31
Fix for single quote escapes in I18n

dotasek:
Single quote escapes in messages without message arguments were being treated as regular strings, which bypassed single quote escapes in MessageFormat.


grahamegrieve (approved):
thanks

PR #1646, opened 2024-06-02
proposed fix for hasPrimitiveValue()

grahamegrieve:


dotasek:
Something that looks suspect: 

- `hasPrimitiveValue()` is calling `super.hasPrimitiveValue()` from `Base`
- `Base.hasPrimitiveValue()` is calling `primitiveValue()`
- `primitiveValue` in `ElementModel` is calling `hasPrimitiveValue()` again:

```java
public String primitiveValue() {
		if (isPrimitive() || value != null)
		  return value;
		else {
			if (hasPrimitiveValue() && children != null) {
				for (Element c : children) {
					if (c.getName().equals("value"))
						return c.primitiveValue();
				}
			}
			return null;
		}
	}
	```

**Comment by dotasek** [comment]
Yeah, that's causing stack overflows in ValidationTests:

```
java.lang.StackOverflowError
	at org.hl7.fhir.r5.model.PrimitiveType.isEmpty(PrimitiveType.java:133)
	at org.hl7.fhir.r5.model.ElementDefinition$TypeRefComponent.hasCodeElement(ElementDefinition.java:2116)
	at org.hl7.fhir.r5.model.ElementDefinition$TypeRefComponent.getWorkingCode(ElementDefinition.java:2599)
	at org.hl7.fhir.r5.elementmodel.Property.getType(Property.java:205)
	at org.hl7.fhir.r5.elementmodel.Element.isPrimitive(Element.java:600)
```

PR #1647, opened 2024-06-03
Include validation of local files

bvdh:
I've added support for validation of local files. 
For this I've updated IGloader to return the loaded resource type and id.
This ValidationEngine will use that information to create a R5EvaluationContext class which is passed to the InstanceValidator. This class holds the files as Elements. When a resource is retrieved using FHIRpath, this element is returned and validated.

To test this I've added a test that loads files and validates them.

Is this an approach that would work?


grahamegrieve:
Thanks for this contribution but I don't think this is how it should be solved. But stimulated by what you contributed, I did make some other changes. See https://chat.fhir.org/#narrow/channel/179166-implementers/topic/Validation.20error.20on.20type.20slicing.20with.20resolve.28.29

PR #1649, opened 2024-06-06
Fix case typo in Consent.ProvisionComponent

dotasek:


dotasek:
Pinging @jamesagnew for an opinion. Though the change here is breaking the JSON parser right now, would changing the case of a method in one of the previous releases cause any trouble downstream in HAPI or HAPI dependent projects?

dotasek:
From a conversation with @jamesagnew 

> **james** IMO I'd say it's worth making.. it's a breaking change, but I agree it's super easy to fix and woerth fixing

PR #1651, opened 2024-06-09
First draft of ResourceElement - rewrite rendering layer

grahamegrieve:


org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/rendering/ResourceElementTests.java
```diff
@@ -0,0 +1,226 @@
...
+
+    assertEquals(children.get(2), obs.child("value"));
+    assertEquals(children.get(2), obs.child("value[x]"));
+    checkObsValue(children.get(2));
+    
+    assertEquals(children.get(0), obs.child("contained"));
+    checkContained(children.get(0));
+  }
+
+  private void checkContained(ResourceElement cont) {
```
dotasek, line 142: Should we check the reference value?

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/rendering/ResourceElementTests.java
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  private void checkObservation(ResourceElement obs) {
+    Assertions.assertTrue(obs.fhirType().equals("Observation"));
+    Assertions.assertEquals("resource", obs.name());
+    Assertions.assertEquals("Bundle.entry[0].resource", obs.path());
+    Assertions.assertEquals(ElementKind.BundleEntry, obs.kind());
+    
+    List<ResourceElement> children = obs.children();
+    assertEquals(3, children.size());
```
dotasek: Something doesn't seem quote right here. Are `id` and `text` not children?

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/rendering/ResourceElementTests.java
```diff
@@ -0,0 +1,226 @@
...
+    Assertions.assertEquals("Provenance", cont.fhirType());
+    Assertions.assertEquals("contained", cont.name());
+    Assertions.assertEquals("Bundle.entry[0].resource.contained[0]", cont.path());
+    Assertions.assertFalse(cont.isPrimitive());
+    Assertions.assertFalse(cont.hasPrimitiveValue());
+    Assertions.assertTrue(cont.hasChildren());
+    Assertions.assertEquals(ElementKind.ContainedResource, cont.kind());
+  }
+
+  private void checkObsValue(ResourceElement obsValue) {
```
dotasek, line 152: Should we check the value here?

dotasek (changes requested):
There are some things about the tests that don't add up, or seem to be missing. See comments.

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/rendering/ResourceElementTests.java
grahamegrieve: I added those properties to the test case - will update the source for the test

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/rendering/ResourceElementTests.java
grahamegrieve: which value?

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/rendering/ResourceElementTests.java
grahamegrieve: I suppose we should

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/rendering/ResourceElementTests.java
dotasek: After the merge, but I did not want to leave this hanging. Should we check the id value?

```xml
<id value="obs1"/>
```

Issue #1658, opened 2024-06-18
Mime type checking can't handle dash character


pieter-edelman-nictiz:
Hi,

The most recent version of the Validator complains when it encounters the following URL in the narrative of a resource: `data:application/octet-stream;base64,volmacht`, with the message:
> The URL is not valid because '(The mimetype portion of the data: URL is not valid (Mime type invalid) in URL: data:application/octet-stream;base64,volmacht)': data:application/octet-stream;base64,volmacht

See [test.zip](https://github.com/user-attachments/files/15884620/test.zip) for the offending file.

So it claims 'application/octet-stream' is not a valid mime type, although it actually is valid. I think this is due to the ` checkValidMimeType` method in InstanceValidator.java. Here, the following regex is used to check for validity of the mime type:
```
^(\w+|\*)\/(\w+|\*)((;\s*(\w+)=\s*(\S+))?)$
```
However, the `\w` character class doesn't include the normal dash, although this character is allowed and quite common in the subtype part of mime types. Changing the regex to something like:
```
^(\w+|\*)\/([\w-]+|\*)((;\s*(\w+)=\s*(\S+))?)$
```
should probably solve the problem.

PR #1664, opened 2024-06-20
Issue #1658: recognize that mimetype subtypes can contain dashes

pieter-edelman-nictiz:
Fix for issue #1658

### DIFF
--- a/org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/instance/utils/UrlUtilTest.java
+++ b/org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/instance/utils/UrlUtilTest.java
@@ -35,6 +35,10 @@ public class UrlUtilTest {
   @ValueSource(strings = {
     "",
     "data:text/plain", // Actual data is missing
+    "data:application/octet-stream;base64", // Actual data is missing
+    "data:application/octet-stream;base64,A==", // Improperly padded base64 data
+    "data:application/octet-stream;base64,A(B=", // Invalid character in the default encoding
+    "data:application/octet-stream;base64AAAA", // Missing comma in data URL
   })
   public void testInvalidUrl(String url) {
     assertNotNull(UrlUtil.checkValidUrl(url, context));
--- a/org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/instance/utils/UrlUtilTest.java
+++ b/org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/instance/utils/UrlUtilTest.java
@@ -1,0 +1,43 @@
+package org.hl7.fhir.validation.instance.utils;
+
+import org.junit.jupiter.params.ParameterizedTest;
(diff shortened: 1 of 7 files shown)


pieter-edelman-nictiz:
> I tested manually here:

That's because backslashes are double escaped in the Java source code, you need to remove all double backslashes in the manual tester. Also, the `data:` part needs to be removed from the test string.

I'm not so familiar with the code base that I can say something useful about the MimeTypeUtil suggestion. This seemed like an easy fix.

dotasek:
@pieter-edelman-nictiz You are correct regarding the escapes and the ';' char.

From a reviewing standpoint, none of that was clear in the original method, which I assumed would include all forms of mime types.

See the test case PR I created. It actually revealed that there's valid mimetypes (commented for now) that should pass, but do not.

pieter-edelman-nictiz:
Hi @dotasek , this makes a lot of sense. But wouldn't it be better to break out the whole validity check for data url's or url's to such an utility class?

pieter-edelman-nictiz:
> Hi @dotasek , this makes a lot of sense. But wouldn't it be better to break out the whole validity check for data url's or url's to such an utility class?

Well, asking the question is answering it, it seems. Using unit tests for the url checker, I found a few valid `data:` url's that were rejected by the current algorithm.

org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/instance/utils/UrlUtilTest.java
```diff
@@ -0,0 +1,43 @@
...
+    "https://en.wikipedia.org/w/index.php?title=Fast_Healthcare_Interoperability_Resources",
+  })
+  public void testValidUrl(String url) {
+    assertNull(UrlUtil.checkValidUrl(url, context));
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {
+    "",
+    "data:text/plain", // Actual data is missing
```
dotasek, line 37: It would be good to have one or two failing base64 tests here, since there is testing logic specifically for that.

dotasek (changes requested):
This looks way better. If you add one or two more test cases to cover base64 where I commented, I will merge.

org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/instance/utils/UrlUtilTest.java
pieter-edelman-nictiz: Ok, added some tests that I could think of.

I also tried some malformed URLs ("http:/hl7.org" and such), but they don't get rejected as the original code only checks for validity of URL characters. But I feel like fixing this would again expand the scope of this PR, plus that it seems more sensible to use an existing library when you actually want to have a full-blown URL validator.

dotasek (approved):

pieter-edelman-nictiz:
@dotasek many thanks for the guidance with this PR :)

PR #1665, opened 2024-06-20
Use v2 of PublishCodeCoverageResults

dotasek:
PublishCodeCoverageResults version 1 is being deprecated.

This PR should only be merged if it reports code coverage correctly in the checks section.


dotasek:
This does not work as anticipated:

Screenshot 2024-06-20 at 11 35 02 AM

It looks to me like module should point to an actual module, not a source file, and that furthermore, the source file should be an http link to coverage, like it used to be.

dotasek:
This PR now includes two tasks to publish code coverage:

**v2 of PublishCodeCoverage** (default) - executes fast, but doesn't include line-level previews of covered code.

**v5 of reportgenerator** - includes detailed coverage reports, but executes much slower.

To execute detailed coverage reports, run a new pull request pipeline and set the `verboseCoverage` parameter to true.

PR #1682, opened 2024-07-09
Introduce backwards compatible constructor in ValidationEngineBuilder and deprecate

dotasek:
This re-introduces a previous version of a constructor for backwards compatibility purposes, and deprecates all ValidationEngineBuilder constructors with args, as they should be private and only used internally in that class.

See: https://github.com/hapifhir/org.hl7.fhir.core/commit/60bf358bfa978b95b39041d8412f92ea401cefa9#r143962456


dotasek:
Merging after review with @grahamegrieve

PR #1683, opened 2024-07-09
Try using reportgenerator task to get better line coverage in Azure Pipeline

dotasek:
PublishCodeCoverageResults version 1 is being deprecated, and version 2 doesn't display detailed line coverage like version 1 did. 

The suggested workaround is to directly use the reportgenerator task that version 1 used to use, but to trigger it directly, removing the need for PublishCodeCoverageResults:

https://marketplace.visualstudio.com/items?itemName=Palmmedia.reportgenerator


dotasek:
Included in https://github.com/hapifhir/org.hl7.fhir.core/pull/1665, so this PR is no longer needed.

PR #1687, opened 2024-07-15
2024 07 gg measure validation

grahamegrieve:


dotasek:
This pipeline succeeded here: https://dev.azure.com/fhir-pipelines/fhir-core-library/_build/results?buildId=18367&view=results

Not sure why it's not reporting that in the GitHub UI.

PR #1689, opened 2024-07-16
Fix code validation for locale

dotasek:
If stack.getWorkingLang() is null, code validation should use the context locale instead.


grahamegrieve (approved):
ok

grahamegrieve (approved):
great

PR #1692, opened 2024-07-18
Return null for all toCode enum params that are null (DSTU2)

dotasek:
This adds the following null handling for any null Enums in toCode methods:

```java
if (code == null)
        return null;
```


grahamegrieve (approved):
looks great to me

PR #1700, opened 2024-07-24
Fix interaction between terminology cache and validation location 

dotasek:
When the BaseWorkerContext retrieves a previously cached code validation, other parts of the code can then mutate it, resulting in inconsistent responses from the cache. In one noted case, this code causes incorrect locations to be reported by subsequent validations: https://github.com/hapifhir/org.hl7.fhir.core/blob/1ad319ee35f63f7163c336261fcf296293e69dc4/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/model/OperationOutcome.java#L1421

This PR solves the issue (at least pertaining to ValidationResult) by using a copy of the cached ValidationResult instead of the cached object.

Tests that verify that the correct objects are used have been included, and a refactor of the tests was also performed after realizing the tests could directly use BaseWorkerContext instead of SimpleWorkerContext.


dotasek:
The extended discussion of additional improvements can take place here: https://github.com/hapifhir/org.hl7.fhir.core/issues/1705

grahamegrieve (approved):
ok

Issue #1703, opened 2024-07-26
FML transform performance - excessive creation of FHIRPathEngine in new Properties


mrunibe:
When using FML transform with large input/output data, a lot of runtime is spent in creating new Properties. Looks like this is caused by:

- In `TransformSupportServices.createType()` https://github.com/hapifhir/org.hl7.fhir.core/blob/1ad319ee35f63f7163c336261fcf296293e69dc4/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/TransformSupportServices.java#L37-L40
- for every new Property, a `new ProfileUtilities` is instantiated: https://github.com/hapifhir/org.hl7.fhir.core/blob/1ad319ee35f63f7163c336261fcf296293e69dc4/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/Property.java#L89-L91
- and therefore also a somewhat heavyweight `FHIRPathEngine`  object is created: https://github.com/hapifhir/org.hl7.fhir.core/blob/1ad319ee35f63f7163c336261fcf296293e69dc4/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java#L458-L460

**Possible solution: if ProfileUtilities/FHIRPathEngine object is be re-used from StructureMapUtilities, the transform is substantially faster, especially for larger input data:**

| # Elements  | Time [s]  | Time with Workaround [s] <br>Skip numberChildren https://github.com/hapifhir/org.hl7.fhir.core/issues/1699 | Time with Tuning [s] <br>Re-use FHIRPathEngine https://github.com/mrunibe/org.hl7.fhir.core/commit/0c62b94d8200e26c5438e8737ef35fd68f337d87  |
| ------------- | ------------- |------------- |------------- |
| 10  | 0.2  | 0.2 | **0.1** |
| 100  | 0.5  | 0.5 | **0.1** |
| 1'000  | 2  | 2 | **0.2** |
| 10'000  | 64  | 11 | **0.5** |
| 20'000  | 250  | 23 | **0.8** |
| 100'000  | 5316  | 126 | **2.4** |
| 200'000  | - | 240 | **3.6** |

It also significantly reduces the memory footprint (case with 200'000 elements now runs with 3.5gb heap, before required ~7gb).

The issue/numbers can be reproduced with the attached [testcase.zip](https://github.com/user-attachments/files/16391662/testcase.zip):

[...]

Issue #1699, opened 2024-07-23
FML transform getting very slow for growing/large lists


mrunibe:
We have noticed that the FML transform is getting very slow if it is used with a large number of elements (target has an element with many children). I could reproduce it with a simple dummy testcase: A bundle with a large number of observations.

Runtimes for the transformations (time measured just for the [transform() ](https://github.com/hapifhir/org.hl7.fhir.core/blob/1ad319ee35f63f7163c336261fcf296293e69dc4/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/structuremap/StructureMapUtilities.java#L1261)without loading resources):

| # Elements  | Time [s] |
| ------------- | ------------- |
| 10  | 0.2  |
| 100  | 0.5  |
| 1000  | 2  |
| 10000  | 64  |
| 20000  | 250  |
| 100000  | 5316  |

The approx. **quadratic increase of runtime** is caused by the recursive call to `numberChildren()` for every child that is created with `makeElement()`. For every element that is added, the entire list of children is recursively re-numbered every time.

![profiler](https://github.com/user-attachments/assets/69d72da1-5455-4a37-bb29-ad7c5c822635)

With a **workaround by disabling the numberChildren** (https://github.com/mrunibe/org.hl7.fhir.core/commit/a0f67d46ecd02d46b81c4f521e253d32fce89bcd), the performance becomes linear again:

| # Elements  | Time [s]  | Time with Workaround [s] |
| ------------- | ------------- |------------- |
| 10  | 0.2  | **0.2** |
| 100  | 0.5  | **0.5** |
| 1000  | 2  | **2** |
| 10000  | 64  | **11** |
| 20000  | 250  | **23** |
| 100000  | 5316  | **126** |

The issue/numbers can be reproduced with the attached [testcase.zip](https://github.com/user-attachments/files/16346831/testcase.zip):

`java org.hl7.fhir.validation.ValidatorCli testcase/bundle-10000.json -transform http://test.ch/DummyBundleToBundle -version 4.0 -ig testcase/map  -output result-10000.json`

Not sure if `numberChildren()` is required in the `makeElement()` (the index seems to be used for the $index FHIRPath expression)? At least I have not observed any issues for the transform with the workaround.

CC @oliveregger

mrunibe:
Fixed by https://github.com/hapifhir/org.hl7.fhir.core/pull/1704, successfully verified with version 6.5.0

PR #1704, opened 2024-07-26
FML transform performance tuning

mrunibe:
Imrove performance of FML transform, especially for data with large lists.

- Skip numberChildren() when adding an element to a list: https://github.com/hapifhir/org.hl7.fhir.core/issues/1699
- Avoid creating new FHIRPathEngine in every new Property: https://github.com/hapifhir/org.hl7.fhir.core/issues/1703

For large amount of input/output data, FML transform is > 1000x faster and using less memory.

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/Element.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/Element.java
@@ -950,6 +950,8 @@ public class Element extends Base implements NamedItem {
         child.sort();
         if (child.isEmpty())
           remove.add(child);
+        else
+          child.numberChildren();
       }
       children.removeAll(remove);
       children.sort(new ElementSortComparator(this, this.property));
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/Element.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/Element.java
@@ -563,7 +563,6 @@ public class Element extends Base implements NamedItem {
         } else {
           Element ne = new Element(child).setFormat(format);
           children.add(ne);
-          numberChildren();
           return ne;
         }
       }
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/Element.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/Element.java
@@ -475,7 +475,7 @@ public class Element extends Base implements NamedItem {
         } else {
           Element ne = new Element(child).setFormat(format);
           children.add(ne);
-          numberChildren();
+          ne.index = children.getSizeByName(ne.getListName()) - 1;
           childForValue = ne;
           break;
         }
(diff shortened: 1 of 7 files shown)


grahamegrieve:
why is it ok to skip numberChildren() when changing the list?

mrunibe:
> why is it ok to skip numberChildren() when changing the list?

At least **during** the transform I have not seen where the index is used. But maybe we can just defer numberChildren() until the end of the transform? For example, do it when sorting: https://github.com/hapifhir/org.hl7.fhir.core/blob/1ad319ee35f63f7163c336261fcf296293e69dc4/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/structuremap/StructureMapUtilities.java#L1274-L1277

I've also noticed the numbering is not done in other place like [forceElement](https://github.com/hapifhir/org.hl7.fhir.core/blob/1ad319ee35f63f7163c336261fcf296293e69dc4/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/Element.java#L591) or [addElement](https://github.com/hapifhir/org.hl7.fhir.core/blob/1ad319ee35f63f7163c336261fcf296293e69dc4/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/Element.java#L1364), making me naively assume it might not be required in makeElement either.

Do you think defering it to the end of transform() is ok?

oliveregger:
@mrunibe thanks for that performance optimisation, this has a huge speed impact!

I had a look at the issue with numberChildren(), I think one cause of the performance issue numberChildren is that numberChildren is called recursively within the children, 

https://github.com/hapifhir/org.hl7.fhir.core/blob/master/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/Element.java#L291
			child.numberChildren();

if @grahamegrieve aggres, maybe you could update your PR and adapt that numberChildren() is not calling child.nubmerChildren?

i tested the approach with matchbox for the upcoming next version and all transformation and validation tests are passing with that approach, looks really promising.

grahamegrieve:
The point of numberChildren is that it calls recursively. @dotasek and I reviewed this, and I have to investigate when it is and isn't safe to not call numberChildren

oliveregger:
> The point of numberChildren is that it calls recursively. 

why would that be necessary? if a child is inserted and that child would have itself a list of elements, should there the indexes not already be populated?

>@dotasek and I reviewed this, and I have to investigate when it is and isn't safe to not call numberChildren

thanks!

An alternative would be maybe to handle the indexes directly in NamedItem and sync the index there during manipulation add/remove? if you think that would be an option i could work on a pr for that.

mrunibe:
Following up on the idea from @oliveregger handling the indices directly:
 
NamedItemList already contains a per list map that we can use to get the current size and therefore set the index immediately when adding a child. This should be equivalent to calling numberChildren.

I changed this accordingly (https://github.com/hapifhir/org.hl7.fhir.core/pull/1704/commits/8d780e078492cc310603f71bc47e8326fef8740e) and also added a check in the testcases (https://github.com/hapifhir/org.hl7.fhir.core/pull/1704/commits/f909a0c228d84ab5232e3f335c9eab4dfb6121d8)

grahamegrieve (approved):
ok - thanks

PR #1706, opened 2024-07-26
CapabiltyStatement add space after capability expectation verb

caspears:
There was no space between the capability expectation verb (SHALL, SHOULD, MAY, etc.) and the text that follows it. This commit adds the space.


caspears:
Added several other fixes including for 

https://github.com/HL7/fhir-ig-publisher/issues/929
CapabilityStatement rendering of supported profiles should display the profile Title where possible #929

and

https://github.com/HL7/fhir-ig-publisher/issues/927
CapabilityStatement with document #927
(For support of rendering document and messaging

grahamegrieve:
Why only those two uses of addResourceLink()? @vassilpeytchev what's supposed to happen here?

caspears:
@grahamegrieve , I was doing some more testing and the last part did not quite yet get my final check-in pushed for the document and message rendering (https://github.com/HL7/fhir-ig-publisher/issues/927). I have made that push now.

Regarding the addResourceLink() calls, I did not have the history on its use. I only changed the ones I saw that seemed wrong given the resource "supported profiles" appear in the summary table correctly with a table, but only with the urls in the resource detail sections. I just wanted to fix what I was reasonably assured could use fixing (didn't want to break anything I was less familiar with).

caspears:
@grahamegrieve, I will be meeting with Vassil this week to go over these changes and any needed additional changes. I will pull this PR for now.

Issue #1641, opened 2024-05-30
Concurrency issues with package cache in validator


dehall:
When multiple validator sessions are starting up at the same time we've seen Exceptions that appear to be concurrency issues when reading from certain files in the package cache.
Similar to #1491 - this issue was initially observed using the [org.hl7.fhir.validator-wrapper](https://github.com/hapifhir/org.hl7.fhir.validator-wrapper) but able to reproduce just with the core validation library.
An example stack trace:
```
java.io.IOException: Error parsing JSON source: Unexpected content at start of JSON: Eof at Line 1 (path=[null]) from '/Users/dehall/.fhir/packages/hl7.fhir.uv.extensions.r4#1.0.0/usage.ini'
	at org.hl7.fhir.utilities.json.parser.JsonLexer.error(JsonLexer.java:120)
	at org.hl7.fhir.utilities.json.parser.JsonParser.parseSource(JsonParser.java:244)
	at org.hl7.fhir.utilities.json.parser.JsonParser.parseJsonObject(JsonParser.java:230)
	at org.hl7.fhir.utilities.json.parser.JsonParser.parseObject(JsonParser.java:71)
	at org.hl7.fhir.utilities.npm.FilesystemPackageCacheManager.loadPackageInfo(FilesystemPackageCacheManager.java:262)
	at org.hl7.fhir.utilities.npm.FilesystemPackageCacheManager.loadPackageFromCacheOnly(FilesystemPackageCacheManager.java:463)
	at org.hl7.fhir.utilities.npm.FilesystemPackageCacheManager.loadPackage(FilesystemPackageCacheManager.java:615)
	at org.hl7.fhir.validation.IgLoader.loadIg(IgLoader.java:108)
	at org.hl7.fhir.validation.IgLoader.loadIg(IgLoader.java:121)
	at org.hl7.fhir.validation.cli.services.ValidationService.loadIgsAndExtensions(ValidationService.java:549)
	at org.hl7.fhir.validation.cli.services.ValidationService.buildValidationEngine(ValidationService.java:495)
	at org.hl7.fhir.validation.cli.services.ValidationService.initializeValidator(ValidationService.java:476)
```

In practice with the validator-wrapper, I seem to recall other files causing issues but maybe it's only usage.ini. Other examples:
```
[...]

PR #1715, opened 2024-08-15
Package Cache Concurrency

dotasek:
This change reworks concurrency in the FileSystemPackageCacheManager.

Expanding on the test originally outlined in #1641, this adds a centralized series of locks that takes into account the lifecycle of the FileSystemPackageCacheManager and packages (creation, clear, package delete, etc), and provides additional utilities to ensure that all validation engines are using the same locks.


grahamegrieve (approved):
ok

PR #1717, opened 2024-08-16
Move all instantiation of transformerFactory to XMLUtils and set ACCESS_EXTERNAL flags automatically

### LINKED ISSUE #1571: Insecure Transformer used in XmlParsers
The Transformer used in XmlParsers (all versions) is insecure to XEE attacks (r5 shown here):

https://github.com/hapifhir/org.hl7.fhir.core/blob/dd9dd6b64aca4645aa3a444d011a6930d3acf58c/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/XmlParser.java#L146-L147

See https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#transformerfactory

The current Saxon-HE version (9.8) does not support XMLConstants.ACCESS_EXTERNAL_DTD or XMLConstants.ACCESS_EXTERNAL_SCHEMA (_java.lang.IllegalArgumentException: Unknown configuration property http://javax.xml.XMLConstants/property/accessExternalDTD_).

One way forward is to upgrade Saxon-HE to the latest version (9.8 → 12), which is anyway a good thing to do.

Another way is to use `newDefaultInstance()` instead of `newInstance()`, which returns the implementation provided by the JDK (Xalan) instead of Saxon.

Also, the use of a utility class that would configure XML factories would be nice, instead of reconfiguring them at each use (kind of what I did in Husky: https://github.com/project-husky/husky/blob/develop/husky-common/husky-common-gen/src/main/java/org/projecthusky/common/utils/xml/XmlFactories.java).

I can provide a PR if you agree with these solutions.

dotasek:
Fixes #1571

### DIFF
--- a/.github/workflows/license-check/license-whitelist.txt
+++ b/.github/workflows/license-check/license-whitelist.txt
@@ -10,6 +10,7 @@ Apache-2.0
 Apache 2
 Apache 2.0
 Apache License 2.0
+Apache License version 2.0
 Eclipse Public License v2.0
 BSD licence
 The BSD License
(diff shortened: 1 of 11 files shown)


grahamegrieve (approved):
ok

PR #1729, opened 2024-08-30
Added support for tracking whether elements are ellipsed and for rend…

lmckenzi:
…ering XHTML versions of XML and JSON content with ellipsed elements


grahamegrieve (approved):
ok

PR #1730, opened 2024-08-30
Update staging plugin

dotasek:
Update staging plugin and use `skipNexusStagingDeployMojo`, which is functionally similar, but more precisely documented than `skipStaging`

Source: https://help.sonatype.com/en/configuring-your-project-for-deployment.html

> skipNexusStagingDeployMojo
>
>Set to false by default, this flag will cause to skip any execution of the deploy goal of the plugin when set to true similar to maven.deploy.skip. In multi-module builds the staging of all components is performed in the last module based on the reactor order. If this property is set to true in the module, all staging will be skipped. You have to ensure that this property evaluates as true in the last module of the reactor. If necessary, you can add a dummy module.

>skipStaging

>Set to false by default this flag will cause to skip any execution of the plugin when set to true.

### DIFF
--- a/org.hl7.fhir.report/pom.xml
+++ b/org.hl7.fhir.report/pom.xml
@@ -89,7 +89,7 @@
                     <groupId>org.sonatype.plugins</groupId>
                     <artifactId>nexus-staging-maven-plugin</artifactId>
                     <configuration>
-                        <skipStaging>true</skipStaging>
+                        <skipNexusStagingDeployMojo>true</skipNexusStagingDeployMojo>
                     </configuration>
                 </plugin>
                 <plugin>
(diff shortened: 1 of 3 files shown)


dotasek:
The above description seems to indicate that the last module in the project dictates whether or not deployment happens via `skipNexusStagingDeployMojo`. If this is the case, org.hl7.fhir.report is the last module in the build, and this could cause deployment to be skipped altogether...

BUT when we deploy, we are skipping the following: -pl "!org.hl7.fhir.report, !org.hl7.fhir.validation.cli", the only modules where `skipNexusStagingDeployMojo` is used. So this should be OK, but should have a comment in the master and release pipelines.

PR #1735, opened 2024-09-03
Fix for ConcurrentModificationException for language translation

gjergjsheldija:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/LanguageUtils.java
```diff
@@ -296,7 +296,10 @@ private int importFromTranslations(Element parent, Element element, List<Transla
         }
       }
     }
-    for (Element c : element.getChildren()) {
+
+    // Create a copy of the children collection before iterating
+    List<Element> childrenCopy = new ArrayList<>(element.getChildren());
```
dotasek: Minor nitpick here; this could be `List.copyOf(element.getChildren());` which would ensure childrenCopy is immutable as well.

dotasek (approved):

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/LanguageUtils.java
gjergjsheldija: makes a lot of sense.
done
thank you :)

dotasek (approved):

PR #1739, opened 2024-09-10
Make sure this.messages is always initialized

pieter-edelman-nictiz:
Demonstration for initializing messages in the ProfileUtilities class. See #1740 for details.


grahamegrieve:
This is a problem because the error handling actually differs if messages == null, which is desired behaviour. (see ProfileUtilities line 2968).

But the way this is handled and we develop encourages us to forget that messages can be null, and that results in NPEs as you experienced, which is less than optimal. 

So while we won't be accepting we are thinking about how to refactor the code to improve matters

pieter-edelman-nictiz:
Thanks for the clarification @grahamegrieve . Is there something I can do to help with the refactoring?

grahamegrieve:
see https://github.com/hapifhir/org.hl7.fhir.core/pull/1747

PR #1741, opened 2024-09-10
Remove cqlframework dependencies

dotasek:
These dependencies don't actually appear to be used, and bring in numerous transitive dependencies.

They are explicitly used downstream in ig-publisher, but are included in its own pom.xml: https://github.com/HL7/fhir-ig-publisher/blob/9cade65f8ce5e061e4e224639bf2dd4433e038ec/org.hl7.fhir.publisher.core/pom.xml#L47-81


grahamegrieve (approved):
ok

PR #1743, opened 2024-09-11
Fix for cache init on existing directories

dotasek:
Cache was being cleared on initialization when the packages.ini file was found to be non-existent, or an older version.

Since the packages.ini is not generated by other tools (firely, etc), this was causing the cache to be cleared of any packages they installed. 

The only case in which the cache should be cleared is if the package cache version contained in packages.ini exists and is older than the current version.


dotasek:
Merging after approval by @grahamegrieve in Zulip

PR #1745, opened 2024-09-13
Package Cache lock cleanup

dotasek:
The creation of .lock files in order to protect the package cache from multiple processes (not just threads from within the same JVM) was incorrectly implemented, resulting in unlocked files being prematurely generated with no checks being done for incomplete package installs

The changes here should solve the following:

- .lock files should be generated, and the current process ID should be written to them _while locked_.
- There should be a mechanism to time out when lock files are not removed in a timely fashion.
- For testing and general flexibility, timeouts for locking should be implemented, and identifiable exceptions should be thrown.
- There should be a mechanism for identifying package cache installs that are incomplete (.lock files without a process lock)
- There should be a mechanism to clean up incomplete package cache installs on package cache instantiation.

Some nice-to-have functionality:

- Cleanup for incomplete package cache installs when they are detected after the cache is instantiated.


grahamegrieve (approved):
ok

PR #1746, opened 2024-09-13
Fix argument confusion in format

dotasek:
IntelliJ complained about this, and on investigation, this incorrectly formats the I18n output, leaving {{0}} in place of an actual `null`, which could be confusing. 

This will need the corresponding update to the test cases.


grahamegrieve (approved):
ok. but weird

grahamegrieve (commented):
ok

PR #1747, opened 2024-09-14
refactor error handling in ProfileUtilities

grahamegrieve:


pieter-edelman-nictiz:
Seems like an elegant solution ...

PR #1750, opened 2024-09-17
CapabilityStatement Import Processing

lmckenzi:
Adds a mechanism for determining an 'effective' CapabilityStatement where one statement imports another.  Handles SHALL/SHOULD/MAY assertions.  Includes test cases

Also see https://github.com/FHIR/fhir-test-cases/pull/191


grahamegrieve (approved):
ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/model/CodeType.java
```diff
@@ -66,7 +66,11 @@ public int compareTo(CodeType theCode) {
 			return 1;
 		}
 		return defaultString(getValue()).compareTo(defaultString(theCode.getValue()));
-	}	
+	}
+
+  public boolean equals(CodeType theCode) {
+    return this.getCode().equals(theCode.getCode());
+  }
```
grahamegrieve, line 73: How can just comparing the code be correct?

lmckenzi: If the data type is 'code', the system is implicit.  You can't have multiple codes with the same code value and different systems.  Version also doesn't come into play.  Something like 'id' isn't relevant, I don't think?  We could add extensions - is that what you're looking for?

Issue #1748, opened 2024-09-14
FHIR-46548: where clause on alias


oliveregger:
see [FHIR-46548](http://jira.hl7.org/browse/FHIR-46548) and [zulip thread](https://chat.fhir.org/#narrow/stream/379173-FHIR-Mapping-Language/topic/where.20clause)

```fml
group CapabilityStatementRest(source src, target tgt) extends BackboneElement {
  src.resource as s where (src.resource.type != 'CatalogEntry') log(src.resource.type) -> tgt.resource as t then CapabilityStatementRestResource(s, t);

// vs
  src.resource as s where (s.type != 'CatalogEntry') log(s.type) -> tgt.resource as t then CapabilityStatementRestResource(s, t);
}
```

Brian: Here src.resource is a repeating property. So the rule here is run for each resource.
If I just evaluate this rule without using the alias, it will evaluate based on src which is the parent of the item we're iterating.
Hence this will fail due to either: Performing a != comparison on a collection, the logging in the sample ...

PR #1752, opened 2024-09-18
FHIR-46548: where clause on alias #1748

oliveregger:
This PR adds source variable handling to FHIR Path expression for FHIR-46548: where clause on alias #1748.

The source variables are not prefixed with %, the FHIRPath expressions need to be adapted for that, this is done by not caching the FHIRPath expression first and then when the source variables are known, the expression is adapted and cached, following the same approach as @brianpos has done in https://github.com/brianpos/fhir-net-mappinglanguage/blob/develop2/Hl7.Fhir.MappingLanguage/JavaStructureMapUtils-Execute.cs.

In addition the source variable in the iteration has been added that the where, check and log clause are working on with the current iteration variable.

a test case has been created, for that the https://github.com/FHIR/fhir-test-cases/pull/190/files needs to be integrated first, before the test can run successfully.


oliveregger:
reverted the source variables prefixes logic, this is not necessary end even let fail the subsequent tests.

grahamegrieve (approved):
ok

PR #1765, opened 2024-10-04
Refactor package cache + more concurrency improvements

dotasek:
* Make deleting packages atomic (first rename to temp directory, then delete)
* Apply indexing and renaming code operations to temporary package install before rename
* Track last query to CI build server. Should refresh only when an hour has elapsed


grahamegrieve:
I don't remember where this ended up?

dotasek:
I do not recall either; I'll have to review where this was before we started on the ManagedWebAccess refactor.

dotasek:
@grahamegrieve this is ready to review now (providing the checks pass, which they should)

grahamegrieve (approved):
ok

PR #1769, opened 2024-10-06
Update maven wrapper version

vadi2:
Update mvnw - the linked version was 8 years old, and I was getting a warning about the maven version it downloaded no longer being compatible.


vadi2:
@dotasek done.

dotasek (approved):

PR #1774, opened 2024-10-09
Move new DocumentBuilderFactory calls to XMLUtils

dotasek:
Have a single utility method for generating DocumentBuilderFactory that always sets "http://apache.org/xml/features/disallow-doctype-decl" to true, and `setXIncludeAware(false)`


grahamegrieve (approved):
ok

Issue #1777, opened 2024-10-11
FML: Use FMLParser in StructureMapUtilities and support for identity transform


oliveregger:
There are currently mutliple parsers for FML, FmlParser which is used by the FHIR Validator and IG Publisher and one in StructureMapUtilities for each FHIR version.

They have slight different behaviour 
- FmlParser creates a DefaultAnonymousGroup while the one in StructureMapUtilities did not
- FmlParser has different comment handling

Will provide a PR which uses FmlParser also in StructureMapUtilities for R5 and that the identity transform is running through:
https://github.com/FHIR/fhir-test-cases/pull/195

In addition FmlParser will be extended with:
- multiline support for description as described in the spec https://hl7.org/fhir/mapping-language.html#metadata
- support for backtick, quote in source, target rules for the element
- log expression

PR #1778, opened 2024-10-11
#1777: FML Parser in StructureMapUtilities and id transform

oliveregger:
That the test run through the PR https://github.com/FHIR/fhir-test-cases/pull/195 needs to be integrated.

See description for PR in https://github.com/hapifhir/org.hl7.fhir.core/issues/1777

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/FmlParser.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/FmlParser.java
@@ -82,7 +82,29 @@ public class FmlParser extends ParserBase {
         String fid = lexer.takeDottedToken();
         Element e = result.makeElement(fid).markLocation(lexer.getCurrentLocation());
         lexer.token("=");
-        e.setValue(lexer.readConstant("meta value"));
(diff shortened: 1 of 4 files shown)


grahamegrieve (commented):
what's with this bit:
````java
  if (!(context.getVersion().equals("5.0.0"))) {
    throw new FHIRException("FHIR version needs to be 5.0.0");
  }
````
     
     ??

grahamegrieve (commented):
Also, ObjectConverter has a method to do the ElementModel -> Resource conversion

oliveregger:
> what's with this bit:
> 
> ```java
>   if (!(context.getVersion().equals("5.0.0"))) {
>     throw new FHIRException("FHIR version needs to be 5.0.0");
>   }
> ```
> 
> ??

I added this, because we use the R5 functionality in an R4 instance. StructureMapUtilities does not work with an R4 context and after noticing the problem, I added this safeguard. but no problem in removing it.

oliveregger:
the narrative generation has also changed with FML Parser, that the test run through, I updated the testcase here:
https://github.com/FHIR/fhir-test-cases/pull/211

grahamegrieve (approved):
ok

PR #1782, opened 2024-10-15
base workercontext adjustements

oliveregger:
added support for NamingSystem and adapting actors, requirements, questionnaire and search parameters.


grahamegrieve (approved):
thanks

PR #1785, opened 2024-10-18
Adding missing lombok anntation (cannot build without it from scratch)

jkiddo:


dotasek:
I'm not sure I follow here; all of the pipelines are building just fine without this in place, and they always start on a completely new VM. 

What situation produces the issue that this solves?

oliveregger:
> I'm not sure I follow here; all of the pipelines are building just fine without this in place, and they always start on a completely new VM.
> 
> What situation produces the issue that this solves?

for me it was a local maven build in the console for a dependent build (building upon org.hl7.fhir.core), strangely in the IDE (VSCode) I did not have any issue

dotasek (approved):
This is consistent with the lombok docs:

https://projectlombok.org/setup/maven

If it fixes problems, it's likely the right solution. In any case, isn't impacting the build negatively. I'm surprised I haven't run into it before now.

dotasek:
Validation tests are failing for unrelated fhir-test-cases reasons.

PR #1788, opened 2024-10-25
Manage FhirRequestBuilders via ManagedWebAccess

dotasek:
Rework ManagedWebAccess and the various FhirRequestBuilders to ensure the following:

* FhirRequestBuilders and SimpleHTTPClient should always use ManagedWebAccess to:
  * Set the 'User-Agent'
  * Set appropriate 'Authorization' or 'Api-Key' headers


dotasek:
Merging after Zulip confirmation from @grahamegrieve

PR #1793, opened 2024-10-28
Fix expected generated code

dotasek:


dotasek:
The original PR for this change was merged to a non-master branch.

PR #1795, opened 2024-10-30
fhirpath `tostring` implementation fix

brianpos:
Update the implementation of the `toString` fhirpath function to: **(pending FHIR-48737 decision)**
* not return the text 'null' when no primitive value exists (when extension only)
* throw an error if more than 1 item is to be returned (doesn't support collections)
* return an empty set when no primitive values exist (when extension only)

Applied to the R4, R4B and R5 fhirpath engines.
Includes a unit to test the results

Also some readme notes for windows developers and a fix for the output of a unit test message.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,10 @@ To skip unit tests:
 ```
 mvn -Dmaven.test.skip install
 ```
+> **Note:** If you're on Windows and use PowerShell, The `-` needs to be escaped with a backtick (`)
+> ```
+> mvn `-Dmaven.test.skip install
+> ``` 
 
 To clean and rebuild the terminology server caches:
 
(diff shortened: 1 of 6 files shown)


grahamegrieve:
I'm not entirely thrilled with this, because "throw an error if more than 1 item is to be returned" reduces the utility of .toString() for debugging. Should we add a new function debugString() or something that is more robust?

brianpos:
> I'm not entirely thrilled with this, because "throw an error if more than 1 item is to be returned" reduces the utility of .toString() for debugging. Should we add a new function debugString() or something that is more robust?

You can always do `.trace('blah', join(',').toString())` if you want to see a comma seperated list of values in the string while debugging - which is roughly what the java code is doing.

And that part of the fhirpath spec is normative too.

grahamegrieve (approved):
ok

PR #1805, opened 2024-11-11
Use server type to select server settings in combination with URL

dotasek:
This allows clients to pick specific server types for matching from FHIR Settings.

For example, the package client supports either npm-package or fhir-package servers, and needs to know which type of server it's dealing with when sending some requests.


grahamegrieve (approved):
ok

PR #1808, opened 2024-11-13
Fix mode based header addition in TxTester

dotasek:


grahamegrieve (commented):
doesn't this change replace the existing headers? We don't want to replace them, only add one header (and then remove it)

dotasek:
I can't see anywhere else the headers have been set, though. They look to be initialized and then aren't touched. Any headers added by FHIRToolingClient or ManagedWebAccess will be automatically added when the http request is made.

grahamegrieve:
but if some part of the code elsewhere adds a header, won't "terminologyClient.setClientHeaders(new ClientHeaders(List.of(header)));" delete that as a side effect?

dotasek:
@grahamegrieve what we discussed (addHeader, removeHeader, etc) looks like it's probably a bigger refactor than this. 

In the interim, I'm returning an Iterable<HTTPHeader> instead. Anywhere where someone was going to getClientHeaders().addHeader(...) would no longer work. And everything works. No 

As for setClientHeaders() possibly overwriting some previous header, it's only called in three places in the entire codebase... once if there's a header to set, and then immediately after to remove it before the next test (in a catch statement too, in case the previous didn't get called).

grahamegrieve (approved):
ok

PR #1809, opened 2024-11-13
Fix deprecated discriminator type warning message

cmoesel:
The warning related to the deprecated `pattern` discriminator type incorrectly said to use `type=fixed`, but `fixed` is not a valid discriminator type. This PR corrects it so it says to use `type=value`.


grahamegrieve (approved):
thanks

PR #1813, opened 2024-11-17
Show deprecation in snapshot and key

lmckenzi:
Needed for https://github.com/HL7/fhir-ig-publisher/issues/567


grahamegrieve (commented):
tgt.removeExtension(EXT_IG_DEPENDSON_REASON); - this isn't right. pass EXT_IG_DEPENDSON_REASON to the base copy() so it doesn't get copied in the first place

PR #1811, opened 2024-11-15
Fix IG dependsOn.reason conversion

lmckenzi:

PR #1818, opened 2024-11-19
Show deprecation in snapshot and key

lmckenzi:
Re-created https://github.com/hapifhir/org.hl7.fhir.core/pull/1813 without changes from https://github.com/hapifhir/org.hl7.fhir.core/pull/1811


grahamegrieve (approved):
ok

PR #1819, opened 2024-11-19
Automate testing for XMLUtils factory methods

dotasek:
There are already helper methods in place in XMLUtils to address potential XXE vulnerabilities. 

This PR does two things:

1. Verifies that those helper methods actually catch 'evil' xml and xslt.
2. Uses the checkstyle plugin to disallow instantiation of those various factories anywhere except where they are called in XMLUtils.


grahamegrieve (approved):
great

PR #1823, opened 2024-11-21
Remove IPS NZ support

dotasek:
IPS NZ is no longer being maintained, so the CLI param to automatically set up IPS NZ validation is no longer needed.


grahamegrieve (approved):
ok

PR #1824, opened 2024-11-21
Add gson annotations for fields and getter/setters

dotasek:
At present only Jackson is supported as a JSON serializer/deserializer. This adds annotations that will allow gson to also be used.


dotasek:
This is reviewed in a downstream project here: https://github.com/hapifhir/org.hl7.fhir.validator-wrapper/pull/195#pullrequestreview-2452847735

PR #1839, opened 2024-12-02
Refactor processTxIssues

dotasek:
The processTxIssues method in InstanceValidator has frequently resulted in confusion for maintainers.

This refactor will maintain the same functionality while providing some clarity through method naming and documentation.


grahamegrieve (approved):
ok

Issue #1844, opened 2024-12-04
NPE in DecimalType


dotasek:
In the following code, if theString is null, it will result in an NPE

https://github.com/hapifhir/org.hl7.fhir.core/blob/f5e218c279485441caf9b560ecd9f49684028736/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/model/DecimalType.java#L151-L151

PR #1845, opened 2024-12-04
Fix NPE in DecimalType

dotasek:
Fix for #1844 

~~Draft will include suggested fix for one FHIR version. Once approved, it will be applied to all FHIR versions.~~

Update: see comments. Suggested fix has been approved, applied to all versions, and this is ready for review.


grahamegrieve (approved):
ok

PR #1851, opened 2024-12-08
Value set interversion fix

lmckenzi:
The extension was using a child extension of "value" rather than "value[x]", which is what we'd agreed to.  Also added support for this in R4B


grahamegrieve (approved):
ok

PR #1852, opened 2024-12-09
Fix issue 1848

### LINKED ISSUE #1848: Validator Bundle entry reference resolution
When validating a bundle that contains versioned references, the validator is not able to find the matching entry based on the fullUrl leading to the incorrect warning / error that the entry is not present in the bundle.

Attached is a zip that contains a simple bundle example in JSON that leads to the warnings / errors and the validation output as a HTML file.
warning: Entry 'http://example.org/Observation/ObservationExample' isn't reachable by traversing forwards from the Composition. Only Provenance is approved to be used this way (R4 section 3.3.1)
error: Can't find 'http://example.org/Observation/ObservationExample/_history/1' in the bundle (Section Entry)

This seems to arise because BaseValidator.resolveInBundle does not take into account versioned references and the resolution for references in bundles https://hl7.org/fhir/R4/bundle.html#references .

[bundle-test.zip](https://github.com/user-attachments/files/18039241/bundle-test.zip)

sebg-mio42:
Adds the versioned reference to a resource (if meta.versionID is present) to the maps used to resolve entries in the bundle. 

Resolves #1848

### DIFF
--- a/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/BaseValidator.java
+++ b/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/BaseValidator.java
@@ -168,6 +168,7 @@ public class BaseValidator implements IValidationContextResourceLoader, IMessagi
   protected final String TYPE = "type";
   protected final String BUNDLE = "Bundle";
   protected final String LAST_UPDATED = "lastUpdated";
+  protected final String VERSION_ID = "versionId";
 
   protected BaseValidator parent;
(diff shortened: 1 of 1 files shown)


grahamegrieve (commented):
this one needs matching tests for success and failure in the test cases repository

sebg-mio42:
Thank you for the feedback!
Test cases have been added: https://github.com/FHIR/fhir-test-cases/pull/210

grahamegrieve (approved):
ok

PR #1858, opened 2024-12-12
If a base is not provided for Turtle generation, default to the fhir namespace

balhoff:
This matches the RDF output found in the R4 Turtle examples, rather than using a blank node for the main resource. This also fixes output of missing `fhir:link` values (fixes https://github.com/w3c/hcls-fhir-rdf/issues/121).

I attached a diff of the output for all FHIR examples: 
[issue-121-diff.txt](https://github.com/user-attachments/files/18116246/issue-121-diff.txt)

I am also attaching a folder of all the Turtle outputs from this change:
[fhir-examples-out-issue-121.zip](https://github.com/user-attachments/files/18116258/fhir-examples-out-issue-121.zip)


grahamegrieve (approved):
ok

PR #1883, opened 2025-01-15
2025 01 gg ai

grahamegrieve:


org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/filesystem/ManagedFileAccess.java
```diff
@@ -129,6 +129,22 @@ public static File file(String path, String filepath) throws IOException {
       throw new IOException("Internal Error");
     }
   }
+
+  public static File file(File root, String filepath) throws IOException {
```
dotasek, line 133: This looks like duplication. I think the following functions the same way without duplicating the logic:

```java
  public static File file(File root, String filepath) throws IOException {
   return file(root.getAbsolutePath(), filepath);
  }
 ```

File: org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ai/AIAPI.java [hunk 2]
```diff
@@ -0,0 +1,20 @@
...
+import java.util.List;
+
+public abstract class AIAPI {
+  
+  public abstract List<CodeAndTextValidationResult> validateCodings(List<CodeAndTextValidationRequest> requests) throws IOException;
+  
+
+  protected String getSystemName(String system) {
+    switch (system) {
+    case "http://snomed.info/sct": return "SNOMED CT";
```
dotasek, line 13: Are these always http and never https?

org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/tests/ValidationEngineTests.java
```diff
@@ -285,4 +294,149 @@ public static void execute() throws Exception {
...
+      ve.setFetcher(fetcher);
+      ve.getContext().setLocator(fetcher);
+      ve.setPolicyAdvisor(fetcher);
+      fetcher.setReferencePolicy(ReferenceValidationPolicy.CHECK_VALID);
+      fetcher.setResolutionContext("file:"+folder);
+      ve.seeResource(new JsonParser().parse(TestingUtilities.loadTestResourceStream("validator", "resolution", "StructureDefinition-Observation.json")));
+      ve.seeResource(new JsonParser().parse(TestingUtilities.loadTestResourceStream("validator", "resolution", "StructureDefinition-Patient.json")));
+      OperationOutcome op = ve.validate(FhirFormat.JSON, TestingUtilities.loadTestResourceStream("validator", "resolution", "relative-url-invalid.json"), null);
+      Assertions.assertTrue(checkOutcomes("testResolveRelativeFileValid", op, 
+          "Observation.subject null error/structure: Unable to find a profile match for Patient/example-newborn among choices: http://hl7.org/fhir/test/StructureDefinition/PatientRule\n"+
```
dotasek, line 337: The error message here is a little hard to follow. It can't find the subject because the reference is `Patient/example-newborn` and the closest available ID is `newborn`, which means the mentioned profile is not satisfied. Is that the correct interpretation?

org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/tests/ValidationEngineTests.java
```diff
@@ -285,4 +294,149 @@ public static void execute() throws Exception {
...
+      ve.setFetcher(fetcher);
+      ve.getContext().setLocator(fetcher);
+      ve.setPolicyAdvisor(fetcher);
+      fetcher.setReferencePolicy(ReferenceValidationPolicy.CHECK_VALID);
+      fetcher.setResolutionContext("file:"+folder);
+      ve.seeResource(new JsonParser().parse(TestingUtilities.loadTestResourceStream("validator", "resolution", "StructureDefinition-Observation.json")));
+      ve.seeResource(new JsonParser().parse(TestingUtilities.loadTestResourceStream("validator", "resolution", "StructureDefinition-Patient.json")));
+      OperationOutcome op = ve.validate(FhirFormat.JSON, TestingUtilities.loadTestResourceStream("validator", "resolution", "relative-url-error.json"), null);
+      Assertions.assertTrue(checkOutcomes("testResolveRelativeFileValid", op, 
+          "Observation.subject null error/structure: Unable to resolve resource with reference 'patient/example-newborn-x'\n"+
```
dotasek, line 363: How is this different from example-newborn when Patient/newborn should be the only resource found?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/utils/NodeStack.java
```diff
@@ -64,7 +64,12 @@ public NodeStack(IWorkerContext context, Element element, String refPath, String
     this.context = context;
     ids = new HashMap<>();
     this.element = element;
-    literalPath = refPath + "->" + element.getName();
+    int i = element.getName().indexOf(".");
+    if (i == -1) {
+      literalPath = refPath+".resolve().ofType(" + element.getName()+")";
```
dotasek, line 69: Do elements have path formatted names?

Patient.name with refpath `Some.ref.path` becomes `Some.ref.path.resolve().ofType(Patient).name`?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
```diff
@@ -1015,6 +1023,30 @@ public void validate(Object appContext, List<ValidationMessage> errors, String p
     codingObserver.finish(errors, stack);
     errors.removeAll(messagesToRemove);
     timeTracker.overall(t);
```
dotasek, line 1025: Should AI run after time tracking for overall validation, or should AI be included in overall time?

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/filesystem/ManagedFileAccess.java
grahamegrieve: ok

org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/tests/ValidationEngineTests.java
grahamegrieve: it resolves to a resource that is not valid against the profile it is required to conform to. Agree they are hard to follow because what is being tested is an obscure corner case

org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/tests/ValidationEngineTests.java
grahamegrieve: because it won't find any resource in this case

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/utils/NodeStack.java
grahamegrieve: well, they do. It was -> instead of .resolve() which is illegal, so I'm just normalising that. I looked at reworking the way element paths happen completely but it was a harder problem than I expected, so I backed out those changes. maybe another day

PR #1898, opened 2025-02-05
Optimize the JSON parsing in NpmPackageIndexBuilder.seeFile

qligier:
While working on some performance issues in Matchbox, I noticed that `NpmPackageIndexBuilder.seeFile` was using a lot of resources (time and memory allocation). The reason is that it fully parses many JSON files, allocating memory for objects representing the whole file, just to read a few properties and discard the object immediately.

This is a use case where streaming parsers shine: they only parse and allocate memory for the parts we're interested in. 

This PR uses the streaming parser provided in Jackson.

In Matchbox startup, we measured a reduction of ~6 seconds (12%) and ~7 Go of allocation (23%), when loading `hl7.fhir.r4.core`, `hl7.terminology.r4` and `hl7.fhir.uv.extensions.r4`.


grahamegrieve:
The R4B test failures look related to this

qligier:
Yep, I fixed it.

dotasek (approved):

PR #1915, opened 2025-02-26
Use Turtle shorthand syntax for booleans, integers, and decimals

balhoff:
Fixes https://github.com/w3c/hcls-fhir-rdf/issues/154

I set up this branch with a commit of all Turtle examples before and after the code change, so it should be easy to check the output diff there: https://github.com/balhoff/org.hl7.fhir.core/commit/6a07c45f54745987f6d6a579476a5577fde1b363


grahamegrieve (approved):
well, ok

PR #1916, opened 2025-02-26
Rename http.disabled + refactor proxy setting

dotasek:
The CLI param for disabling http access is now renamed to better match the naming patterns of existing CLI params.

In addition, the somewhat confusing and very Java-specific proxy code has been moved to its own class.


grahamegrieve (approved):
looks good to me

PR #1929, opened 2025-03-12
Fix copy of BaseWorkerContext.libraries into itself

qligier:
There's a typo in `BaseWorkerContext` in R4B and R5, the `copy` method will copy the `libraries` field into itself instead of copying from the `other` instance.


grahamegrieve (approved):
ok. thx

PR #1938, opened 2025-03-22
Feat/slash in folder

jkiddo:
Added more tests NpmPackage parsing


jkiddo:
Some harness around https://github.com/hapifhir/org.hl7.fhir.core/commit/3c7d13fe9223e45a776160ae244748cf513fd541

grahamegrieve:
Note that cause of the failure - something in your packages. You can investigate as to whether the bi-di check is correct or not, or just withdraw the PR, since the actual problem is fixed now

jkiddo:
bi-dir already ignores other package.tgz files - probably because its binary content: https://github.com/hapifhir/org.hl7.fhir.core/blob/d86ef5fa80e13ac47678e8a67a40a0b535e2bdee/.github/workflows/bidi-checker.yml#L28 - so that could be expanded to include those in the PR as well.

grahamegrieve:
Well, do you want to make those changes then?

PR #1941, opened 2025-03-26
Handle hashes

jamesagnew:
Handle FHIRPath `resolve()` function when contained resource IDs include hash mark (as is the case from HAPI's base parser)

### DIFF
--- a/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/fhirpath/FHIRPathEngine.java
+++ b/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/fhirpath/FHIRPathEngine.java
@@ -5483,7 +5483,7 @@ public class FHIRPathEngine {
           Property p = context.rootResource.getChildByName("contained");
           if (p != null) {
             for (Base c : p.getValues()) {
-              if (t.equals(c.getIdBase())) {
+              if (chompHash(s).equals(chompHash(c.getIdBase()))) {
                 res = c;
                 break;
               }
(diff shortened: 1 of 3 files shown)


grahamegrieve:
I'm bothered by this one - this is part of the validator, and I think it will start accepting invalid internal references if this change is made as is. Do we have test cases ?

grahamegrieve (changes requested):
I'm bothered by this one - this is part of the validator, and I think it will start accepting invalid internal references if this change is made as is. Do we have test cases ? We definitely need them somewhere, at least to ensure that whatever we do doesn't change the validator behaviour

jamesagnew:
@grahamegrieve The problem is that HAPI's parser puts the `#` in the ID of parsed contained resources.

So for example, if I parse a Patient with a contained Organization, HAPI's parser will effectively do: `theOrganization.setId("#1")` where presumably your parser does `theOrganization.setId("1")`.

This may or may not have been a good decision ,but that's the way it has worked for 10 years now and people rely on that behaviour. When you removed the `chomp()` from the FHIRPath evaluator back in december, it basically means that we can no longer upgrade to the latest core so HAPI users are stuck using an older version of the FP/validator stack.

I've added a test `FHIRPathEngineTest` which demonstrates this.

I understand your hesitance to revert to the old chomp behaviour if you're worried about it letting people in IG Publisher and other stacks to illegal things.. I'm not sure what to do about it though. I suppose I could make this behaviour configurable so that only HAPI enables it if you'd be more comfortable that way.

jamesagnew:
Update: New approach. Fixing the DSTU3 evaluator to work like the rest, and I'll declare this a breaking change in HAPI FHIR 8.2.0.

grahamegrieve (approved):
ok

Issue #1942, opened 2025-03-26
liquid template generation with lookup function


oliveregger:
according to https://build.fhir.org/ig/FHIR/ig-guidance/testfactory.html#data-lookup the lookup function takes three parameters: lookup(lookupCol, value, outputCol) looks up a value in lookupCol, and returns the value in outputCol (or null) where the name of the data table can be prefixed.

running this returns however an error:

```
Exception in thread "main" org.hl7.fhir.exceptions.FHIRException: org.hl7.fhir.r5.fhirpath.FHIRLexer$FHIRLexerException: Error @1, 2: The function "lookup" requires between 4 and 4 parameters
	at org.hl7.fhir.r5.testfactory.TestDataFactory.executeLiquid(TestDataFactory.java:423)
	at org.hl7.fhir.r5.testfactory.TestDataFactory.execute(TestDataFactory.java:239)
	at org.hl7.fhir.validation.cli.services.ValidationService.instanceFactory(ValidationService.java:969)
```

specifying for parameters (with providing the table as the code indicates https://github.com/hapifhir/org.hl7.fhir.core/blob/master/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/testfactory/TestDataFactory.java#L179 does also not work).

will add a PR that the lookup function works for liquid template as documented.

oliveregger:
the only drawback with this approach is, that the variable row is empty during the lookup call and I need to assign a tmp variable before the lookup call:

 {% assign genderfhir = row.gender %} 
gender: {{ genderfhir }} - Gender V2: {{ genders.lookup(2,genderfhir, 1) }} 

whereas:

gender: {{ row.gender }} - Gender V2: {{ genders.lookup(2, row.gender, 1) }} 

returns no lookup

PR #1943, opened 2025-03-26
liquid template generation with lookup function #1942

oliveregger:
see #1942 for details


grahamegrieve (commented):
do we have test cases for this change?

oliveregger:
i added the example to the fhir-guidance ig see [add lookup functionality in liquid template FHIR/ig-guidance#49](https://github.com/FHIR/ig-guidance/pull/49), but I/we can certainley also add a testcase, we will get back

PR #1944, opened 2025-03-27
FML resolve() in Bundle

mrunibe:
Proof-of-concept to support resolve() in FML to resolve references within a Bundle.

Related zulip discussion: https://chat.fhir.org/#narrow/channel/379173-FHIR-Mapping-Language/topic/resolve.28.29.20references.20within.20a.20Bundle/with/508251737


mrunibe:
I simplified the PR further by just lazy caching the references in TransformSupportServices, no longer requiring modifictions in ValidationEngine.

grahamegrieve (commented):
I'm suspicious of this change for two reasons - the bundle resolution code is more complicated than this, I think (more ways to resolve than the simple one I see here) 

But more importantly, this is just inside bundles? I think we need to agree what the spec should say should happen with resolve(), because in some use cases for FML, resolve() is a very significant problem - I think?

mrunibe:
> I'm suspicious of this change for two reasons - the bundle resolution code is more complicated than this, I think (more ways to resolve than the simple one I see here)
> 

I agree. @oliveregger suggested to use the mechanism from the [InstanceValidator.resolveReference()](https://github.com/ahdis/matchbox/pull/362) instead. It's now implemented in [matchbox](https://github.com/ahdis/matchbox/pull/362) and I have [changed ](https://github.com/hapifhir/org.hl7.fhir.core/pull/1944/commits/d0cd55efd93a2e8ab059396088aaf884fcf504bd)this PR accordingly. FML would then use the same logic as validate.

> But more importantly, this is just inside bundles? I think we need to agree what the spec should say should happen with resolve(), because in some use cases for FML, resolve() is a very significant problem - I think?

Indeed the spec does not say much about _resolve()_ except referring to FHIRPath. The example in the tutorial seems incoherent, see related [zulip discussion](https://chat.fhir.org/#narrow/channel/379173-FHIR-Mapping-Language/topic/resolve.28.29.20references.20within.20a.20Bundle/with/508923721).

Now that we use _InstanceValidator.resolveReference()_ it should more generic. Additionally, if we would initialize it with:
```java
validator.setFetcher(new StandAloneValidatorFetcher(pcm, context, this));
```
It even resolves external resource (I have successfully tested it locally).

grahamegrieve:
so how does this PR compare to the discussion that we had on Zulip about this issue?

mrunibe:
> so how does this PR compare to the discussion that we had on Zulip about this issue?

This PR **enables resolve() on the target variable using FHIRPath** which can then be used in dependent rules:

`src.prac as prac -> execute(prac.resolve()) as pracResource then { ... }
`

Participants of the discussion agreed this syntax is valid and should work.

The Zulip discussion was also whether resolve() should be possible on the source element (like `src.prac.resolve()`) which was dismissed (JIRA ticket closed). Consequently, the PR does not change that.

PR #1946, opened 2025-03-28
DO NOT MERGE - Tests for logging framework improvement

dotasek:
This draft PR demonstrates several things we can do to move to slf4j logging in the core library (and downstream projects).

`SystemOutTests` illustrates how we can use carriage returns to track progress without polluting logs.
`LoggingTest` illustrates programmatic setting of log levels (so we can still use the `-debug` option in the CLI, and maybe add '-trace`) as well as programmatic setting of logback config files.

The CLI classes are now partially moved into the org.hl7.fhir.validation.cli module; this is intentional, as it splits the validator CLI from the actual validation logic, enabling the CLI to access logback, while leaving the logic with only slf4j, which is important for anyone integrating org.hl7.fhir.validation into their project.


dotasek:
Closing. A full PR is in progress here: https://github.com/hapifhir/org.hl7.fhir.core/pull/2009

PR #1958, opened 2025-04-10
Move CLI to cli module

dotasek:
Move the ValidatorCLI to the org.hl7.fhir.validation.cli module.

Note that this is not a complete move. There are still many interacting classes that should be moved to the cli module, but that refactor will likely introduce some breaking changes to class names and packages.


grahamegrieve (approved):
ok

PR #1968, opened 2025-04-22
2025 04 gg uml generation

grahamegrieve:


org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/conv40_50/resources40_50/ConceptMap40_50.java
```diff
@@ -50,6 +50,7 @@ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWIS
 */
 // Generated on Sun, Feb 24, 2019 11:37+1100 for FHIR v4.0.0
 public class ConceptMap40_50 {
+;
```
dotasek: Spare ';' ?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/ConceptMapRenderer.java
```diff
@@ -385,6 +389,9 @@ public void render(RenderingStatus status, ResourceWrapper res, XhtmlNode x, Con
           }
         }
       }
+      if (props.size() > 0) {
+        properties = true;
```
dotasek: hasProperties would be more descriptive here.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/SnapshotGenerationPreProcessor.java
```diff
@@ -468,4 +469,119 @@ private SliceInfo getSlicing(ElementDefinition ed) {
...
+ 
+  private String makeTail(String[] pathCurrent, int start, int index) { 
+    CommaSeparatedStringBuilder b = new CommaSeparatedStringBuilder("."); 
+    for (int i = start; i <= index; i++) { 
+      b.append(pathCurrent[i]); 
+    } 
+    return b.toString(); 
+  }
+
+  public StructureDefinition trimSnapshot(StructureDefinition profile) {
```
dotasek, line 544: I only see this method used in the ClassDiagramRenderer. Could it be moved there?

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/XhtmlNode.java
```diff
@@ -1215,4 +1238,63 @@ public XhtmlNode supr(String tx) {
     return this;
   }
 
+  // -- SVG functions ------------
```
dotasek, line 1241: Could these be in an extension of this class called SvgNode?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/ClassDiagramRenderer.java
```diff
@@ -0,0 +1,1703 @@
...
+    return item;  
+  }
+
+  private void drawLink(XhtmlNode svg, Link l, XhtmlNode insertionPoint) throws FHIRException, IOException {
+    Point start;
+    Point end;
+    Point p1;
+    Point p2;
+    String id = l.source.id + "-" + l.target.id;
+    if (l.source == l.target) {
```
dotasek, line 1353: This if doesn't seem to be satisfied in the test cases, so the whole branch is missed. It also hits XhtmlNode.svgPath, which has some unused parameters, so maybe there's something here that's not quite right.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/XhtmlNode.java
```diff
@@ -1215,4 +1238,63 @@ public XhtmlNode supr(String tx) {
...
+    x.attribute("y", Double.toString(top));
+    x.attribute("width", Double.toString(width));
+    x.attribute("height", Double.toString(height));
+    return x;
+  }
+
+
+  public XhtmlNode svgPath(XhtmlNode insertionPoint, String height, String width) {
+    var x = addTag("path", insertionPoint);
+//    x.attribute("height", height);
```
dotasek: The height and width params are never used here. As mentioned elsewhere, this code is not used in the test cases, so it could potentially have some issues.

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/conv40_50/resources40_50/ConceptMap40_50.java
grahamegrieve: fixed

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/ConceptMapRenderer.java
grahamegrieve: renamed

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/SnapshotGenerationPreProcessor.java
grahamegrieve: it could be but I'm planning to use it elsewhere in the future

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/XhtmlNode.java
grahamegrieve: no because this is already complicated by the way that fluent api works on two subclasses, and the way the node tree switches into svg and then back to html again

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/ClassDiagramRenderer.java
grahamegrieve: it's true I forgot to remove those two parameters, and it's true that I don't have a test case for that combination yet in the JUnit tests. But it does this on the right: 
image

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/XhtmlNode.java
grahamegrieve: removed

dotasek (approved):

PR #1972, opened 2025-04-24
FhirPath parser fix for bug in Line numbering during parsing

brianpos:
When parsing an expression like this:
```
true
   and
      where(   glarb =   'true')
```
The whitespace isn't counted correctly causing the start positions to be incorrect.
The parser double counts whitespace (effectively calling the incColumn twice for whitespace).
)
Also included unit tests to verify the functionality


grahamegrieve (approved):
great thanks

PR #1978, opened 2025-04-29
R5 FHIRToolingClient should request 5.0 not 4.0 responses

lawley:
The FHIRToolingClient is explicitly asking for fhirVersion=4.0
This fixes the requests to explicitly ask for fhirVersion=5.0


grahamegrieve (approved):
agree

PR #1985, opened 2025-04-30
Subscription topic current rendering

caspears:
SubscriptionTopic redndering of trigger.queryCriteria.current uses a label of 'create {0}'. This PR adds the current label in the same pattern as 'previous', 'current: {0}'. 
An example rendering:
image

Can be found here: 
https://build.fhir.org/ig/HL7/davinci-pct/SubscriptionTopic-davinci-pct-aeob-available-author-subscriptiontopic.html


grahamegrieve (approved):
ok

PR #1987, opened 2025-05-02
Fix places where locale was set to null instead of using context

dotasek:
In many locations, the locale was set to null in the constructor when the calling class was using a non-null locale, resulting on a fallback to system locale.


org.hl7.fhir.utilities/src/main/resources/rendering-phrases_de.properties
```diff
@@ -4,3 +4,5 @@ BUNDLE_IF_NONE = If-None-Exist = {0}
 BUNDLE_LOCATION = Location = {0}
 BUNDLE_RESOURCE = Resource {0}:
 BUNDLE_SEARCH = Suche:
+CODESYSTEM_CONTENT_NOTPRESENT = Ich bin ein auslander
```
grahamegrieve: I really think we better fix these before we accept the PR

grahamegrieve (changes requested):
per the comment

org.hl7.fhir.utilities/src/main/resources/rendering-phrases_de.properties
dotasek: Good catch. Left over from testing. ~~There are things here that should be fixed, but this at least addresses the originally reported issue~~.

~~I will leave test cases out of this PR for now, as the correct behaviour of rendering phrases, plural rules, and default locales are out of the scope of the original issue~~.

grahamegrieve (approved):
ok

PR #1988, opened 2025-05-04
2025 05 gg go secure

grahamegrieve:


org.hl7.fhir.validation/src/test/resources/txCache/org.hl7.fhir.validation/5.0.0/servers.ini
```diff
@@ -1,4 +1,5 @@
 [servers]
 tx-dev.fhir.org.r5 = http://tx-dev.fhir.org/r5
 local.fhir.org.r5 = http://local.fhir.org/r5
+..tx-dev.fhir.org.r52 = https://tx-dev.fhir.org/r5
```
dotasek, line 4: Why is this being added? At a glance, it looks like the existing entry is being copied and incremented.

org.hl7.fhir.validation/src/test/resources/txCache/org.hl7.fhir.validation/4.3.0/servers.ini
```diff
@@ -1,4 +1,5 @@
 [servers]
 tx-dev.fhir.org.r4 = http://tx-dev.fhir.org/r4
 local.fhir.org.r4 = http://local.fhir.org/r4
+..tx-dev.fhir.org.r42 = https://tx-dev.fhir.org/r4
```
dotasek, line 4: Why is this being added? At a glance, it looks like the existing entry is being copied and incremented.

org.hl7.fhir.validation/src/test/resources/txCache/org.hl7.fhir.validation/4.0.1/servers.ini
```diff
@@ -1,4 +1,5 @@
 [servers]
 tx-dev.fhir.org.r4 = http://tx-dev.fhir.org/r4
 local.fhir.org.r4 = http://local.fhir.org/r4
+..tx-dev.fhir.org.r42 = https://tx-dev.fhir.org/r4
```
dotasek, line 4: Why is this being added? At a glance, it looks like the existing entry is being copied and incremented.

org.hl7.fhir.validation/src/test/resources/txCache/org.hl7.fhir.validation/3.0.2/servers.ini
```diff
@@ -1,4 +1,5 @@
 [servers]
 tx-dev.fhir.org.r3 = http://tx-dev.fhir.org/r3
 local.fhir.org.r3 = http://local.fhir.org/r3
+..tx-dev.fhir.org.r32 = https://tx-dev.fhir.org/r3
```
dotasek, line 4: Why is this being added? At a glance, it looks like the existing entry is being copied and incremented.

org.hl7.fhir.validation/src/test/resources/txCache/org.hl7.fhir.validation/1.4.0/servers.ini
```diff
@@ -1,4 +1,5 @@
 [servers]
 tx-dev.fhir.org.r3 = http://tx-dev.fhir.org/r3
 local.fhir.org.r3 = http://local.fhir.org/r3
+..tx-dev.fhir.org.r32 = https://tx-dev.fhir.org/r3
```
dotasek, line 4: Why is this being added? At a glance, it looks like the existing entry is being copied and incremented.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/utils/RenderingContext.java
```diff
@@ -764,8 +764,12 @@ public boolean hasLink(KnownLinkType link) {
     return links.containsKey(link);
   }
   
-  public String getLink(KnownLinkType link) {
-    return links.get(link);
+  public String getLink(KnownLinkType link, boolean secure) {
+    String url = links.get(link);
+    if (url != null && secure && url.startsWith("http://")) {
+      url = url.replace("http://", "https://");
```
dotasek: This will replace all http:// references, I think this should use a substring: `url = "https://" + url.substring("http://".length());` or something similar to limit the behaviour.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/ManagedWebAccess.java
```diff
@@ -145,4 +149,23 @@ public static void loadFromFHIRSettings(FhirSettings settings) {
     serverAuthDetails = new ArrayList<>();
     serverAuthDetails.addAll(settings.getServers());
   }
+
+  public static String makeSecureRef(String url) {
+    if (url == null || !url.startsWith("http://") || isLocal(url)) {
+      return url;
+    } else {
+      return url.replace("http://", "https://");
```
dotasek, line 157: Duplicated code in RenderingContext. This should go in some util class and get re-used in both places. Same commentary applies; we should only be replacing the first part of the url with https.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/ManagedWebAccess.java
```diff
@@ -145,4 +149,23 @@ public static void loadFromFHIRSettings(FhirSettings settings) {
...
+    }
+  }
+
+  private static boolean isLocal(String url) {
+    URI uri;
+    try {
+      uri = new URI(url);
+      // todo: check in serverAuthDetails to see if policy is set for this server
+      return Utilities.existsInList(uri.getHost(), "localhost", "local.fhir.org", "127.0.0.1", "[::1]") || (uri.getHost() != null && uri.getHost().endsWith(".localhost"));
```
dotasek, line 166: Where are "local.fhir.org" and ".localhost" coming from?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/model/Element.java
```diff
@@ -361,6 +362,29 @@ public Extension getExtensionByUrl(String theUrl) {
...
+      ArrayList<Extension> retVal = new ArrayList<Extension>();
+      for (Extension next : getExtension()) {
+        if (Utilities.existsInList(next.getUrl(), theUrls)) {
+          retVal.add(next);
+        }
+      }
+      if (retVal.size() == 0)
+        return null;
+      else {
+        org.apache.commons.lang3.Validate.isTrue(retVal.size() == 1, "Url "+CommaSeparatedStringBuilder.join(",", theUrls)+" must have only one match");
```
dotasek, line 383: Is this IllegalArgument exception intended to be caught, or will it blow up for specific cases?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/QuestionnaireRenderer.java
```diff
@@ -258,8 +258,8 @@ private Row addTreeRoot(HierarchicalTableGenerator gen, List<Row> rows, Resource
     return r;     
   } 
 
-  private String getSpecLink(String path) { 
-    return Utilities.pathURL(context.getLink(KnownLinkType.SPEC), path); 
+  private String getSpecLink(String path, boolean secure) {
```
dotasek: It looks like secure is always 'true' here.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/QuestionnaireResponseRenderer.java
```diff
@@ -566,7 +566,7 @@ public boolean renderForm(XhtmlNode x, ResourceWrapper q) throws UnsupportedEnco
   private boolean renderLinks(RenderingStatus status, XhtmlNode x, ResourceWrapper q) {
     x.para().tx(context.formatPhrase(RenderingContext.QUEST_TRY_QUEST));
     XhtmlNode ul = x.ul();
-    ul.li().ah("http://todo.nlm.gov/path?mode=ig&src="+Utilities.pathURL(context.getLink(KnownLinkType.SELF), "package.tgz")+"&q="+q.getId()+".json").tx(context.formatPhrase(RenderingContext.QUEST_NLM));
+    ul.li().ah("http://todo.nlm.gov/path?mode=ig&src="+Utilities.pathURL(context.getLink(KnownLinkType.SELF, false), "package.tgz")+"&q="+q.getId()+".json").tx(context.formatPhrase(RenderingContext.QUEST_NLM));
```
dotasek, line 569: 'http://todo.nlm.gov' doesn't resolve, so I'm assuming it is a special placeholder URL and doesn't need https.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/TerminologyClientManager.java
```diff
@@ -629,7 +630,7 @@ public SourcedValueSet findValueSetOnServer(String canonical) {
       TerminologyClientContext client = serverMap.get(server);
       if (client == null) {
         try {
-          client = new TerminologyClientContext(factory.makeClient("id"+(serverList.size()+1), server, getMasterClient().getUserAgent(), getMasterClient().getLogger()), cacheId, false);
+          client = new TerminologyClientContext(factory.makeClient("id"+(serverList.size()+1), ManagedWebAccess.makeSecureRef(server), getMasterClient().getUserAgent(), getMasterClient().getLogger()), cacheId, false);
```
dotasek, line 633: Will this not force all terminology servers to be https? For dev purposes, is this desireable?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java
```diff
@@ -3095,9 +3095,9 @@ private static Extension markExtensionSource(Extension extension, boolean overri
       extension.setUserData(UserDataNames.SNAPSHOT_EXTENSION_SOURCE, srcSD);
     }
     if (Utilities.existsInList(extension.getUrl(), ToolingExtensions.EXT_OBLIGATION_CORE, ToolingExtensions.EXT_OBLIGATION_TOOLS)) {
-      Extension sub = extension.getExtensionByUrl("source");
+      Extension sub = extension.getExtensionByUrl(ToolingExtensions.EXT_OBLIGATION_SOURCE, ToolingExtensions.EXT_OBLIGATION_SOURCE_SHORT);
       if (sub == null || overrideSource) {
-        ToolingExtensions.setUriExtension(extension, "source", srcSD.getVersionedUrl());
+        ToolingExtensions.setUriExtension(extension, ToolingExtensions.EXT_OBLIGATION_SOURCE, srcSD.getVersionedUrl());
```
dotasek, line 3100: Should this also be with ToolingExtensions.EXT_OBLIGATION_SOURCE_SHORT?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/ImplementationGuideValidator.java
```diff
@@ -73,8 +75,10 @@ private boolean checkDependency(List<ValidationMessage> errors, Element ig, Node
     ok = rule(errors, "2024-06-13", IssueType.BUSINESSRULE, dependency.line(), dependency.col(), stack.getLiteralPath(), packageId == null || packageId.matches(FilesystemPackageCacheManager.PACKAGE_REGEX), I18nConstants.IG_DEPENDENCY_INVALID_PACKAGEID, packageId) && ok;         
 
     try {
+      ImplementationGuide igD = context.fetchResource(ImplementationGuide.class, url);
+      warning(errors, "2024-06-13", IssueType.BUSINESSRULE, dependency.line(), dependency.col(), stack.getLiteralPath(), igD != null, I18nConstants.IG_DEPENDENCY_INVALID_URL, url);                   
       FilesystemPackageCacheManager pcm = new FilesystemPackageCacheManager.Builder().build();
-      if (url != null && packageId != null) {
+      if (url != null && packageId != null && (igD == null || !igD.hasUserData(UserDataNames.IG_FAKE))) {
```
dotasek: Maybe this would be clearer as `fetchedIgDependency != null || igD.hasUserData(UserDataNames.IG_FAKE)`

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/BaseValidator.java
```diff
@@ -1532,7 +1532,7 @@ protected boolean isExampleUrl(String url) {
     return 
         Utilities.containsInList(url, "example.org/", "acme.com/", "acme.org/", "example.com/", "example.net/") ||
         Utilities.endsWithInList(url, "example.org", "acme.com", "acme.org", "example.com", "example.net") ||
-        url.startsWith("urn:oid:1.3.6.1.4.1.32473.");    
+        url.startsWith("urn:oid:1.3.6.1.4.1.32473.") || url.startsWith("urn:oid:2.999.");
```
dotasek, line 1535: Is there some spec or testing documentation that says which urn's are examples? allow-example-urls only specifies example.org and nothing else.

dotasek: Ideally, that spec should be referenced in the JavaDoc for this method.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
```diff
@@ -537,7 +537,7 @@ public NpmPackage addPackageToCache(final String id, final String version, final
       log("");
       log("Installing " + id + "#" + version);
 
-      if ((extractedNpm.name() != null && id != null && !id.equalsIgnoreCase(extractedNpm.name()))) {
+      if ((extractedNpm.name() != null && id != null && !id.equalsIgnoreCase(extractedNpm.name()) && !id.equalsIgnoreCase(extractedNpm.name()+"."+VersionUtilities.getNameForVersion(extractedNpm.fhirVersion())))) {
         if (!suppressErrors && (!id.equals("hl7.fhir.r5.core") && !id.equals("hl7.fhir.us.immds"))) {// temporary work around
           throw new IOException("Attempt to import a mis-identified package. Expected " + id + ", got " + extractedNpm.name());
```
dotasek, line 542: This error could be misleading, as `!id.equalsIgnoreCase(extractedNpm.name()+"."+VersionUtilities.getNameForVersion(extractedNpm.fhirVersion()))))` isn't mentioned but is required to trigger the Exception.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/comparison/CodeSystemComparer.java
```diff
@@ -501,14 +501,24 @@ private void mergeProps(ConceptDefinitionComponent cd, ConceptDefinitionComponen
       } else {
         matchR.add(rp);
         cd.getProperty().add(lp);
-        if (lp.getValue().equalsDeep(rp.getValue())) {
-          cd.getProperty().add(rp.setCode(res.getPropMap().get(rp.getCode())));
+        if (!lp.getValue().equalsDeep(rp.getValue())) {
```
dotasek, line 504: Was the change to `!` here intentional? It reverses the condition in the original code.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/comparison/StructureDefinitionComparer.java
```diff
@@ -532,21 +533,50 @@ private boolean compareDiff(String path, String sliceName, DefinitionNavigator l
...
 
-  private boolean compareDiffChildren(String path, DefinitionNavigator left, DefinitionNavigator right, Base parent, ProfileComparison res) throws DefinitionException, IOException, FHIRFormatError {
+  private List<Base> newParents(List<Base> parents, ElementDefinition edr) {
+    List<Base> list = new ArrayList<Base>();
+    list.addAll(parents);
+    list.add(edr);
+    return list;
+  }
+
+  private boolean compareDiffChildren(String path, DefinitionNavigator left, DefinitionNavigator right, List<Base> parents, ProfileComparison res) throws DefinitionException, IOException, FHIRFormatError {
```
dotasek, line 545: `path` is not used here.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/comparison/StructureDefinitionComparer.java
```diff
@@ -390,12 +392,11 @@ private boolean compareChildren(ProfileComparison comp, StructuralMatch<ElementD
   }
 
 
-  private boolean compareDiff(String path, String sliceName, DefinitionNavigator left, DefinitionNavigator right, ProfileComparison res, Base parent) throws DefinitionException, FHIRFormatError, IOException {
+  private boolean compareDiff(String path, String sliceName, DefinitionNavigator left, DefinitionNavigator right, ProfileComparison res, List<Base> parents) throws DefinitionException, FHIRFormatError, IOException {
     assert(path != null);  
     assert(left != null);
     assert(right != null);
     assert(left.path().equals(right.path()));
```
dotasek, line 399: Based on other methods returning `parents.get(0)`, should we also have an assert of `parents.size() > 0`?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/SnapshotGenerationPreProcessor.java
dotasek, line 1: Is there a test demonstrating what all this new code is supposed to be doing? It's difficult to evaluate this change otherwise.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/DefinitionNavigator.java
dotasek, line 1: Is there an example or test for this new logic?

dotasek (changes requested):

org.hl7.fhir.validation/src/test/resources/txCache/org.hl7.fhir.validation/4.3.0/servers.ini
grahamegrieve: the caching algorithm added it. Why do we care too think about this?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/utils/RenderingContext.java
grahamegrieve: under what circumstances would http:// appear later in the URL and we would not want https:// replaced? I couldn't think of any, but maybe there's one, so I'll make this change

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/ManagedWebAccess.java
grahamegrieve: It is not duplicated even though it's the same thing right now - the code is in ManagedWebAccess because it's going to get dependent on server settings, where as the rendering code isn't going to subject to that

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/ManagedWebAccess.java
grahamegrieve: local.fhir.org is defined in fhir.org DNS and I use it routinely. .localhost - see the Zulip thread

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/model/Element.java
grahamegrieve: blow up. This is matching other similar code that works the same way

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/QuestionnaireRenderer.java
grahamegrieve: ok I'll change

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/QuestionnaireResponseRenderer.java
grahamegrieve: indeed. I made myself a note to follow up on that

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/TerminologyClientManager.java
grahamegrieve: I discussed this change with all other terminology server authors. But one reason it's in ManagedWebAccess is so that you can turn this off in the http settings file (though I didn't hook that up yet)

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java
grahamegrieve: no this is a deliberate change.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/ImplementationGuideValidator.java
grahamegrieve: ok

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/BaseValidator.java
grahamegrieve: There's no spec to my knowledge.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
grahamegrieve: That's true, but I'm not sure how to explain that, but even more, I'm not sure that it'll ever matter - it's a hack workaround for an already released package that is wrong and was addressed elsewhere

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/comparison/CodeSystemComparer.java
grahamegrieve: it is, yes. The old code cloned the property when merging if and only if the property value was the same, but overwrite the property if the value was different. backwards.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/comparison/StructureDefinitionComparer.java
grahamegrieve: it's there for debugging.

org.hl7.fhir.validation/src/test/resources/txCache/org.hl7.fhir.validation/4.3.0/servers.ini
dotasek: Actually, I just noted that this is with the new `https://` fix, which makes sense. But, our tests are looking for unexpected cache hits, maybe that original `http://` entry shouldn't be there anymore. If we regenerate caches at this point, would `http://tx-dev.fhir.org/r4` show up?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/comparison/StructureDefinitionComparer.java
grahamegrieve: well ok, but it could never be empty

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/SnapshotGenerationPreProcessor.java
grahamegrieve: yes there is, and it's hard code. The R5 snapshot generation test 'multi-profile'

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/DefinitionNavigator.java
grahamegrieve: same snapshot test

org.hl7.fhir.validation/src/test/resources/txCache/org.hl7.fhir.validation/4.3.0/servers.ini
grahamegrieve: it should not.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/SnapshotGenerationPreProcessor.java
```diff
@@ -116,7 +128,567 @@ public SnapshotGenerationPreProcessor(ProfileUtilities utils) {
...
+        if (p2.getValues().size() > 1) {
+          throw new Error("Not supported");
+        }
+        merged.setProperty(p1.getName(), p2.getValues().get(0).copy());
+      }
+    }
+    return merged;
+  }
+
+  private DataType checkFixedValues(String vurl, String path, DataType v1, DataType v2) {
```
dotasek, line 684: This appears to be unused and could be deleted.

org.hl7.fhir.validation/src/test/resources/txCache/org.hl7.fhir.validation/4.3.0/servers.ini
dotasek: Is it possible to do recreate the tx caches before this goes in? It's not critical, as the presence of the old server is explained, but it would be nice.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/ManagedWebAccess.java
dotasek: I can see the thread, but I don't see any justification for it. A little digging, and it appears as though this is in https://www.rfc-editor.org/rfc/rfc2606.html. So... OK.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/BaseValidator.java
dotasek: I think docs are likely out of the scope for this PR. I'll resolve, but I still think this needs enumeration somewhere.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
dotasek: Maybe mention the package name in the error message, or a comment?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/SnapshotGenerationPreProcessor.java
dotasek: I can see the coverage when SnapshotGenerationTests are run, at 73%

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/DefinitionNavigator.java
dotasek: I can see the coverage when SnapshotGenerationTests are run, at 75%

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java
```diff
@@ -549,37 +578,53 @@ else if (myEnableWhenEvaluator.isQuestionEnabled(hostContext, qItem, qstack, fpe
...
     for (Element answer : answers) {
       NodeStack ns = stack.push(answer, i, null, null);
       if (qItem.getType() != null) {
         switch (qItem.getType()) {
           case GROUP:
-            ok.see(rule(errors, NO_RULE_DATE, IssueType.STRUCTURE, answer.line(), answer.col(), stack.getLiteralPath(), false, I18nConstants.QUESTIONNAIRE_QR_ITEM_GROUP));
+            ok.see(rule(errors, NO_RULE_DATE, IssueType.STRUCTURE, answer.line(), answer.col(), stack.getLiteralPath(), false, I18nConstants.QUESTIONNAIRE_QR_ITEM_GROUP, VersionUtilities.isR5Plus(context.getVersion()) ? "group" : "question"));
             break;
           case DISPLAY:  // nothing
+            ok.see(rule(errors, "2024-05-07", IssueType.STRUCTURE, answer.line(), answer.col(), stack.getLiteralPath(), false, I18nConstants.QUESTIONNAIRE_QR_ITEM_DISPLAY_NO_VALUE, VersionUtilities.isR5Plus(context.getVersion()) ? "group" : "question"));
```
dotasek, line 600: It looks like "Items of type ''display'' cannot have answers" is the message here, which doesn't need `VersionUtilities.isR5Plus(context.getVersion()) ? "group" : "question")`

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java
```diff
@@ -758,6 +808,536 @@ private void validateQuestionnaireResponseItemQuantity(List<ValidationMessage> e
...
+  private boolean validateQuestionnaireResponseItemDate(List<ValidationMessage> errors, Element answer, NodeStack ns, QuestionnaireItemComponent qItem) {
+    boolean ok = true;
+    Element v = answer.getNamedChild("value");
+    if (v != null) {
+      NodeStack vns = ns.push(v, -1, null, null);
+      try {
+        DateTimeType vdt = new DateTimeType(v.primitiveValue()); 
+        if (qItem.hasExtension(ToolingExtensions.EXT_MINVALUE)) {
+          DateTimeType dt = new DateTimeType(qItem.getExtensionByUrl(ToolingExtensions.EXT_MINVALUE).getValue().primitiveValue());
+          ok = rule(errors, "2024-05-07", IssueType.INVARIANT, vns, !dt.after(vdt), I18nConstants.QUESTIONNAIRE_QR_ITEM_DATE_MIN, v.primitiveValue(), dt.primitiveValue()) && ok;
```
dotasek, line 858: Would this not be the same as `vdt.after(dt)`?

It would also be self documenting if it said `dateTime.after(minDateTime)`.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java
```diff
@@ -758,6 +808,536 @@ private void validateQuestionnaireResponseItemQuantity(List<ValidationMessage> e
...
+        }
+        if (qItem.hasExtension(ToolingExtensions.EXT_MAXVALUE)) {
+          DateTimeType dt = new DateTimeType(qItem.getExtensionByUrl(ToolingExtensions.EXT_MAXVALUE).getValue().primitiveValue());
+          ok = rule(errors, "2024-05-07", IssueType.INVARIANT, vns, !dt.before(vdt), I18nConstants.QUESTIONNAIRE_QR_ITEM_DATE_MAX, v.primitiveValue(), dt.primitiveValue()) && ok;
```
dotasek, line 862: Same as previous comment re order and variable names.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java
```diff
@@ -758,6 +808,536 @@ private void validateQuestionnaireResponseItemQuantity(List<ValidationMessage> e
...
+            if (item.primitiveValue().equals(v.primitiveValue())) {
+              found = true;
+              break;
+            }
+          }
+          if (!found) {
+            ok = rule(errors, "2024-05-07", IssueType.INVARIANT, vns, found, I18nConstants.QUESTIONNAIRE_QR_ITEM_NODATE, v.primitiveValue()) && ok;
+          }
+        }
+      } else {
```
dotasek, line 890: Empty else here.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java
```diff
@@ -758,6 +808,536 @@ private void validateQuestionnaireResponseItemQuantity(List<ValidationMessage> e
...
+
+  private boolean validateQuestionnaireResponseItemTime(List<ValidationMessage> errors, Element answer, NodeStack ns, QuestionnaireItemComponent qItem) {
+    boolean ok = true;
+    Element v = answer.getNamedChild("value");
+    if (v != null) {
+      NodeStack vns = ns.push(v, -1, null, null);
+      try {
+        TimeType vdt = new TimeType(v.primitiveValue()); 
+        if (qItem.hasExtension(ToolingExtensions.EXT_MINVALUE)) {
+          TimeType dt = new TimeType(qItem.getExtensionByUrl(ToolingExtensions.EXT_MINVALUE).getValue().primitiveValue());
```
dotasek, line 904: Variable would be clearer if it was `minTime`

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java
```diff
@@ -758,6 +808,536 @@ private void validateQuestionnaireResponseItemQuantity(List<ValidationMessage> e
...
+          ok = rule(errors, "2024-05-07", IssueType.INVARIANT, vns, vdt.primitiveValue().compareTo(dt.primitiveValue()) >= 0, I18nConstants.QUESTIONNAIRE_QR_ITEM_TIME_MIN, v.primitiveValue(), dt.primitiveValue()) && ok;
+        }
+        if (qItem.hasExtension(ToolingExtensions.EXT_MAXVALUE)) {
+          TimeType dt = new TimeType(qItem.getExtensionByUrl(ToolingExtensions.EXT_MAXVALUE).getValue().primitiveValue());
```
dotasek, line 908: Variable would be cleaner if it was `maxTime`

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java
```diff
@@ -758,6 +808,536 @@ private void validateQuestionnaireResponseItemQuantity(List<ValidationMessage> e
...
+            if (item.primitiveValue().equals(v.primitiveValue())) {
+              found = true;
+              break;
+            }
+          }
+          if (!found) {
+            ok = rule(errors, "2024-05-07", IssueType.INVARIANT, vns, found, I18nConstants.QUESTIONNAIRE_QR_ITEM_NOTIME, v.primitiveValue()) && ok;
+          }
+        }
+      } else {
```
dotasek, line 937: Empty else.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java
```diff
@@ -758,6 +808,536 @@ private void validateQuestionnaireResponseItemQuantity(List<ValidationMessage> e
...
+        }
+        UcumService ucum = context.getUcumService();
+        
+        if (qItem.hasExtension(ToolingExtensions.EXT_MIN_QUANTITY)) {
+          Quantity dt = qItem.getExtensionByUrl(ToolingExtensions.EXT_MIN_QUANTITY).getValueQuantity();
+          if (!dt.hasSystem() || !dt.hasCode() || !vdt.hasSystem() || !vdt.hasCode()) {
+            ok = rule(errors, "2024-05-07", IssueType.INVARIANT, vns,  false, I18nConstants.QUESTIONNAIRE_QR_ITEM_DECIMAL_CANNOT_COMPARE_MIN_UNITS, gen(dt), gen(vdt)) && ok;            
+          } else if (dt.getSystem().equals(vdt.getSystem()) &&  dt.getCode().equals(vdt.getCode())) {
+            ok = rule(errors, "2024-05-07", IssueType.INVARIANT, vns, dt.getValue().compareTo(vdt.getValue()) >= 0, I18nConstants.QUESTIONNAIRE_QR_ITEM_QUANTITY_MIN, gen(vdt), gen(dt)) && ok;
+          } else if ("http://unitsofmeasure.org".equals(dt.getSystem()) && "http://unitsofmeasure.org".equals(vdt.getSystem())) {
```
dotasek, line 1129: "http://unitsofmeasure.org" should be in a constant.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java
```diff
@@ -758,6 +808,536 @@ private void validateQuestionnaireResponseItemQuantity(List<ValidationMessage> e
...
+            ok = rule(errors, "2024-05-07", IssueType.INVARIANT, vns, found, I18nConstants.QUESTIONNAIRE_QR_ITEM_NODECIMAL, v.primitiveValue()) && ok;
+          }
+        }
+      } else {
+      }
+    }
+    return ok;
+  }
+
+  private String gen(Coding c) {
```
dotasek, line 1221: What does `gen` mean? Could this be a clearer name?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java
```diff
@@ -758,6 +808,536 @@ private void validateQuestionnaireResponseItemQuantity(List<ValidationMessage> e
...
+    return (q.hasComparator() ? q.getComparator().toCode() : "") + q.getValueElement().asStringValue()+(q.hasUnit() ? " "+ q.getUnit() : "")+(q.hasSystem() || q.hasCode() ? " ("+getSystemName(q.getSystem())+"#"+q.getCode()+")" : "");
+  }
+
+  public String getSystemName(String system) {
+    if (Utilities.noString(system)) {
+      return "";
+    }
+    switch (system) {
+    case "http://unitsofmeasure.org": return "UCUM";
+    case "http://snomed.info/sct": return "SNOMED";
```
dotasek, line 1236: Should be a constant.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java
```diff
@@ -758,6 +808,536 @@ private void validateQuestionnaireResponseItemQuantity(List<ValidationMessage> e
...
+    }
+  }
+
+  private boolean validateQuestionnaireResponseItemReference(List<ValidationMessage> errors, Element answer, NodeStack ns, QuestionnaireItemComponent qItem) {
+    boolean ok = true;
+    Element v = answer.getNamedChild("value");
+    if (v != null && v.fhirType().equals("Reference")) {
+      NodeStack vns = ns.push(v, -1, null, null);
+      String ref = v.getNamedChildValue("reference");
+      if (ref != null && settings.isAssumeValidRestReferences()) {
```
dotasek, line 1310: Why do we use isAssumeValidRestReferences here? This looks like it's checking against a list of allowed reference types. Shouldn't it check those regardless of whether or not we're assuming valid references?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java
dotasek, line 1: All the `dt`, `vdt` references throughout could be clearer `min` and `max` names. 

There's a lot of repeated code, which would likely read nicer with a refactor into well named methods, but that can come later if this is passing Questionnaire validation tests now.

org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/tests/ValidationTests.java
```diff
@@ -524,11 +531,37 @@ public void test() throws Exception {
...
+      checkOutcomes(errorsLogical, logical, "logical", "logical", name, suppress);
     }
     logger.verifyHasNoRequests();
   }
 
+  private void moveOutcomes(JsonObject focus, String name, String mode) throws IOException {
+    if (REVISING_TEST_CASES) {
+      if (focus.has("dotnet-brianpos") && focus.get("dotnet-brianpos").isJsonObject()) {
+        byte[] json = JsonTrackingParser.writeBytes(focus.getAsJsonObject("dotnet-brianpos").getAsJsonObject("outcome"), true);
+        String filename = Utilities.path("/Users/grahamegrieve/work/test-cases/validator/outcomes/brianpos", name.replace("/", "-")+"-"+mode+".json");
```
dotasek, line 543: Should this have been committed to the repo?

dotasek (changes requested):

org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/tests/ValidationTests.java
grahamegrieve: it doesn't belong long term, but there's a collection of tests I need to make decisions about and then re-run that code. and I didn't want to lose it and have to rewrite it

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java
grahamegrieve: it looks repeated, but there's subtle variations to the theme all through them

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java
grahamegrieve: see https://confluence.hl7.org/spaces/FHIR/pages/35718580/Using+the+FHIR+Validator#UsingtheFHIRValidator-assumeValidRestReferences

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java
grahamegrieve: ok

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java
grahamegrieve: I use gen() a lot to mean generate a string representation. But sure, it could be clearer

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/SnapshotGenerationPreProcessor.java
grahamegrieve: ok

PR #1992, opened 2025-05-07
Improve tolerance of NPM parser

jamesagnew:
A discussion on the HAPI FHIR mailing list revealed that someone was trying to import an NPM package where all of the files in the TGZ file had `./` prefixed to the filenames. So e.g. the tar directory listed the package index as `./package/package.json` instead of `package/package.json`.

This PR makes the parser a bit more tolerant of this, and addresses a few other warnings and suggestions from IntelliJ.


grahamegrieve:
I'm not sure. you're right that it's postel's law, and generally I apply that. But also I worry because lots of people interpret this code as the definition of correct - if the validator is ok with it, it must be ok. 

and this is not ok. But now I've chased down the source, and I'll find out what's going on with that. So I think I'd be happy if the loader emits an error message on the console when it sees that (once per package) but loads it correctly

jamesagnew:
Makes sense - I've added code to emit a one-time warning when we encounter this.

dotasek:
@grahamegrieve, James' change appears to address your comment. I would approve on re-review.

grahamegrieve (approved):
ok

PR #1993, opened 2025-05-08
Structural replace of this != null conditions

dotasek:
Using IntelliJ Structural search and replace, this replaces the always true `this != null` that was autogenerated for hasXYZ methods.

IntelliJ search expression
```
this != null && this.$Expr$ instanceof $Type$
```

IntelliJ replace expression
```
this.$Expr$ != null && this.$Expr$ instanceof $Type$
```

Example match:

```java
return this != null && this.timing instanceof Period;
```

Example replacement:

```java
return this.timing != null && this.timing instanceof Timing;
```


org.hl7.fhir.dstu3/src/main/java/org/hl7/fhir/dstu3/model/ActivityDefinition.java
```diff
@@ -4190,8 +4190,8 @@ public Timing getTimingTiming() throws FHIRException {
       return (Timing) this.timing;
     }
 
-    public boolean hasTimingTiming() { 
-      return this != null && this.timing instanceof Timing;
+    public boolean hasTimingTiming() {
+      return this.timing != null && this.timing instanceof Timing;
```
jamesagnew: Saying `this.timing != null` is redundant... An `instanceof` check will always return false if the left hand side is null.

grahamegrieve (approved):
ok

PR #1994, opened 2025-05-08
Make cliContext fields deserialize in ValidationRequest

dotasek:
After a refactor of our modules, CliContext was replaced with the more appropriate ValidationContext.

This breaks deserialization of requests that still include CliContext.

This PR creates a backwards compatible setter and annotations that allow backwards compatibility, while adding deprecation to the CliContext method. 

Thanks, @dehall for discovering the problem, and providing the original fix.


grahamegrieve (approved):
ok

PR #1998, opened 2025-05-10
feat: allow .json5 file extension for advisor files in Params

patrick-werner:
This pull request introduces a minor enhancement to the file validation logic in `Params.java` by expanding the list of acceptable file extensions for advisor files.

Validation logic update:

* [`org.hl7.fhir.validation.cli/src/main/java/org/hl7/fhir/validation/cli/param/Params.java`](diffhunk://#diff-c680b8a0c94ab2d061f47151b3a17bedd24bac6116dc1e1f91f624306b8986a4L466-R466): Added support for `json5` file extension in the advisor file validation check, allowing `.json5` files to be used alongside `.json` and `.txt` files.


dotasek (changes requested):
I think this will cause trouble when the code reaches the point at which the advisor file is used:

https://github.com/hapifhir/org.hl7.fhir.core/blob/2a85363318590327f58ca8f3d25a25ceb7545406/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/service/ValidationService.java#L652

There is only json parsing available, which has been implemented in the core lib in order to support line+col awareness for validation.

Issue #2001, opened 2025-05-12
Comparer incorrectly throws “Found slices with no slicing details at Resource.extension” for extension slices


patrick-werner:
**Describe the bug**  
Comparing two differential profiles that define slices on `Resource.extension` fails with:

```
Found slices with no slicing details at Encounter.extension
```

---

**To Reproduce**  
1. Run the comparer:

   ```bash
   java -jar StructureDefinitionComparer.jar \
     --ig de.gematik.isik#5.0.0-rc \
     --ig hl7.fhir.us.core#8.0.0 \
     --left https://gematik.de/fhir/isik/StructureDefinition/ISiKKontaktGesundheitseinrichtung \
     --right http://hl7.org/fhir/us/core/StructureDefinition/us-core-encounter
   ```

2. Observe the exception:

See in the output: `Found slices with no slicing details at Encounter.extension`

---

**Expected behavior**  
- `*.extension` and `*.modifierExtension` elements should be treated as implicitly sliced by `url` (per FHIR core rules), so no “missing slicing” exception should be thrown.  

---

**Actual behavior**  
- The comparer throws a `DefinitionException` when encountering an extension slice without an explicit `<slicing>` block.  
- the output shows: `Found slices with no slicing details at Encounter.extension`

PR #2002, opened 2025-05-12
Fix Implicit-Extension Slicing Regression

patrick-werner:
This pull request fixes a regression around **implicit slicing of extensions** in the `DefinitionNavigator`. Previously, any slice on an `*.extension` or `*.modifierExtension` element without an explicit `<slicing>` block would trigger a `DefinitionException`. With these changes, extension elements are now correctly treated as implicitly sliced by URL, preventing spurious errors.

### Edge case handling:
* Enhanced the `loadChildren()` method to skip missing slicing errors for extensions (e.g., `.extension`, `.modifierExtension`), as they are implicitly sliced by URL. This prevents unnecessary exceptions for valid cases. (`[org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/DefinitionNavigator.javaR190-R198](diffhunk://#diff-402b37d6b126ce2277081d8f5f57e16a8e19bda304227bb1a0f75b8a20598d5eR190-R198)`)

### Null safety improvements:
* Updated the `slices()` method to return an empty list instead of `null` when no slices are found, using `Collections.emptyList()` for better null safety. (`[org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/DefinitionNavigator.javaL159-R161](diffhunk://#diff-402b37d6b126ce2277081d8f5f57e16a8e19bda304227bb1a0f75b8a20598d5eL159-R161)`)

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/DefinitionNavigator.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/DefinitionNavigator.java
@@ -32,6 +32,7 @@ package org.hl7.fhir.r5.utils;
 
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


patrick-werner:
fixes #2001

grahamegrieve:
Since I already fixed this, we aren't accepting a change like this without a test case showing what it fixes

patrick-werner:
> Since I already fixed this, we aren't accepting a change like this without a test case showing what it fixes

@grahamegrieve i created a fhir-test-case to show what the PR fixes:
https://github.com/FHIR/fhir-test-cases/pull/225

PR #2006, opened 2025-05-13
Fix terminology cache hashing

dotasek:
It's possible for different newline encodings to be introduced to the caches stored in resources. This means that entries can have duplicates produced by other operating systems (one for `\r\n\` and one for `\n`) or will miss if only one exists.

This PR normalizes newlines to `\n` when generating hashes for the Terminology Cache.


grahamegrieve (approved):
ok

PR #2009, opened 2025-05-14
Move to Slf4J

dotasek:
**IMPORTANT: This is intended to be merged after release 6.5.22. Please do not merge it earlier.**

This PR replaces our System.out calls with appropriate Slf4J logging.

A few notes for reviewers:

* There is a new Checkstyle rule which prohibits System.out or System.err in java files. There are exceptions for this, where `@SuppressWarnings("checkstyle:systemout")` has been used: 
  * Classes like ResourcePercentageLogger, which attempt to account for cases where print(String) statements are used to construct a single expanding line. Since Slf4j and logging implementations only handle a single message at a time, these classes output both print and println output in the console directly, but DEBUG levels in the log appenders. 
  * Classes that are only intended to be run as command-line utilities (ValidatorCLI excepted). These classes are evidently utilities for building the standard and are not referenced in other files in the codebase.
  * In numerous locations, there are internal 'debug' switches. Where appropriate, these were simply replaced with `log.debug(String)` calls, and the relevant `get/set Debug` methods were deprecated. Logging implementations, such as `logback-text.xml` should be used to set message levels, NOT changes in actual java code.


grahamegrieve (approved):
looks good

PR #2013, opened 2025-05-18
Convert DocumentReference.date/indexed/created conform fhir specs

Corne:
Following the FHIR docs for the conversion of DocumentReference between r3 and r4, `date` should be mapped from/to `indexed`.
At the moment `date` is mapped to `created` and `indexed` isn't mapped at all.

https://hl7.org/fhir/R4/documentreference-version-maps.html

### R3 to R4
```
  src.indexed -> tgt.date;
  src.created as vs ->  tgt.extension as ext,  ext.url = 'http://hl7.org/fhir/3.0/StructureDefinition/extension-DocumentReference.created',  ext.value = vs;
```

### R4 to R3
```
  src.date -> tgt.indexed;
  src.extension as ext where url = 'http://hl7.org/fhir/3.0/StructureDefinition/extension-DocumentReference.created' then {
    ext.value as vs0 -> tgt.created = vs0 "created2";
  } "created";
```

Also `indexed` is a required field in r3, so with the current mapping there is never a valid r3 model created
afbeelding
https://hl7.org/fhir/STU3/documentreference.html

### DIFF
--- a/org.hl7.fhir.convertors/src/test/java/org/hl7/fhir/convertors/conv30_40/DocumentReference30_40Test.java
+++ b/org.hl7.fhir.convertors/src/test/java/org/hl7/fhir/convertors/conv30_40/DocumentReference30_40Test.java
@@ -1,9 +1,16 @@
 package org.hl7.fhir.convertors.conv30_40;
 
+import ca.uhn.fhir.context.FhirContext;
+import org.hl7.fhir.convertors.advisors.impl.BaseAdvisor_30_40;
 import org.hl7.fhir.convertors.factory.VersionConvertorFactory_30_40;
 
 
+import org.hl7.fhir.r4.elementmodel.Manager;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.io.InputStream;
 
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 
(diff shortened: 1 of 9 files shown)


org.hl7.fhir.convertors/src/test/java/org/hl7/fhir/convertors/conv30_40/DocumentReference30_40Test.java
dotasek, line 1: Except where noted in the test resources, this test looks logically correct.

However, the variable names in the test are inaccurate; some say r4 resources when they're clearly from dstu3, and the difference between inputstream and Java resource model is unclear. I realize this is likely from copying some other test, but this is a good opportunity to fix these.

org.hl7.fhir.convertors/src/test/resources/document_reference_40_converted_to_30.json
```diff
@@ -0,0 +1,94 @@
...
+  "meta": {
+    "lastUpdated": "2013-07-01T13:11:33Z"
+  },
+  "text": {
+    "status": "generated",
+    "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\">\n\n\t\t\t\t\t\t<a href=\"http://localhost:9556/svc/fhir/Binary/1e404af3-077f-4bee-b7a6-a9be97e1ce32\">Document: urn:oid:129.6.58.92.88336</a>undefined, created 24/12/2005\n\t\t\t\t\t</div>"
+  },
+  "extension": [
+    {
+      "url": "http://hl7.org/fhir/3.0/StructureDefinition/extension-DocumentReference.created",
```
dotasek: Is it a common conversion practice to maintain the extension, even if the dstu3 resource now has the correct `created` element?

Corne: Personally I would expect not. But I checked how other models handled this, so for example the mapping of `category` on `CarePlanActivityDetailComponent` with the `EXT_CarePlanActivityDetailComponentExtension`. Where it seems like the extension is kept between conversions (unless I am missing something). 

https://github.com/hapifhir/org.hl7.fhir.core/blob/master/org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/conv30_40/resources30_40/CarePlan30_40.java 

However now that I looked into it again, it seems like in other cases these extensions are excluded from the target. So it seems this isn't consistent at the moment. For this PR will change it so the extension will not be mapped to dstu3 model. Might create a new issue to make this consistent behaviour for all similar extension conversions.


Few examples, where it is excluded:
```
org.hl7.fhir.r5.utils.ToolingExtensions.removeExtension(tgt, VersionConvertorConstants.EXT_PAYLOAD_CONTENT);
```
https://github.com/hapifhir/org.hl7.fhir.core/blob/master/org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/conv40_50/resources40_50/CommunicationRequest40_50.java#L286

```
ConversionContext10_40.INSTANCE.getVersionConvertor_10_40().copyDomainResource(src, tgt, IGNORED_EXTENSION_URLS);
```
https://github.com/hapifhir/org.hl7.fhir.core/blob/master/org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/conv10_40/resources10_40/Conformance10_40.java#L27

org.hl7.fhir.convertors/src/test/java/org/hl7/fhir/convertors/conv30_40/DocumentReference30_40Test.java
Corne: Good catch, should be better now.
Not sure if you are asking me to update the naming in all the other tests aswell, I could, but would prefer to do it in a seperate PR.

dotasek:
Changes look good, thank you!

I created this issue regarding the behaviour of extensions: https://github.com/hapifhir/org.hl7.fhir.core/issues/2021

I wholeheartedly agree that cleaning up other tests and convertors are out of scope for this PR, and not your responsibility.

PR #2020, opened 2025-05-23
Use correct `cs` language for Czech.

### LINKED ISSUE #2014: Wrong mapping from country code to lang code for the Czech Republic
A proper language code for the Czech language is not cz-CZ but **cs-CZ**. 

I found this error in RegionToLocaleMapperTest.java

This mapping error most probably has impact on the Czech IG publication where FHIR IG Publisher generates many errors on use of a wrong language code. (I am not sure, but I didn't find any other possible cause as cz-CZ is not set in the IG itself, it seems to be generated from IG jurisdiction property).

dotasek:
Fixes #2014


grahamegrieve (approved):
ok

PR #2039, opened 2025-06-05
Update PECodeGenerator.java

jkiddo:


org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/profilemodel/gen/PECodeGenerator.java
```diff
@@ -499,7 +514,11 @@ private void genSave(boolean isPrim, boolean isAbstract, String name, String sna
             w(save, "      tgt.addChild(\""+fname+"\", item.toCode());");
           }  
         } else {
-          w(save, "      tgt.addChild(\""+fname+"\", item);");
+            if(typeInfo.getUrl().startsWith("http://hl7.org/fhir/StructureDefinition"))
+                w(save, "      tgt.addChild(\""+fname+"\", item);");
+            else
+                w(save, "      tgt.addChild(\""+fname+"\",(Type) item.getData());");
```
jkiddo, line 520: @grahamegrieve i'm gonna need some guidance here - because `getData` is deprecated. This line comes into play when using profiled datatypes.

grahamegrieve: I'm not sure why it's deprecated

PR #2044, opened 2025-06-09
Add Uzbek translations for languages

vadi2:
Add Uzbek translations for languages - verified with a native speaker.


grahamegrieve (approved):
ok

PR #2048, opened 2025-06-12
Only use ucum service when needed + refactor ucum equality check

dotasek:
The Questionnaire validator gets the ucum service even in cases where it isn't necessary.

This PR only gets it when actually validating ucum, and also reuses ucum validation code (less duplicated code).


grahamegrieve (approved):
ok

PR #2049, opened 2025-06-13
Support conversion of CommunicationRequest 30-40

Corne:
Implemented onversion for CommunicationRequest betweeen dstu3 en r4 model

Mapping is based on the conversion maps at:
https://hl7.org/fhir/R4/communicationrequest-version-maps.html


grahamegrieve (approved):
ok

PR #2050, opened 2025-06-13
2025 06 gg pinning support

grahamegrieve:


org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/elementmodel/SignatureTest.java
dotasek, line 1: It looks like createJWSWithNimbus isn't actually used anywhere else in the code, and there are no assertions in the test. Maybe this should be removed or restructured in some way?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/UserDataNames.java
```diff
@@ -155,5 +155,6 @@ public class UserDataNames {
   public static final String IG_DEP_ALIASED = "IG_DEP_ALIASED";
   public static final String CS_MARKDOWN_FLAG = "CS_MARKDOWN_FLAG";
   public static final String PROFILE_RENDERED_MAPPINGS = "PROFILE_RENDERED_MAPPINGS";
+  public static final String EXP_REVIEWED = "EXP_REVIEWED";
```
dotasek, line 158: I see no usages of this in the code. Is this used in a downstream project somewhere, or can it be deleted?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/DataRenderer.java
```diff
@@ -790,7 +790,7 @@ public boolean canRenderDataType(String type) {
   }
 
   public boolean renderDataType(RenderingStatus status, XhtmlNode x, ResourceWrapper type) throws FHIRFormatError, DefinitionException, IOException {
-    return renderDataType(status, null, x, type);
+    return renderDataType(status, x, x, type);
```
dotasek, line 793: This looks a bit weird to me, as it looks like the logic adds features to both a parent node and a child, and in this case they'd be the same node.

Is there maybe a way to handle the previous null? Maybe even to use the child as a parent if that's the correct way to render.

dotasek (changes requested):
See comments.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/DataRenderer.java
grahamegrieve: it is weird, but when I looked at the context, it was technically correct. Kind of screwy thing about html and <td> that do or don't contain <p>. It probably will come back as a bigger issue somewhere, but we'll let it lie for now

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/UserDataNames.java
grahamegrieve: in the IG publisher which also uses these constants

PR #2055, opened 2025-06-17
fix line breaks in pt translations

costateixeira:
there was a line break missing when I added lots of markers for unfinished translations. This now addresses that to unblock contributors


dotasek (approved):
I found this earlier and was meaning to bring it up. Thank you!

PR #2056, opened 2025-06-18
Update some PT_BR translations, mark AI generated translations

italomacedo:


dotasek:
@italomacedo there is some odd formatting happening which doesn't look correct. There is meant to be a line break separating `fuzzy` and `msgid...`

```
, fuzzymsgid "Max Binding"
```

italomacedo:
After a very thorough talk with @costateixeira , we're moving away for now from branching pt in pt_BR pt_PT and so on. We can get back to it later if cultural differences demand to do it otherwise.

italomacedo:
Closing it for now so We can understand if branching pt into pt_BR and pt_PT is really necessary.

PR #2057, opened 2025-06-18
Fixed error with OperationDefinition Parameter cross-validation:

lmckenzi:
- Can't have target profiles if the type is Element due to existing OperationDefinition invariant
- Old code was sticking profiles into targetProfiles, which generated spurious errors

### DIFF
--- a/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/OperationDefinitionValidator.java
+++ b/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/OperationDefinitionValidator.java
@@ -135,9 +135,6 @@ public class OperationDefinitionValidator extends BaseValidator {
       for (TypeRefComponent tr : resDefn.current().getType()) {
         String t = tr.getWorkingCode();
         profileTypes.add(t);
-        for (CanonicalType ct : tr.getProfile()) {
-          profileTargets.add(ct.asStringValue());
-        }
         ok = rule(errors, "2025-04-08", IssueType.INVALID, nsp, allowedTypes.contains(t), I18nConstants.OPDEF_PROFILE_TYPE_NOT_IN_PARAMS, use, paramName, t, CommaSeparatedStringBuilder.join2(",", " and ", allowedTypes)) && ok;
       }
     }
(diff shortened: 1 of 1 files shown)


org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/OperationDefinitionValidator.java
```diff
@@ -135,9 +135,6 @@ private boolean compareParameterDefinitions(List<ValidationMessage> errors, Node
       for (TypeRefComponent tr : resDefn.current().getType()) {
         String t = tr.getWorkingCode();
         profileTypes.add(t);
-        for (CanonicalType ct : tr.getProfile()) {
```
dotasek, line 138: Should this not be the following?

```java
 for (CanonicalType ct : tr.getTargetProfile()) {
     profileTargets.add(ct.asStringValue());
 }
 ```

File: org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/OperationDefinitionValidator.java [hunk 2]
```diff
@@ -149,8 +146,14 @@ private boolean compareParameterDefinitions(List<ValidationMessage> errors, Node
       ok = rule(errors, "2025-04-08", IssueType.INVALID, nsp, profileTargets.contains(t), I18nConstants.OPDEF_PROFILE_PROFILE_NOT_IN_PARAMS, use, paramName, t, CommaSeparatedStringBuilder.join2(",", " and ", profileTargets)) && ok;        
     }
     
-    for (String t : profileTargets) {
-      ok = rule(errors, "2025-04-08", IssueType.INVALID, nsp, allowedTargets.contains(t), I18nConstants.OPDEF_PROFILE_PROFILE_NOT_IN_PROFILE, use, paramName, t, CommaSeparatedStringBuilder.join2(",", " and ", allowedTargets)) && ok;        
+    String paramType = param.getNamedChildValue("type");
+    if (paramType!=null && !paramType.equals("Element")) {
+      // If the type is Element, then it's a polymorphic type and target profiles can't be declared in the operation because doing so
```
dotasek, line 151: Is there a test case that would illustrate this, or at least some resource that would cause the spurious results noted?

dotasek (changes requested):
Needs a merge from master; the rule(...) call seems to have changed there.

dotasek (commented):
See comments.

PR #2059, opened 2025-06-23
2025 06 gg digital signatures

grahamegrieve:


grahamegrieve:
@dotasek

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
```diff
@@ -354,17 +354,21 @@ public int compare(T arg1, T arg2) {
...
   private IWorkerContextManager.ICanonicalResourceLocator locator;
   protected String userAgent;
+  protected ContextUtilities cutils;
+  private List<String> suppressedMappings;
 
   protected BaseWorkerContext() throws FileNotFoundException, IOException, FHIRException {
     setValidationMessageLanguage(getLocale());
     clock = new TimeTracker();
     initLang();
+    cutils = new ContextUtilities(this, suppressedMappings);
```
dotasek, line 364: cutils gets instantiated here, but later, suppressedMappings can be explicitly set without re-initing cutils. That could lead to odd cases where cutils is operating with a different list of suppressedMappings.

Also, suppressedMappings is never explicitly set via setSuppressedMappings in this project. Maybe downstream?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/JsonParser.java
```diff
@@ -852,6 +852,9 @@ public void compose(Element e, JsonCreator json) throws Exception {
   }
 
   private void compose(String path, Element e, Set<String> done, Element child) throws IOException {
+    if (canonicalFilter.contains(child.getPath())) {
```
dotasek, line 856: What happens to Bundle.id, Bundle.meta, etc. if this returns early? Where would it be persisted in json?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/XmlParser.java
```diff
@@ -809,6 +815,9 @@ public void compose(Element e, IXMLWriter xml) throws Exception {
   }
 
   private void composeElement(IXMLWriter xml, Element element, String elementName, boolean root) throws IOException, FHIRException {
+    if (canonicalFilter.contains(element.getPath())) {
```
dotasek, line 818: Same question as JsonParser: what happens to canonical paths?

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/formats/CanonicalJsonTest.java
```diff
@@ -0,0 +1,127 @@
...
+    for (JsonElement p : array) {
+      writeValue(json, p);
+    }
+    json.endArray();
+  }
+  
+
+  @Test
+  public void testCanonicalJsonNumbers() throws IOException {
+      String[] testCases = {
```
dotasek, line 90: This should just use either a parameterized test using a method, or csv input. It would be a pain to jump back and forth between these arrays to debug.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/Base64.java
```diff
@@ -25,6 +25,15 @@ public static boolean isBase64(byte[] arrayOctet) {
 
     return true;
   }
+  public static boolean isBase64(String content) {
+    for (char ch : content.toCharArray()) {
+      if (!isBase64((byte) ch) && !isWhiteSpace((byte) ch)) {
```
dotasek: !isBase64((byte) ch) && !isWhiteSpace((byte) ch) should probably go into a single method so this and the byte[] version of the method are always using the same code.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/JsonParser.java
grahamegrieve: It won't be persisted at all. that's the point

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/XmlParser.java
grahamegrieve: same answer

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/formats/CanonicalJsonTest.java
grahamegrieve: that's true, and if I ever have to look at these again, I'll do that. or you can

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
grahamegrieve: discussion moved to Zulip

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/Base64.java
grahamegrieve: fixed

PR #2061, opened 2025-06-23
Deprecation removal 1

dotasek:
This is the first in a series of code deletions that target deprecated methods and classes. 

This should remove all deprecated and not-in-use classes and methods that have been tagged as deprecated for over 1 year.

There is also a housekeeping deletion of classes intended to be run as main(String[] args) classes that other deprecated classes may reference: 9c27c58858a0ceb68fb6ec9e3d063e3d285fed3a


grahamegrieve (approved):
ok

PR #2071, opened 2025-07-01
fix error in tx error processing

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/DataRenderer.java
```diff
@@ -1482,7 +1471,7 @@ protected void renderCodeableConcept(RenderingStatus status, XhtmlNode x, Resour
         } 
       } 
 
-      x.span(null, context.formatPhrase(RenderingContext.DATA_REND_CODES) +b.toString()).addText(s);
```
dotasek, line 1485: As I understand this, addTextWithLineBreaks will add text with <br> inserted instead of `\r\n`

It works because addTextWithLineBreaks calls tx(String), which loops back to the original addText(String) and does the additional logic there.

Do we want to account for linux style line breaks as well? `\n`?

org.hl7.fhir.r5/src/test/resources/txCache/org.hl7.fhir.r5/cs-externals.json
```diff
@@ -1,5 +1,6 @@
 {
   "http://hl7.org/fhir/tools/CodeSystem/obligation|null" : null,
   "http://snomed.info/sct|null" : null,
+  "http://acme.org/messagetypes|null" : null,
```
dotasek, line 4: Out of scope for this PR: do we want to be cacheing acme.org and example.org results? Could that go sideways if those URLs are cached and later validations don't allow example.org/acme.org?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ValidationEngine.java
```diff
@@ -716,12 +719,25 @@ public OperationOutcome validate(String location, ByteProvider source, FhirForma
     }
     InstanceValidator validator = getValidator(cntType);
     Element res = validator.validate(null, messages, new ByteArrayInputStream(source.getBytes()), cntType, asSdList(profiles));
+    boolean first = true;
     for (String fn : matchetypes) {
+      if (first) {
+        messages.clear();
```
dotasek: Why are we deleting all the errors we already evaluated here?

Does any content in `matchetypes` override the results of the validation we just did? 

If so, we should make the code branch earlier instead of all that logic that gets run and then thrown out at this point.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ValidatorUtils.java
```diff
@@ -132,19 +132,19 @@ protected static Document parseXml(byte[] cnt) throws ParserConfigurationExcepti
     return builder.parse(new ByteArrayInputStream(cnt));
   }
 
-  protected static List<ValidationMessage> filterMessages(List<ValidationMessage> messages) {
+  protected static List<ValidationMessage> removeDuplicates(List<ValidationMessage> messages) {
     List<ValidationMessage> filteredValidation = new ArrayList<ValidationMessage>();
```
dotasek: I like this renaming. It would make it even clearer if `filteredValidation` was `uniqueMessages`, or something similar.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/BundleValidator.java
```diff
@@ -1183,7 +1224,9 @@ private boolean validateSignatureJose(List<ValidationMessage> errors, Element bu
         try {
           toSign = makeSignableBundle(bundle, canon, xml);
         } catch (Exception e) {
-          // nothing - this won't happen
+          if (settings.isDebug()) {
```
dotasek, line 1227: Better to use log.debug(e.getMessage, e) here and wherever debugging just involves leaving a log.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/BundleValidator.java
```diff
@@ -1393,8 +1473,41 @@ private boolean validateSignatureDigSig(List<ValidationMessage> errors, Element
...
+      for (Element type : types) {
+        String p = null;
+        String system = type.getNamedChildValue("system");
+        String code = type.getNamedChildValue("code"); 
+        if (OIDUtilities.isValidOID(code)) {
+          p = "urn:oid:"+code;
+        } else {
+          p = system+"#"+code;
+        }
+        if (purpose == null) {
```
dotasek, line 1492: This seems a bit weird to me. How does the spec work here? If the sig doesn't have a purpose, it feels like we should have branched earlier to handle that case.

dotasek (changes requested):
See comments.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/DataRenderer.java
grahamegrieve: I don't understand about the loopback thing. but good pick about the linux thing - I thought I tested that. will fix

org.hl7.fhir.r5/src/test/resources/txCache/org.hl7.fhir.r5/cs-externals.json
grahamegrieve: I think it's tested before it gets to caching, so it's ok

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ValidationEngine.java
grahamegrieve: we don't the underlying validation output, just the matchetype evaluation. Unless we decide we want both - I was wondering about making it a parameter? So I decided not to invest in pulling out of the full validation (yet?) and just delete all the validation output. But also, I shouldn't delete anything fatal - I'll fix that up

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ValidatorUtils.java
grahamegrieve: ok

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/BundleValidator.java
grahamegrieve: then we don't get a stack?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/BundleValidator.java
grahamegrieve: I don't think so. we'd still have to iterate the type element like we're doing, we'd just have to do that twice. So that would be better?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/BundleValidator.java
dotasek: You do get a stack if the exception is included in the arg:

```java
      Exception e = new Exception("meh", new Throwable("meh").fillInStackTrace());
      log.error(e.getMessage(), e);
```

The above produces this logging:

```log
meh
java.lang.Exception: meh
	at org.hl7.fhir.validation.cli.ValidatorCli.<init>(ValidatorCli.java:120)
	at org.hl7.fhir.validation.cli.ValidatorCli.main(ValidatorCli.java:254)
Caused by: java.lang.Throwable: meh
	... 2 common frames omitted
```

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/FhirValidatorHttpService.java
```diff
@@ -0,0 +1,343 @@
...
+    server.setExecutor(null); // Use default executor
+    server.start();
+
+    System.out.println("FHIR Validator HTTP Service started on port " + port);
+  }
+
+  /**
+   * Start the HTTP server and load the validator
+   */
+  public void start(String version, String txServer, String log) throws Exception {
```
dotasek, line 72: This seems to be only really used in testing code or the main(String[] args) method. I don't think this method should really exist.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/FhirValidatorHttpService.java
```diff
@@ -0,0 +1,343 @@
...
+
+  public FhirValidatorHttpService(int port) {
+    this.port = port;
+  }
+
+
+  /**
+   * Start the HTTP server on an existing loaded validator
+   */
+  public void bind(ValidationEngine validationEngine, IgLoader igLoader) throws IOException {
```
dotasek, line 47: This only ever gets called once, directly after instantiation. I would prefer if validationEngine and igLoader were final in this class, since they really shouldn't be changed after the server is started.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/FhirValidatorHttpService.java
```diff
@@ -0,0 +1,343 @@
...
+    exchange.sendResponseHeaders(statusCode, responseBytes.length);
+    try (OutputStream os = exchange.getResponseBody()) {
+      os.write(responseBytes);
+    }
+  }
+
+  /**
+   * Main method to start the service
+   */
+  public static void main(String[] args) {
```
dotasek, line 314: I think this should only be accessed through the ValidatorCli, where there's already logic to produce the ValidatorEngine from a well documented set of Params.

dotasek (changes requested):
The FHIRValidatorHttpService seems to function OK, but its instantiation and start methods look like they could lead to some confusion. I made some suggestions; I could even do a refactor myself.

PR #2075, opened 2025-07-02
Fix escaped quotes

dotasek:
Fixes complaints from the PO file editor about double escaped quotes.

PR #2076, opened 2025-07-03
Fix escaped quotes in PO files and properties

dotasek:
Replaces https://github.com/hapifhir/org.hl7.fhir.core/pull/2075

This PR refactors the POGenerator for legibility and fixes PO and properties files that contain invalid escapes for double quotes.

Based on the POGenerator code:

* PO files should always use an escape in the form of `\\\"`
* properties files should always use an escape in the form of `\"`

At this point, the properties and PO files should be consistent and fixed. 

It may be necessary to follow up with some enforcement of this in the POGenerator


dotasek:
@grahamegrieve some of the issues we're experiencing could be caused here:

https://github.com/hapifhir/org.hl7.fhir.core/blob/654095e4d53cacbcc29f3a2edfc58739be7ceb4d/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/i18n/POSource.java#L55

The msgid can contain single escapes (maybe someone's tooling is doing this?) and it may be messing with our expected format.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/i18n/POSource.java
```diff
@@ -151,57 +150,56 @@ private static String trimQuotes(String s) {
...
       } else {
-        b.append("# "+o.getComment()+"\r\n");
+        sb.append("# "+poObject.getComment()+"\r\n");
       }
-      b.append("#: "+o.getId()+"\r\n");
-      if (!o.getFlags().isEmpty()) {
-        b.append("#, "+ CommaSeparatedStringBuilder.join(",", o.getFlags()));
+      sb.append("#: "+poObject.getId()+"\r\n");
+      if (!poObject.getFlags().isEmpty()) {
+        sb.append("#, "+ CommaSeparatedStringBuilder.join(",", poObject.getFlags()) + "\r\n");
```
dotasek, line 171: @grahamegrieve here's where the POGenerator was missing a split between flags and the next line.

grahamegrieve (approved):
ok

PR #2077, opened 2025-07-03
Implement Converage.class/grouping conversion between dstu3 and r4

Corne:
Implemented conversion support for class and grouping elements of Coverage resource between DSTU3 and R4 versions.
Implementation is based on mapping description from https://hl7.org/fhir/R4/coverage-version-maps.html

Test resources are based on example files at https://hl7.org/fhir/R4/coverage-example.json.html

Note that the mapping definition does not define the mapping of the sequence element between DSTU3 and R4 versions.
I think this is an error in the mapping definition, so I still implemented it based on the conversion from R4 to DSTU3
```
    //note mapping of sequence from dstu3 to r4 is not described in mapping
    if (src.hasSequence()) {
      tgt.addClass_(convertCoverageGrouping("sequence", src.getSequence(), null));
    }
```


grahamegrieve (approved):
ok

PR #2083, opened 2025-07-10
Rework FhirValidatorHttpService

dotasek:
Simplifies FhirValidatorHttpService and ensures that ValidationEngine is only set once in the service's constructor.


grahamegrieve (approved):
ok

PR #2084, opened 2025-07-10
Expiring terminology capabilities in terminology cache

dotasek:
Since terminology server capabilities may change, the cached capabilities should expire after 24 hours.

This adds an interface for caching these capabilities and an implementation based on the apache commons expiring map, which is already available in our dependencies.


grahamegrieve (approved):
ok

PR #2091, opened 2025-07-13
2025 07 gg r3 clean up

grahamegrieve:


org.hl7.fhir.dstu3/src/main/java/org/hl7/fhir/dstu3/elementmodel/Property.java
```diff
@@ -1,399 +0,0 @@
-package org.hl7.fhir.dstu3.elementmodel;
```
dotasek, line 1: This class removal impacts a tested portion of HAPI:

https://github.com/hapifhir/hapi-fhir/compare/do-20250714-test-core-dstu3-deprecation#diff-35dea24f869a5d7e77aaeb657c985113898a0597e031461407c9050f2ff3d276

I do not however, see it used anywhere else in the code.

org.hl7.fhir.dstu3/src/main/java/org/hl7/fhir/dstu3/utils/StructureMapUtilities.java
```diff
@@ -1,2783 +0,0 @@
-package org.hl7.fhir.dstu3.utils;
```
dotasek, line 1: The removal of this class impacts a test in HAPI

https://github.com/hapifhir/hapi-fhir/compare/do-20250714-test-core-dstu3-deprecation#diff-7bda8917513f3cf1ac1954f81b66d9594cb047e6740ee24af617c73385bef1f9

However, this does not appear to be used in any other part of the code base.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/ResourcePercentageLogger.java
```diff
@@ -10,6 +10,8 @@ public class ResourcePercentageLogger {
   private final Logger logger;
   private final int total;
   private final String fhirType;
+
+  private static boolean loggingSuppressed;
```
dotasek, line 14: This repeats functionality already in the log field. I think it would be better to introduce a logValidationProgress field in ValidationEngine that is propagated to InstanceValidator and then this logging utility.

dotasek (changes requested):
See requested changes.

@grahamegrieve aside from the logging change I requested, we should see what @jamesagnew thinks of the impact of the deletions. I THINK they only impact some very old tests, but maybe he has some deeper insight.

org.hl7.fhir.dstu3/src/main/java/org/hl7/fhir/dstu3/elementmodel/Property.java
grahamegrieve: does that mean HAPI is still using this stuff? surely not?

org.hl7.fhir.dstu3/src/main/java/org/hl7/fhir/dstu3/utils/StructureMapUtilities.java
grahamegrieve: good get rid of that part too

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/ResourcePercentageLogger.java
grahamegrieve: ok sure, that'd be better. Do you want to do that? or me?

org.hl7.fhir.dstu3/src/main/java/org/hl7/fhir/dstu3/elementmodel/Property.java
jamesagnew: @dotasek We can definitely safely remove those.. They're just long outdated tests.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/ResourcePercentageLogger.java
dotasek: Done in https://github.com/hapifhir/org.hl7.fhir.core/pull/2095

PR #2093, opened 2025-07-15
FHIRPath fixes

brianpos:
* div operation doesn't handle quantity parameters - remove un-wanted testing
* Add makeQuantity/readAsQuantity to convert from the R5 element model to the object model Quantity
* Use the makeQuantity in the fhirpath engine when handling quantity parameters to work with both the object model and element model
* pre-apply coalesce implementation FHIR-43294
* fhirpath parsing position/length indicators updated in some edge cases (specifically whitespace handling after/in tokens)
*


grahamegrieve (approved):
ok

PR #2094, opened 2025-07-15
Improve validator search url detection

jamesagnew:
InstanceValidator has two separate code paths for validating search URLs within resources.  This PR collapses these into a single method. In addition:

* The regex has been modified to allow `.` and `:` within parameter names
* The regex is now pre-compiled for better evaluation performance
* Two calls to `String#split` have been replaced with `StringUtil.countMatches` for better performance


grahamegrieve (approved):
ok

PR #2095, opened 2025-07-15
Setter/getter for validation progress logging

dotasek:
This field allows a caller of ValidationEngine to enable or disable validation progress logging when the InstanceValidator is used.


grahamegrieve (approved):
ok

PR #2096, opened 2025-07-16
HTTPService: warn about ignored args + remove IG endpoint

dotasek:
This warns when watch-mode is used in server mode and removes the IG loading endpoint.

The IG endpoint was a source of mutation for the state of the ValidationEngine, which increased the possibility of misuse.


grahamegrieve (approved):
ok

PR #2098, opened 2025-07-18
2025 07 gg sweep

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java
```diff
@@ -207,11 +207,25 @@ public abstract class FunctionDefinition {
...
+     * @param name - name reference to resolve -
      * @param beforeContext - whether this is being called before the name is resolved locally, or not
      * @return the value of the reference (or null, if it's not valid, though can throw an exception if desired)
      */
     public List<Base> resolveConstant(FHIRPathEngine engine, Object appContext, String name, boolean beforeContext, boolean explicitConstant)  throws PathEngineException;
+
+
+    /**
+     * A constant reference - e.g. a reference to a name that must be resolved in context.
+     * The % will be removed from the constant name before this is invoked.
```
dotasek: This JavaDoc is unclear, here, and in the above method. Should this not be:

"The % must be removed from the constant name before it is used as the 'name' param." 

and then

"The % should remain the first character of the constant name when it is used as the 'name' param."

?

Also, see the following two calls of this function, where this behaviour is inconsistent:

Here, the call does not strip the leading char:

https://github.com/hapifhir/org.hl7.fhir.core/blob/43ac3d440258740eef867998bb5247c4cf574b72/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java#L1733

While here, it appears to, which, from the comment, is not the expected input.

https://github.com/hapifhir/org.hl7.fhir.core/blob/43ac3d440258740eef867998bb5247c4cf574b72/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java#L3370

dotasek: Also, is the inconsistency between the two methods ("%name" vs "name") truly a bug?

At present, all tests pass with the current code (except maybe that edge case from above, which I'm not sure is real).

But having the methods be consistent would be ideal, particularly if we're already doing breaking changes. If HAPI is our primary consumer, it would have to make a few fixes in its implementation.

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/misc/ProfileVersionAdaptor.java
```diff
@@ -234,6 +238,87 @@ public StructureDefinition convert(StructureDefinition sd, List<ConversionMessag
     return sd;
   }
 
+  private StructureDefinition convertLogical(StructureDefinition sdSrc, List<ConversionMessage> log) {
```
dotasek, line 236: I see no test coverage of this new method. How are we sure this works?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java
```diff
@@ -1200,7 +1202,7 @@ private void copyInheritedExtensions(StructureDefinition base, StructureDefiniti
   }
 
   private String getExtensionAction(String url) {
-    StructureDefinition sd = context.fetchResource(StructureDefinition.class, url);
+    StructureDefinition sd = context.fetchResourceRaw(StructureDefinition.class, url);
```
dotasek, line 1207: I'm assuming this is done to speed things up by avoiding snapshot generation.

On a related note, the `fetchResourceRaw` method would benefit from some JavaDoc indicating that this is the case.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/extensions/Extensions.java
```diff
@@ -11852,7 +11852,7 @@ public static List<String> getProfileSummaryList(StructureDefinition context) {
 // Select By Map
 
   public static Extension makeSelectByMap(String value) {
-    return new Extension(ExtensionConstants.EXT_SELECT_BY_MAP).setValue(new CanonicalType(value));
+    return new Extension(ExtensionConstants.EXT_SELECT_BY_MAP_OLD).setValue(new CanonicalType(value));
```
dotasek: Should this not be EXT_SELECT_BY_MAP_NEW?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/DefinitionNavigator.java
```diff
@@ -79,17 +80,19 @@ public DefinitionNavigator(IWorkerContext context, StructureDefinition structure
...
-      this.path = current().getPath(); // first element
+      this.globalPath = current().getPath(); // first element
     }
+    this.localPath = this.globalPath;
     names.add(nameTail());
   }
   
-  private DefinitionNavigator(IWorkerContext context, StructureDefinition structure, boolean diff, boolean followTypes, int index, String path, List<String> names, String type) {
-    this.path = path;
+  private DefinitionNavigator(IWorkerContext context, StructureDefinition structure, boolean diff, boolean followTypes, int index, String globalPath, String localPath, List<String> names, String type) {
```
dotasek, line 93: Is there a case where globalPath and localPath will not be equal? They appear to be so in all of the ComparisonTests, at least.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/TerminologyFunctions.java
dotasek, line 1: Are these new functions tested anywhere in the FHIRPathTests? I can't see any coverage for them.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/Renderer.java
```diff
@@ -191,38 +190,39 @@ protected XhtmlNode renderStatusRow(Base b, XhtmlNode tbl, XhtmlNode tr) {
     }
   }
 
-  public static void renderStatusSummary(RenderingContext context, Base base, XhtmlNode x, String version, String... metadataFields) {
+  public static boolean renderStatusSummary(RenderingContext context, Base base, XhtmlNode x, String version, String... metadataFields) {
```
dotasek, line 203: This could use a comment indicating that the return indicates whether or not the data has changed.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/SQLiteINpmPackageIndexBuilderDBImpl.java
```diff
@@ -1,4 +1,83 @@
...
 
-public class SQLiteINpmPackageIndexBuilderDBImpl {
+import org.hl7.fhir.utilities.filesystem.ManagedFileAccess;
+import org.hl7.fhir.utilities.json.model.JsonObject;
+import org.hl7.fhir.utilities.npm.NpmPackageIndexBuilder;
+
+import java.sql.*;
+
+public class SQLiteINpmPackageIndexBuilderDBImpl implements NpmPackageIndexBuilder.INpmPackageIndexBuilderDBImpl {
+  private String dbFilename;
```
dotasek, line 10: This appears to be unused.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ValidatorSettings.java
```diff
@@ -104,5 +106,20 @@ public List<String> getCertificateFolders() {
...
+
+  public String getMinVersion() {
+    return minVersion;
+  }
+
+  public String getMaxVersion() {
+    return maxVersion;
+  }
+
+  public void setMinVersion(String minVersion) {
```
dotasek, line 118: Neither of these settings appears to be in use in the codebase? I think this means the StructureValidator is always getting min and max from the context.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/VersionUtilities.java
```diff
@@ -778,4 +773,24 @@ public static String getResourceTypesUrl(String version) {
   }
 
 
+  public static List<String> iterateCoreVersions(String startVer, String endVer) {
```
dotasek, line 776: getCoreVersionRange seems like a better description here. And a minor nitpick; this could be an immutable list.

dotasek (changes requested):
See comments.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java
grahamegrieve: it's not so much to speed things up as to avoid circular dependencies in snapshot generation

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/DefinitionNavigator.java
grahamegrieve: yes once the navigator walks across a type boundary, they diverge

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/TerminologyFunctions.java
grahamegrieve: yes - see  https://github.com/FHIR/fhir-test-cases/blob/5f6c712a31ed0be3f5a7efc71c38078c29fe0e44/r5/fhirpath/tests-fhir-r5.xml#L1753

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/Renderer.java
grahamegrieve: ok

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/SQLiteINpmPackageIndexBuilderDBImpl.java
grahamegrieve: used in ValidationEngine

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ValidatorSettings.java
grahamegrieve: It's used in the IG Publisher

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/VersionUtilities.java
grahamegrieve: I don't think so - it gets given a range, and iterates the list inside the range

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java
grahamegrieve: ok I rewrote that whole section completely for consistency.

PR #2105, opened 2025-07-21
Split support package contents to prevent duplicate packages

dotasek:
The `bundle` packaging of dstu3 and dstu3.support produces duplicate classes. This creates a `support` package into which the original package tree can be put to prevent this duplication.


grahamegrieve (approved):
ok

PR #2107, opened 2025-07-22
Use Java 17 for build

dotasek:
Builds for Java 17, and also introduces tests for Java 21


org.hl7.fhir.utilities/src/test/java/org/hl7/fhir/utilities/tests/PackageCacheTests.java
```diff
@@ -45,18 +45,18 @@ public void testPath() throws IOException {
...
   }
 
   @Test
   public void testNotCaseSensitive() throws IOException {
     FilesystemPackageCacheManager cache = new FilesystemPackageCacheManager.Builder().withTestingCacheFolder().build();
-    cache.clear();    
-    Assertions.assertEquals(cache.loadPackage("KBV.Basis", "1.1.3").version(), "1.1.3");
-    Assertions.assertEquals(cache.loadPackage("kbv.basis", "1.1.3").version(), "1.1.3");    
+    cache.clear();
+    Assertions.assertEquals("1.7.0", cache.loadPackage("KBV.Basis", "1.7.0").version());
```
dotasek, line 58: This is an odd case: kbv.basis#1.1.3 contains a single entry that contains a character incompatible with Java 21+ on Windows.

My understanding here is that tgz files build file names based on the filename encoding on the system that created them. Version 1.7.0 of this IG no longer contains that entry, but I can't find a git repository to examine to see if this is something they intentionally fixed, or something that the IG publisher or some other tool fixed.

PR #2115, opened 2025-07-24
FHIRPath R5 engine $this vs focus for parameter args

brianpos:
Update the context for evaluating fhirpath arguments from focus to $this.
Consistency across the board.
This is how the Firely SDK does it, and how Bryn expected that things were going, and based on discussion, where things are going.
Also updated the processing of the `is` operation to work in the element model approach.
(and a few comments where functionality in the spec is missing)
(needs the updated tests project too)


grahamegrieve (approved):
ok

PR #2116, opened 2025-07-27
Fix issue when rendering a CDSHookRequest.  Was getting something tha…

lmckenzi:
…t looked like this:

```then the type is <<code>/>http://hl7.org/fhir/us/davnci-crd/StructureDefinition/CRDHookAppointmentBookContext<</code>/>```


grahamegrieve (approved):
ok

PR #2119, opened 2025-07-29
Fixed bad index error caused by no type existing (because there was a…

lmckenzi:
… contentReference instead)


lmckenzi:
Closing as other fixes I made to the IG seem to have made this go away, so apparently the issue was a side-effect of something else that was broken.

PR #2121, opened 2025-07-29
Only report missing lcoalization messages the first time

dotasek:
Only logs missing localization messages once. This also tracks by locale, in case multiple localizations are used in a single run.


grahamegrieve (approved):
ok

PR #2125, opened 2025-08-01
2025 08 gg tx cache

grahamegrieve:


org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/Utilities.java
```diff
@@ -141,6 +141,13 @@ public static boolean isValidId(String id) {
     return id.matches("[A-Za-z0-9\\-\\.]{1,64}");
   }
 
+  public static String[] appendStr(String[] array1, String[] array2) {
```
dotasek: I think concatStringArray would be a clearer name.

Also, this test will at least cover the code:

```java
@Test
  void testAppendStringArray() {
    String[] a = new String[] {"A", "B"};
    String[] b = new String[] {"C", "D", "E"};
    String[] result = Utilities.concatStringArrays(a, b);
    assertThat(result).hasSize(5);
    assertThat(result[0]).isEqualTo("A");
    assertThat(result[1]).isEqualTo("B");
    assertThat(result[2]).isEqualTo("C");
    assertThat(result[3]).isEqualTo("D");
    assertThat(result[4]).isEqualTo("E");
  }
```

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
```diff
@@ -4262,6 +4262,7 @@ private boolean checkAttachment(List<ValidationMessage> errors, String path, Ele
     long max = -1;
     byte[] cnt = null;
     byte[] hash = null;
+    String h64 = null;
```
dotasek: hash64 would make this clearer

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java
```diff
@@ -4638,6 +4638,22 @@ private List<Base> funcRound(ExecutionContext context, List<Base> focus, Express
...
+      List<Base> n1 = execute(context, baseToList(context.thisItem), expr.getParameters().get(0), true);
+      if (!n1.isEmpty()) {
+        name =convertToString(n1);
+      }
+    }
+    CommaSeparatedStringBuilder b = new CommaSeparatedStringBuilder(", ");
+    for (Base base : focus) {
+      b.append(base.fhirType()+":"+convertToString(base));
+    }
+    log.info("Debug Point '"+name+"': "+b.toString());
```
dotasek, line 4653: Shouldn't this be log.debug("...") ?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/validation/ValueSetValidator.java
```diff
@@ -263,8 +263,15 @@ public ValidationResult validateCode(String path, CodeableConcept code) throws F
           if (cs == null || (cs.getContent() != CodeSystemContentMode.COMPLETE && cs.getContent() != CodeSystemContentMode.SUPPLEMENT)) {
             if (context.isNoTerminologyServer()) {
               if (c.hasVersion()) {
-                String msg = context.formatMessage(I18nConstants.UNKNOWN_CODESYSTEM_VERSION, c.getSystem(), c.getVersion() , resolveCodeSystemVersions(c.getSystem()).toString());
-                unknownSystems.add(c.getSystem()+"|"+c.getVersion());
+                Set<String> set = resolveCodeSystemVersions(c.getSystem());
+                String msg;
+                if (set.isEmpty()) {
```
dotasek: This looks like repeated logic from around line 628. Is there maybe some way to create a helper method for this?

dotasek (changes requested):
See comments for requested changes.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/Utilities.java
grahamegrieve: ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/Property.java
```diff
@@ -57,11 +58,15 @@ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWIS
...
 import org.hl7.fhir.utilities.Utilities;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @MarkedToMoveToAdjunctPackage
+@Slf4j
 public class Property {
 
-	private IWorkerContext context;
+  private static final Logger log = LoggerFactory.getLogger(Property.class);
```
dotasek: This explicit instantiation isn't needed. We should only be using the '@Slf4j' annotation unless there's some really tricky logging stuff that necessitates it.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/extensions/ExtensionUtilities.java
```diff
@@ -984,6 +967,26 @@ public static void addUriExtension(DomainResource dr, String url, String value)
...
+    if (!StringUtils.isBlank(content)) {
+      Extension ex = getExtension(e, url);
+      if (ex != null)
+        ex.setValue(new CanonicalType(content));
+      else
+        e.getExtension().add(Factory.newExtension(url, new CanonicalType(content), true));
+    }
+  }
+
+  public static void addCanonicalExtension(DomainResource dr, String url, String value) {
```
dotasek, line 990: I believe there should be a null check here as well for 'value'.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/extensions/ExtensionUtilities.java
```diff
@@ -984,6 +967,26 @@ public static void addUriExtension(DomainResource dr, String url, String value)
       dr.getExtension().add(Factory.newExtension(url, new UriType(value), true));
   }
 
+
+  public static void addCanonicalExtension(Element e, String url, String content) {
+    if (!StringUtils.isBlank(content)) {
```
dotasek: An early return null check would be slightly cleaner:

```java
if (StringUtils.isBlank(content)) { return; }
```

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/CodeSystemUtilities.java
```diff
@@ -658,6 +658,37 @@ public static boolean hasProperty(ConceptDefinitionComponent c, String code) {
     return false;
   }
 
+  public static String getProperty(CodeSystem cs, ConceptDefinitionComponent c, String uri, String code) {
```
dotasek: Calls for this and hasProperty all seem to have non-null params in the code, but an explicit javax.annotation.Nonnull might be good to include if we're not checking for nulls here.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/CodeSystemUtilities.java
```diff
@@ -1122,5 +1153,25 @@ public static CodeSystem convertSD(StructureDefinition sd) {
...
+    return hasProperty(cs, c, "http://hl7.org/fhir/concept-properties#comments", "comments");
+  }
+
+
+  public static String getCSComments(CodeSystem cs, ConceptDefinitionComponent c) {
+    return getProperty(cs, c, "http://hl7.org/fhir/concept-properties#comments", "comments");
+  }
+
+  public static void addCSComments(CodeSystem cs, ConceptDefinitionComponent cc, String comments) {
+    String code = "comments";
```
dotasek, line 1165: We just leave 'comments' if no explicit comments are found?

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/CDANarrativeFormat.java
```diff
@@ -257,6 +258,12 @@ private void processTr(Element e, XhtmlNode xn) throws FHIRException {
     processChildren(e, xc);
   }
 
+  private void processLinkHtml(Element e, XhtmlNode xn) throws FHIRException {
+    XhtmlNode xc = xn.addTag("linkHtml");
+    processAttributes(e, xc, "ID", "language", "styleCode", "align", "char", "charoff", "valign");
```
dotasek, line 263: This sequence of attributes is repeated a few times. A good descriptive constant would be useful.

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/misc/ProfileVersionAdaptor.java
```diff
@@ -459,7 +458,11 @@ private String adaptPath(String path) {
     if (sd == null) {
       StructureDefinition ssd = sCtxt.fetchTypeDefinition(base);
       if (ssd != null && ssd.getKind() == StructureDefinitionKind.RESOURCE) {
-        return "Basic";
+        if ("CanonicalResource".equals(base)) {
```
dotasek, line 469: What is adaptPath actually doing? It looks like it's reducing the path to it's base, but making an exception for Resource. This gets returned to `np` when this is called, but what does `np` mean? `newPath` is my guess, but this doesn't tell me much.

dotasek (changes requested):
See comments.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java
dotasek: Answered in zulip. @grahamegrieve wants this info printed without having to set debug level logging.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/validation/ValueSetValidator.java
grahamegrieve: ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/Property.java
grahamegrieve: weird, I wouldn't have added that. I've removed it

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/extensions/ExtensionUtilities.java
grahamegrieve: ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/extensions/ExtensionUtilities.java
grahamegrieve: ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/CodeSystemUtilities.java
grahamegrieve: Yes. It's a property defined in the core spec that can be used without definition. I'm checking here that it's not defined differently

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/CodeSystemUtilities.java
grahamegrieve: ok

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/CDANarrativeFormat.java
grahamegrieve: really? Actually, it's not. they're all subtly different. I don't think this is a good idea.

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/misc/ProfileVersionAdaptor.java
grahamegrieve: this is just adding a special case for a particular kind of past brokenness. I'll rewrite it more clearly in the future when additional infrastructure becomes available

PR #2131, opened 2025-08-09
Update, primarily for use of {} as bracketing

lawley:
Switch to {} for bracketing codeLists and filterLists (avoid infinite lookahead).
Expand scope of language coverage: "of" operator and nesting via implicit VCL ValueSets


grahamegrieve (approved):
ok

PR #2133, opened 2025-08-13
Split ValidationContext into task-specific classes 

dotasek:
# DO NOT MERGE YET

The ValidationContext class contains many fields which do not have to co-exist for specific tasks to function. This PR will split this class into separate classes, so that fields specific to particular tasks are not available in all of them.

For example, LangTransformTask utilizes `srcLang` and `tgtLang`. These are irrelevant to validation and other tasks, but are present in the same 400-line parsing method that handles all the other args from the `main(String[])` call to the CLI.

Some fields will overlap by design, particularly those involved in ValidationEngine construction and initialization, which should be moved to a ValidationEngineSettings class that can be used to initialize the correct engine for all tasks that need one.

The ValidationContext class that these new classes will replace will remain, but will be deprecated. Due to validation-wrapper's use of this class and the need for a backwards compatible web api (primarily to support Inferno's use of this app), this class cannot be removed immediately.

This change will make it easier to offload our command line parsing and logic to a third party library, like Picocli.


dotasek:
No longer needed, this PR contains new changes: https://github.com/hapifhir/org.hl7.fhir.core/pull/2235

PR #2135, opened 2025-08-14
Add codes 25 and 26 to ObjectRole

qligier:
Codes 25 and 26 were recently added to the [CodeSystem AuditEventEntityRole](https://terminology.hl7.org/CodeSystem-object-role.html), let's add them to the R4 enum.


grahamegrieve (approved):
ok

Issue #2021, opened 2025-05-26
Existing cross version extension is needlessly kept for R3 -> R4 -> R3 conversions


dotasek:
As described here: https://github.com/hapifhir/org.hl7.fhir.core/pull/2013#discussion_r2105784228

When the convertors convert from a newer version to an older one (R4 -> R3), any extensions from an earlier conversion (R4 -> R3, for example), should be removed. As noted in the comment above, this behaviour is inconsistent in our conversion code.

> for example the mapping of category on CarePlanActivityDetailComponent with the EXT_CarePlanActivityDetailComponentExtension. Where it seems like the extension is kept between conversions (unless I am missing something).

> https://github.com/hapifhir/org.hl7.fhir.core/blob/master/org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/conv30_40/resources30_40/CarePlan30_40.java

> However now that I looked into it again, it seems like in other cases these extensions are excluded from the target. So it seems this isn't consistent at the moment.

Corne:
I'd like to work on this. 
Shouldn't this be made consistent for all version conversions not only R3 <-> R4?

grahamegrieve:
Sure, any version, any conversion. But it's really big rearchitecting of everything, so nothing has happened. Yet. And also, it's got all sorts of dangerous corner cases that we're just starting to work through as a community, so another reason nothing has happened yet

PR #2141, opened 2025-08-20
Don't keep MODIFIER_REASON extension when converting ElementDefintion from dstu3 to r4 (part of #2021)

Corne:
PR is part of https://github.com/hapifhir/org.hl7.fhir.core/issues/2021


grahamegrieve (approved):
ok

PR #2143, opened 2025-08-21
Don't keep ACTIVITY_DETAIL_COMPONENT extension when converting ElementDefintion from dstu3 to r4 (part of #2021)

Corne:
PR is part of https://github.com/hapifhir/org.hl7.fhir.core/issues/2021

- Added some tests
- Also cleaned up some duplicate if's
- Renamed Extension so it's SCREAMING_SNAKE_CASE


grahamegrieve (approved):
ok

PR #2148, opened 2025-08-24
Don't map EXT_PAT_ANIMAL when converting back r4 to dstu3 (part of #2021)

Corne:
PR is part of #2021

Prevents the mapping of EXT_PAT_ANIMAL when converting back to dstu3 from r4.


grahamegrieve (approved):
ok

PR #2149, opened 2025-08-26
mulithreading issues in R5 validation with defaultDisplayLanguage / locale /

oliveregger:
we encountered memory and multithreading issues with our latest matchbox version which incorporates the java validator.

1.) we have seen a huge memory increase when performing the same validation multiple times. Root cause we think is multiple defaultDisplayLanguage parameters that are added to the expansion Parameters, this leads to that for every validation new cache token are generated, this PR reduces it to one instance

2) the modification with the defaultDisplayLanguage in expansionParameters led in our setup to a Concurrent Modification Exception, we could solve that, that we make a copy before serializing it in generationCacheToken

3) Similiar setting the locale again led to a Nullpointer Exception in the BaseContext because this.messages was zero, we set the local only explicitly now if it changed.

I don't know how I could reproduce this directly in the ValidatorCli but we might think it might also have an impact in ValidatorCli or derived solutions. see also https://github.com/ahdis/matchbox/issues/425

### DIFF
--- a/org.hl7.fhir.r5/pom.xml
+++ b/org.hl7.fhir.r5/pom.xml
@@ -23,6 +23,7 @@
 		<dependency>
 			<groupId>ca.uhn.hapi.fhir</groupId>
 			<artifactId>org.hl7.fhir.utilities</artifactId>
+			<version>6.6.5</version>
 		</dependency>
 
 		<dependency>
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
@@ -347,6 +347,9 @@ public abstract class BaseWorkerContext extends I18nBase implements IWorkerConte
   protected String userAgent;
   protected ContextUtilities cutils;
   private List<String> suppressedMappings;
+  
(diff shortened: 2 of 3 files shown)


dotasek (changes requested):
I agree with the change to using a copy of the expParameters to avoid concurrentModificationExceptions. I do feel like it could be moved up one level to where BaseWorkerContext calls TerminologyCache though, to keep all the concurrency management isolated to the single class (any expParameters exposed are copies, and not references)

I don't think the changes to setLocale are correct. 

super.setLocale(locale) should be the only call necessary to track locale via a field.


 @grahamegrieve, could you clarify what the expected expParameters should contain when a locale is set? From what existed previously I'd suggest the following.

* there should be two parameters, one named 'displayLanguage' and one named 'defaultDisplayLanguage' which are both set to the languageTag of that locale.
* at any point, there should not be multiple params with those same names.
* if the locale has changed, these two params values should just be reset to the new languageTag. No new params should be generated.

dotasek:
@oliveregger the following PRs have a test that replicates the original error message and some fixes to expansionParameter setting that should remove duplicates:

https://github.com/hapifhir/org.hl7.fhir.core/pull/2160
https://github.com/hapifhir/org.hl7.fhir.core/pull/2156

These make this PR no longer applicable. I'd like to close it. If multithread issues persist after these PRs, we should open another issue.

PR #2150, opened 2025-08-26
Pick ips 2.0.0-ballot version to replace not found 2.0.0 in -ips option

dotasek:
The `-ips` option uses 2.0.0 of IPS, which is not yet released. This switches over to the 2.0.0-ballot version, which is available.


grahamegrieve (approved):
ok

PR #2152, opened 2025-08-27
Fix for invalid package versions in package listing JSON

dotasek:
If an invalid version existed in the package listing JSON, it would produce an exception when comparing, and cause the entire loading process to fail. 

This reports invalid versions via logging, and doesn't accept invalid versions as the current version of an IG.

### DIFF
--- a/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/PackageClient.java
+++ b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/PackageClient.java
@@ -246,21 +246,21 @@ public class PackageClient {
   }
 
   protected PackageInfo getPackageInfoFromJSON(JsonObject o, String name, String canonical, String fhirVersion) {
-      String id = o.asString("npm-name");
-      String pname = o.asString("name");
-      String pcanonical = o.asString("canonical");
-      String description = o.asString("description");
-      Instant d = o.has("date") ? o.asDate("date") : null;
+      String packageId = o.asString("npm-name");
+      String packageName = o.asString("name");
+      String packageCanonical = o.asString("canonical");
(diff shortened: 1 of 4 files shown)


org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/PackageClient.java
```diff
@@ -246,41 +249,50 @@ public String getLatestVersion(String id, String specVersion) throws IOException
...
-            String v = e.asString("ig-version");
-            if (version == null || VersionUtilities.isThisOrLater(version, v, VersionUtilities.VersionPrecision.MINOR)) {
-              version = v;
-              fVersion = e.getJsonArray("fhir-version").get(0).asString();
-              url = e.asString("url");
+        // Get the latest version out of the available valid editions of this IG.
+        for (JsonObject edition : o.getJsonObjects("editions")) {
+          if (fhirVersion == null || fhirVersion.equals(edition.asString("fhir-version"))) {
+            String igVersion = edition.asString("ig-version");
+            try {
```
grahamegrieve: Wouldn't it be better to just check if .isValidSemver() rather than try...catch and exposing private methods?

dotasek: I moved up the method that caused the initial exeption, but now that I read this, I realize only explicit versions should be used here. I'll use your approach.

dotasek: Fixed in https://github.com/hapifhir/org.hl7.fhir.core/pull/2152/commits/2a689db12f87b7ecdd38dad96d416326d35ddb0d

PR #2154, opened 2025-08-28
Make ips:au param use new definitions from hl7.fhir.au.ps

dotasek:
The `-ips:au` param in the validator CLI was using an older version of the IG. This has been updated with the new IG name and updated profile URLs


grahamegrieve (approved):
ok

PR #2155, opened 2025-08-28
Document debug and trace logging + warn CLI users if old debug param is used

dotasek:
This adds documentation to the `-help` option to provide information on debug and trace logging. 

It will also print a warning message if the older `-debug` option is used.


grahamegrieve (approved):
ok

PR #2160, opened 2025-09-02
Atomically set expansionParameters in BaseWorkerContext

dotasek:
BaseWorkerContext modified the contents of expansionParameters and passed references to them. When accessed from multiple threads, it was possible for the lists in Parameters to be modified (by `setLocale` in particular) while another thread was in the process of reading.

This ensures that expansionParameters is only ever returned as a copy, and that whenever its value is modified, it is done so with an AtomicReference `set` call instead of modified directrly.


grahamegrieve (approved):
ok

PR #2162, opened 2025-09-05
Preparatory refactors and deprecations for ValidationContext split

dotasek:
This refactors some of the code used to generate ValidationEngine instances for tasks in the CLI and in the validator-wrapper project.

Its primary purpose is to move ValidationContext processing code into fewer places to enable replacement with more modular POJOs. Before ValidationContext is completely removed, methods can be put in these places to derive those modular POJOs to ensure backward compatibility.


dotasek:
Including this question from @grahamegrieve so we can track design decisions.

> do we use any of the deprecated methods?

Yes. ValidationContext is used to pass params from the CLI and from validator-wrapper. These should be maintained to give downstream projects like matchbox and Inferno (@oliveregger, Stephen MacVicar) a period to adjust.

The intent here is to eventually get rid of them and replace them in the following manner:

```java
//Old way
ValidatorEngine engine = validationService.initializeValidator(validationContext, definitions, timetracker);

// Note that there's no option to change things like hintAboutNonMustSupport, anyExtensionsAllowed, etc. as these are pulled from the engine.
InstanceValidator instanceValidator = getValidator(FhirFormat.JSON);
```

```java
//New way. 
Note that definitions is gone because of the changes in the PR, which automatically build it from validationEngineSettings.

ValidatorEngine engine = validationServive.buildEngine(validationEngineSettings, TimeTracker tt)

// Now we have a validationEngineSettings, which can be used to set hintAboutNonMustSupport, anyExtensionsAllowed, etc. So now an existing ValidationEngine has fewer reasons to change and is easier to re-use by things like server mode.
InstanceValidator instanceValidator = getValidator(validationInstanceSettings, FhirFormat.JSON);
```

PR #2170, opened 2025-09-11
#2021 don't map EXT_CS_PROFILE when converting r4 back to dstu3

### LINKED ISSUE #2021: Existing cross version extension is needlessly kept for R3 -> R4 -> R3 conversions

Corne:
This should be the last extension that was kept between versions for R4 and DSTU3 
So closes #2021

### DIFF
--- a/org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/conv30_40/resources30_40/CapabilityStatement30_40.java
+++ b/org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/conv30_40/resources30_40/CapabilityStatement30_40.java
@@ -18,7 +18,7 @@ import java.util.stream.Collectors;
 public class CapabilityStatement30_40 {
 
   private static final String[] IGNORED_EXTENSION_URLS = new String[]{
-    VersionConvertorConstants.EXT_ACCEPT_UNKNOWN_EXTENSION_URL
+    VersionConvertorConstants.EXT_ACCEPT_UNKNOWN_EXTENSION_URL, VersionConvertorConstants.EXT_CS_PROFILE
   };
 
   public static org.hl7.fhir.dstu3.model.CapabilityStatement convertCapabilityStatement(org.hl7.fhir.r4.model.CapabilityStatement src) throws FHIRException {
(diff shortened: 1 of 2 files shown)


org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/conv30_40/resources30_40/CapabilityStatement30_40.java
```diff
@@ -65,20 +66,23 @@ public static org.hl7.fhir.dstu3.model.CapabilityStatement convertCapabilityStat
...
         for (org.hl7.fhir.r4.model.CanonicalType t : rr.getSupportedProfile())
           tgt.addProfile(Reference30_40.convertCanonicalToReference(t));
+    }
     for (org.hl7.fhir.r4.model.Extension ext : src.getExtension()) {
       if (VersionConvertorConstants.EXT_CS_PROFILE.equals(ext.getUrl())) {
         tgt.addProfile(Reference30_40.convertReference((org.hl7.fhir.r4.model.Reference) ext.getValue()));
       }
     }
-    for (org.hl7.fhir.r4.model.CapabilityStatement.CapabilityStatementRestComponent t : src.getRest())
-      tgt.addRest(convertCapabilityStatementRestComponent(t));
```
Corne, line 81: Note this is moved to line 76 where we already looped over 'rest' items

PR #2179, opened 2025-09-21
hapifhir/org.hl7.fhir.core#2178 Update Hapi Fhir dependency

### LINKED ISSUE #2178: Accessing Reference Object using reflection might cause unexpected behavior when using the Hapi-fhir 8.4 (with fhir core 6.6.7)
**Background** 
* In Hapi-FHIR-base 8.4 the interface org.hl7.fhir.instance.model.api.IBaseReference was modified and getIdentifier and setIdentifier was added as default methods.
* The fhir core artifact package 6.6.7 is still compiled with Hapi-fhir 6.4.1

**Why is this causing an issue:**
Java is determining some method invocation order things already during the compile phase which influence the bytecode. Since the fhir core artifact classes use an older version of hapi-fhir-base, the new introduced interface methods are not visible to the compiler. The compile do not detect, that the methods getIdentifier in `org.hl7.fhir.r4.model.Reference` is a overriden method definition from the default method in the interface (in newer releases).

**In which situation do we have an issue:**
I identitfy an issue when using hapi-fhir in context of groovy. Groovy is havily using reflection and when trying to access the "identifier" of a reference object, it now always return null, because the JVM invoke the default method of the interface instead of the instance method.

**Resolving the issue:**
Short term, updating the hapi-fhir version to 8.4.0 should fix the issue in upcoming releases of the fhir core artifact.

Thopap:
Update Hapi Fhir dependency to resolve #2178


Thopap:
@grahamegrieve Can you please review this PR?

grahamegrieve:
no, this one is for @dotasek

dotasek (changes requested):
Apologies for the delay, there were many updates that needed review and merge before I got to this.

This looks good to me, but needs a merge from master. @Thopap you can use https://github.com/hapifhir/org.hl7.fhir.core/tree/do-20251125-update-hapi to see how I resolved this, and do the same thing to this branch so you get credit for the PR.

To review, I first created another branch in order to merge recent changes from master.

This was then used in two runs of the QA Delta pipeline for the IG publisher (as this is normally where dependency related issues show up).

The resulting QA files were functionally the same, differing only on generated OIDs, which is correct.

This is the branch: https://github.com/hapifhir/org.hl7.fhir.core/tree/do-20251125-update-hapi

This is the relevant GIT hash: https://github.com/hapifhir/org.hl7.fhir.core/commit/c74135f024a3d0850c2f34978fd41f51a6bede31

To verify that the QA Diff pipeline was using the updated version, this run output contains the reference to the commit (see https://github.com/HL7/fhir-ig-publisher/actions/runs/19675034458/job/56353725040):
```sh
  /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --depth=1 origin c74135f024a3d0850c2f34978fd41f51a6bede31
  From https://github.com/hapifhir/org.hl7.fhir.core
   * branch            c74135f024a3d0850c2f34978fd41f51a6bede31 -> FETCH_HEAD
```

The two QA Delta runs used for comparison are:

branch -> https://github.com/HL7/fhir-ig-publisher/actions/runs/19675034458
master -> https://github.com/HL7/fhir-ig-publisher/actions/runs/19676769578

pom.xml
```diff
@@ -413,6 +419,12 @@
                         <redirectTestOutputToFile>false</redirectTestOutputToFile>
                         <!-- For JUnit5 use an excludedInSurefire tag to exclude a test from surefire executions -->
                         <excludedGroups>excludedInSurefire</excludedGroups>
+                        <includes>
```
dotasek: Why would this be needed? Current tests run with JUnit default patterns.

pom.xml
```diff
@@ -358,6 +358,12 @@
                 <artifactId>assertj-core</artifactId>
                 <version>3.22.0</version>
             </dependency>
+            <dependency>
```
dotasek: This should be removed.

In my last review, I discovered we were using hamcrest internally, which is a mistake, as we should be using assertj only. 

I fixed this, so now it shouldn't be reintroduced in the PR either.

See https://github.com/hapifhir/org.hl7.fhir.core/pull/2243

dotasek (changes requested):
See comments regarding the removal of hamcrest and include patterns for Surefire.

Otherwise, this looks good to me, and I will be happy to merge it once the changes are made.

Thank you for your patience.

pom.xml
Thopap: Running mvn effective pom, it shows, that the include config is set to "*Test.java" only, since hapi 8.4.0 explicit set this (was not the case in 6.4.0, see https://github.com/hapifhir/hapi-fhir/blob/v8.4.0/pom.xml#L2610). Since this project use also other test class naming, some adoption is required to avoid certain tests are no longer executed by the surefire pugin.

To resolve this, i see two options:
* Explicitly set the junit default to override hapi parent pom
* OR setting `configuration combine.self="override"` to disable the mixture of the inheritance of the config from the hapi parent pom.

I currently pick option A, but i can also switch to option B if you prefer this.

dotasek: Thanks for taking a look at that, it's an excellent catch. Option B sounds cleaner, could you please commit that, and include a `<!-- comment -->` that gives a brief description of why?

PR #2181, opened 2025-09-22
Corrected summary labels to reflect that the statistics are for the d…

lmckenzi:
…ifferential, not the whole model

https://github.com/HL7/fhir-ig-publisher/issues/415


grahamegrieve (commented):
I think that 'new' is troublesome here, because it might make people think about 'new' in a version sense, not a structural set. What about 'in this profile:' ?

PR #2186, opened 2025-09-24
Fix OWASP check failure

dotasek:


dotasek:
Fixed in https://github.com/hapifhir/org.hl7.fhir.core/pull/2188

PR #2187, opened 2025-09-25
2025 09 gg misc

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/LanguageUtils.java
```diff
@@ -861,8 +861,10 @@ public XhtmlNode divForLang(XhtmlNode xhtml, String lang, String resourceLang, S
     // (if it wasn't generated) or will leave the narrative as null, which will force it to be generated for this language
     if (!lang.equals(defaultLang)) {
       XhtmlNode defaultDiv = divForLang(xhtml, defaultLang, resourceLang, defaultLang, null);
-      if (scanForGeneratedNarrative(defaultDiv, defaultLang)) {
-        return null;
+      if (defaultDiv != null) {
```
grahamegrieve, line 864: @lmckenzi what is supposed to happen in this case?

lmckenzi: If there's a "default" div (i.e. one that doesn't declare a language), it checks to see if that div has the special text in it that says "hi, I'm generated narrative".  If it does, it'll return null, which will in turn force the narrative generator to generate new narrative for the needed language.  (That happens downstream.)  If the default narrative isn't generated narrative, then we keep what's there, but spit out a warning message noting that we've got narrative that's not correct for the target language.

PR #2191, opened 2025-09-29
refactor IWorkerContext

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java
```diff
@@ -440,6 +427,7 @@ public ElementDefinition getExtensionValueDefinition() {
   private List<StructureDefinition> obligationProfiles = new ArrayList<>();
   private boolean wantThrowExceptions;
   private List<String> suppressedMappings= new ArrayList<>();
+  @Getter @Setter private Parameters parameters;
```
dotasek, line 430: I don't see this set anywhere.

org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/tests/utilities/TestUtilities.java
```diff
@@ -34,6 +37,12 @@ public static ValidationEngine getValidationEngine(java.lang.String src, java.la
     TerminologyCache.setCacheErrors(true);
     validationEngine.setLanguage("en-US");
     validationEngine.setLocale(Locale.US);
+
```
dotasek, line 40: This section should be in two places:

https://github.com/hapifhir/org.hl7.fhir.core/blob/82a9b3c724827e33d31c020f6a10d5adb02060f9/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ValidationEngine.java#L417
https://github.com/hapifhir/org.hl7.fhir.core/blob/82a9b3c724827e33d31c020f6a10d5adb02060f9/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ValidationEngine.java#L437

BUT, I have a question regarding this: Should packageManager ever change within SimpleWorkerContext? 

I suspect it should not, in which case, it should be made final and set using SimpleWorkerContext.SimpleWorkerContextBuilder

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/IWorkerContext.java
```diff
@@ -288,248 +299,282 @@ public String getUrl() {
...
+   *
+   * It's an error if the second form doesn't agree with class_. It's an
+   * error if class_ is null for the last form
+   *
+   * class can be Resource, DomainResource or CanonicalResource, which means resource of all kinds
+   *
+   * @param class_ the type of resource
+   * @param uri the URL of the resource, optionally with a |version suffix
+   * @param version the version. Don't provide both a version and a |version suffix
+   * @param sourceOfReference where the reference was found
```
dotasek: sourceOfReference has a special behaviour when it is null, which isn't mentioned here.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/IWorkerContext.java
```diff
@@ -288,248 +299,282 @@ public String getUrl() {
...
+   *  - a logical id e.g. [id]
+   *
+   * It's an error if the second form doesn't agree with class_. It's an
+   * error if class_ is null for the last form
+   *
+   * class can be Resource, DomainResource or CanonicalResource, which means resource of all kinds
+   *
+   * @param class_ the type of resource
+   * @param uri the URL of the resource, optionally with a |version suffix
+   * @param version the version. Don't provide both a version and a |version suffix
```
dotasek, line 354: If the url doesn't include a version, and the version param is also null, what is the expected behaviour?

dotasek (changes requested):
This is an incomplete review:

There are several questions regarding next steps for this PR.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java
grahamegrieve: no. work for me to do yet . Though I see I forgot to go back and initialise it.

org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/tests/utilities/TestUtilities.java
grahamegrieve: I generally agree that it belongs in the SimpleWorkerContext builder but for test utilities, it's building a validation engine, and that handles the SimpleWorkerContext internally, so I wasn't sure how to thread that needle

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/IWorkerContext.java
grahamegrieve: updated doco

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
```diff
@@ -158,9 +158,13 @@ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWIS
 
 @Slf4j
 @MarkedToMoveToAdjunctPackage
-public abstract class BaseWorkerContext extends I18nBase implements IWorkerContext {
+public abstract class BaseWorkerContext extends I18nBase implements IWorkerContext, IWorkerContextManager, IOIDServices {
   private static boolean allowedToIterateTerminologyResources;
 
+  public interface ITerminologyOperationDetails {
```
dotasek: Moving this here creates an issue in HAPI where now ITerminologyOperationDetails needs to be accessed from an implementation (BaseWorkerContext), where before it was from another interface.

I think it should go back to being in the interface, so HAPI can avoid using an implementation.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/IWorkerContext.java
grahamegrieve: updated doco

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
grahamegrieve: I don't know what this is about - I didn't move anything about this

org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/tests/utilities/TestUtilities.java
dotasek: Should what happens in TestUtilities not be the default behaviour for a building any ValidationEngine?

If not, how would it differ from a default behaviour? That would indicate what needs to be added to the builders to prevent these lines from being needed in TestUtilities.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/IWorkerContext.java
```diff
@@ -1,284 +1,295 @@
...
+ * If you provide your own implementation, you need to handle packages properly in
+ * order to get correct version resolution
  *  
  * @author Grahame
  */
 
 @MarkedToMoveToAdjunctPackage
 public interface IWorkerContext {
 
-  public interface ITerminologyOperationDetails {
```
dotasek, line 117: @grahamegrieve  This is the deletion that I believe moved ITerminologyOperationDetails to BaseWorkerContext

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/utilities/ValueSetProcessBase.java
```diff
@@ -8,7 +8,7 @@
 import org.hl7.fhir.exceptions.TerminologyServiceException;
 import org.hl7.fhir.r5.context.ContextUtilities;
 import org.hl7.fhir.r5.context.IWorkerContext;
-import org.hl7.fhir.r5.context.IWorkerContext.ITerminologyOperationDetails;
+import org.hl7.fhir.r5.context.BaseWorkerContext.ITerminologyOperationDetails;
```
dotasek: @grahamegrieve This is an example of how ITerminologyOperationDetails is now bound to an implementation rather than and interface like IWorkerContext.

org.hl7.fhir.validation/src/test/java/org/hl7/fhir/validation/tests/utilities/TestUtilities.java
grahamegrieve: what happens in test utilities in this regard should be normal, though right now I'm just trialing stuff, not yet ready for general release

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/utilities/ValueSetProcessBase.java
grahamegrieve: ok. I forgot about that. I moved it back, but I'm not thrilled about how it's designed.

PR #2197, opened 2025-10-03
Add tests to Params in preparation for more refactoring

dotasek:
Adds tests to Params, trying to cover as many common cases in translating args to ValidationContext as possible.

This will allow greater confidence when args are separated according to task.


grahamegrieve (approved):
ok

PR #2200, opened 2025-10-06
Refactor ValidationEngine Tasks

dotasek:
WIP

This PR will move ValidationContext instantiation into Task implementations. Once in those locations, the Params processing can start to be split according to task, and hopefully allow us to move fields out of ValidationContext.


grahamegrieve (approved):
ok

PR #2207, opened 2025-10-15
Added -ignoreList

jkiddo:
This PR is a code-cleanup refactoring of the RePackageTask as well as the introduction of an `-ignore-list` - that takes a comma separated list of canonicals and removes any resources that matches in that list. This is especially useful for removing CodeSystems with no content but regardless is included in the repackaging. This is even more inconvenient when using the HAPI FHIR stack as it comes with embedded codesystems and valuesets which are potentially shadowed if these resources cannot be ignored.

### DIFF
--- a/org.hl7.fhir.validation.cli/src/main/java/org/hl7/fhir/validation/cli/param/Params.java
+++ b/org.hl7.fhir.validation.cli/src/main/java/org/hl7/fhir/validation/cli/param/Params.java
@@ -620,7 +620,7 @@ public class Params {
         else
           throw new Exception("Can only nominate a single -fhirpath parameter");
       } else if (args[i].equals(SCOPE) || args[i].equals(IGNORE_LIST)) {
-        // this is only here to skip the step below which is not helping
+        //These two params are processed later by the RePackageTask and not included in ValidationContext.
         i++;
       } else if (!Utilities.existsInList(args[i],
         //The following params are handled outside this loop, so should be ignored.
--- a/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
+++ b/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
@@ -77,17 +77,13 @@ public class PackageReGenerator {
   }
 
   public static void main(String[] args) throws Exception {
-    String tmpdir = Files.createTempDirectory("repackage").toFile().getAbsolutePath();
 
     new PackageReGenerator()
-      .addPackage("ch.fhir.ig.ch-core")
+      .addPackage("hl7.fhir.us.davinci-alerts")
       .setJson(true)
-      .addIgnoreList(List.of("http://fhir.ch/ig/ch-core/StructureDefinition/ch-core-patient-epr", "http://fhir.ch/ig/ch-core/StructureDefinition/ch-core-claim", "urn:ietf:bcp:47"))
       .setOutputType(ExpansionPackageGeneratorOutputType.TGZ)
-      .setOutput(tmpdir + "/some-output.tgz")
-      .setModes(Set.of("cnt"))
+      .setOutput("/Users/grahamegrieve/temp/vs-output.tgz")
       .generateExpansionPackage();
-
   }
   
   public enum ExpansionPackageGeneratorOutputType {
@@ -321,7 +317,7 @@ public class PackageReGenerator {
       return;
     }
     set.add(res.getVersionedUrl());
-    
+
     if (scope == ExpansionPackageGeneratorScope.EVERYTHING || !isCore(res.getSourcePackage())) {
       resources.add(res);
       if (res.hasSourcePackage()) {


org.hl7.fhir.validation.cli/src/main/java/org/hl7/fhir/validation/cli/tasks/RePackageTask.java
```diff
@@ -56,48 +56,45 @@ public void logHelp(Logger logger) {
...
+    PackageReGenerator packageReGenerator = new PackageReGenerator()
+      .setContext(validationEngine.getContext())
+      .setOutput(output)
+      .setOutputType(getExpansionPackageGeneratorOutputType(output))
+      .setJson(json)
+      .setModes(validationContext.getModeParams())
+      .setNpmId(validationContext.getPackageName())
+      .addPackages(validationContext.getIgs());
+
+    switch (Objects.requireNonNull(Params.getParam(args, "-scope")))
```
dotasek: I realize this is left from previous code, but it really should reference the constant in Params.

org.hl7.fhir.validation.cli/src/main/java/org/hl7/fhir/validation/cli/tasks/RePackageTask.java
```diff
@@ -56,48 +56,45 @@ public void logHelp(Logger logger) {
...
     }
+
     if (validationContext.getExpansionParameters() != null) {
       validationEngine.loadExpansionParameters(validationContext.getExpansionParameters());
     }
-    ep.setOutput(output).setOutputType(t).setJson(json);
-    ep.setModes(validationContext.getModeParams());
-    ep.generateExpansionPackage();
+
+    String ignoreList = Params.getParam(args, "-ignore-list");
```
dotasek: This should reference the constant in Params.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
```diff
@@ -80,10 +77,12 @@ public static class TerminologyResourceEntry {
 
   public static void main(String[] args) throws Exception {
     new PackageReGenerator()
-      .addPackage("hl7.fhir.us.davinci-alerts")
+      .addPackage("ch.fhir.ig.ch-core")
       .setJson(true)
+      .addIgnoreList(List.of("http://fhir.ch/ig/ch-core/StructureDefinition/ch-core-patient-epr","http://fhir.ch/ig/ch-core/StructureDefinition/ch-core-claim", "urn:ietf:bcp:47"))
       .setOutputType(ExpansionPackageGeneratorOutputType.TGZ)
-      .setOutput("/Users/grahamegrieve/temp/vs-output.tgz")
+      .setOutput("/Users/jkiddo/temp/some-output.tgz")
```
dotasek: I can see that this follows the same pattern as existing code, but I think we should be using the Params.OUTPUT param to manage this file.

@grahamegrieve can we remove the references to specific users filesystems here?

dotasek (changes requested):
Some initial notes; after this, I still need time to review the actual logic changes.

@grahamegrieve would probably have to take a look at the intent of this code to verify that these changes are valid.

org.hl7.fhir.validation.cli/src/main/java/org/hl7/fhir/validation/cli/param/Params.java
```diff
@@ -617,6 +619,9 @@ else if (args[i].equals(HTML_OUTPUT)) {
             validationContext.setFhirpath(args[++i]);
         else
           throw new Exception("Can only nominate a single -fhirpath parameter");
+      } else if (args[i].equals(SCOPE) || args[i].equals(IGNORE_LIST)) {
+        // this is only here to skip the step below which is not helping
```
dotasek: I would like it if this comment referenced the actual code that processed the param:

```
//These two params are processed later by the RePackageTask and not included in ValidationContext.
```

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
```diff
@@ -79,12 +77,17 @@ public static class TerminologyResourceEntry {
...
+    String tmpdir = Files.createTempDirectory("repackage").toFile().getAbsolutePath();
+
     new PackageReGenerator()
-      .addPackage("hl7.fhir.us.davinci-alerts")
+      .addPackage("ch.fhir.ig.ch-core")
       .setJson(true)
+      .addIgnoreList(List.of("http://fhir.ch/ig/ch-core/StructureDefinition/ch-core-patient-epr", "http://fhir.ch/ig/ch-core/StructureDefinition/ch-core-claim", "urn:ietf:bcp:47"))
       .setOutputType(ExpansionPackageGeneratorOutputType.TGZ)
-      .setOutput("/Users/grahamegrieve/temp/vs-output.tgz")
+      .setOutput(tmpdir + "/some-output.tgz")
```
dotasek: I believe I misunderstood this in my last review; please use the `String[] args` object instead of tmpdir. We're already changing the expected behaviour for @grahamegrieve, but using `args` will ensure some consistency vs. a system dependent temp directory.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
```diff
@@ -79,12 +77,17 @@ public static class TerminologyResourceEntry {
   }
 
   public static void main(String[] args) throws Exception {
+    String tmpdir = Files.createTempDirectory("repackage").toFile().getAbsolutePath();
+
     new PackageReGenerator()
-      .addPackage("hl7.fhir.us.davinci-alerts")
+      .addPackage("ch.fhir.ig.ch-core")
```
dotasek: Maybe if this is changing from use to use, it should also be processed from `args`? 

PACKAGE_NAME in Params seems like a good candidate.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
```diff
@@ -79,12 +77,17 @@ public static class TerminologyResourceEntry {
...
       .setJson(true)
+      .addIgnoreList(List.of("http://fhir.ch/ig/ch-core/StructureDefinition/ch-core-patient-epr", "http://fhir.ch/ig/ch-core/StructureDefinition/ch-core-claim", "urn:ietf:bcp:47"))
```
dotasek: Same here, this is something we can get by processing `args`.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
jkiddo: honestly ... this is a test main function. I can just revert the changes on that method. I used this once and then went over to use the ordinary main method in org.hl7.fhir.validation.cli.ValidatorCli which works just as well.

jkiddo: Its now reverted. If it was my call I would delete it entirely.

jkiddo:
@dotasek FYI - Im not authorized to merge.

PR #2210, opened 2025-10-17
Feature/include list

jkiddo:
This feature adds to https://github.com/hapifhir/org.hl7.fhir.core/pull/2207 so that profiles that are referenced in e.g. FHIRpath expressions that are not identified traversing the regular hierarchy can be added using the parameter `-include-list` for the `-re-package`option


jkiddo:
@dotasek please have a look at let me know what you think

jkiddo:
Aka - you run it, you own it

grahamegrieve:
why would it be error prone?

jkiddo:
if you can write the fhirpath expression that does this automatically then be my guest with any kind of nested ways to do this. I don't see it conflicting with being able to have the user explicitly state certain resources that should be included.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
```diff
@@ -114,6 +114,12 @@ public PackageReGenerator addIgnoreList(List<String> ignoreList) {
     return this;
   }
 
+  public PackageReGenerator addIncludeList(List<CanonicalResource> includeList) {
```
dotasek: This is more accurately a `setIncludeList` method and should be named as such.

org.hl7.fhir.validation.cli/src/main/java/org/hl7/fhir/validation/cli/tasks/RePackageTask.java
```diff
@@ -93,4 +119,49 @@ private static ExpansionPackageGeneratorOutputType getExpansionPackageGeneratorO
     }
     return ExpansionPackageGeneratorOutputType.FOLDER;
   }
+
+  private List<CanonicalResource> loadResources(List<String> canonicals, FilesystemPackageCacheManager pcm) {
```
dotasek, line 124: Please check the `search` method of the PackageClient utility class. I think this method is manually constructing a query that's already handled there.

jkiddo: Fixed by now

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
jkiddo: Fixed

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
```diff
@@ -760,4 +795,35 @@ public PackageReGenerator setModes(Set<String> modeParams) {
     return this;
   }
 
+  private void processExpression(ExpressionNode node) {
+    if (node != null) {
+      if (node.getFunction() == ExpressionNode.Function.ConformsTo || node.getFunction() == ExpressionNode.Function.MemberOf) {
+        Base c = getConstantParam(node);
+        if (c != null) {
+          CanonicalResource dependency = (CanonicalResource) context.fetchResource(Resource.class, c.primitiveValue());
+          System.out.println(dependency.getTitle());
```
dotasek: This System.out needs to be replaced with an Slf4j info message. The class already has a logger, so it just needs to use it:

```java
 log.info(dependency.getTitle());
```

org.hl7.fhir.validation.cli/src/main/java/org/hl7/fhir/validation/cli/param/Params.java
```diff
@@ -619,8 +621,8 @@ else if (args[i].equals(HTML_OUTPUT)) {
             validationContext.setFhirpath(args[++i]);
         else
           throw new Exception("Can only nominate a single -fhirpath parameter");
-      } else if (args[i].equals(SCOPE) || args[i].equals(IGNORE_LIST)) {
-        //These two params are processed later by the RePackageTask and not included in ValidationContext.
+      } else if (args[i].equals(SCOPE) || args[i].equals(IGNORE_LIST) || args[i].equals(INCLUDE_LIST) || args[i].equals(INCLUDE_CONFORMS_TO)) {
+        //These three params are processed later by the RePackageTask and not included in ValidationContext.
```
dotasek: Four now... maybe just make it `These params`?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
```diff
@@ -204,7 +219,12 @@ public PackageReGenerator setNpmId(String npmId) {
   private Set<String> sourcePackages = new HashSet<>();
   private Map<String, TerminologyResourceEntry> entries = new HashMap<>();
   private Set<String> modeParams;
-  private List<CanonicalResource> resources = new ArrayList<CanonicalResource>();
+
+  public List<CanonicalResource> getResources() {
```
dotasek: Nitpick: could be a Lombok getter.

Just add `import lombok.Getter;` to the imports and annotate the `resources` field with `@Getter`. We should use these more to reduce clutter.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
jkiddo: Removed now ( was there for some debugging)

org.hl7.fhir.validation.cli/src/main/java/org/hl7/fhir/validation/cli/param/Params.java
jkiddo: Fixed

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
jkiddo: Done

grahamegrieve:
@jkiddo please resolve the conflicts, and also, please update the documentation at https://confluence.hl7.org/spaces/FHIR/pages/35718580/Using+the+FHIR+Validator#UsingtheFHIRValidator-PackageRegeneration

jkiddo:
@grahamegrieve conflicts are now resolved

PR #2211, opened 2025-10-19
2025 10 gg diagnostics

grahamegrieve:


org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/HTTPHeaderUtil.java
```diff
@@ -42,4 +42,8 @@ public static String getSingleHeader(Iterable<HTTPHeader> headers, String key) {
     }
     return null;
   }
+
+  public static boolean hasHeader(Iterable<HTTPHeader> headers, String s) {
+    return getSingleHeader(headers, s) != null;
```
dotasek, line 47: Any need to consider "" or other blank values?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/BaseValidator.java
dotasek, line 1: Maybe not in this PR, but I think we should think about a parameter object for `txHint`, `txRule`, etc. The method signatures are getting very large.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/HTTPHeaderUtil.java
grahamegrieve: if it's "" it has a value. so far as I understand the logic

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/BaseValidator.java
grahamegrieve: maybe, but I like the parameter validation. What I can do is stop passing some of those things around when I don't need to. A thing for the future

PR #2217, opened 2025-10-25
Allow relative path to assets

lmckenzi:


grahamegrieve:
this change will not be accepted

PR #2221, opened 2025-10-27
Use explicitly atomic file move in package cache locks

dotasek:
## DO NOT MERGE ON PASSING TESTS YET

_The pipelines should be re-run many times until we are satisfied that the change doesn't fail. Then, the two `@RepeatedTest(100)` annotations should be restored to `@Test` to ensure that these are restored to regular test status._

This uses nio Files.move with the StandardCopyOption.ATOMIC_MOVE to atomically delete lock files. This may provide some improvements for intermittently failing FilesystemPackageManagerTests on some operating systems.


grahamegrieve (approved):
ok

PR #2225, opened 2025-10-29
Strip outdated translations

vadi2:
There are two systems in place for dealing with outdated translations - both in Java code and Crowdin, and they are clashing.

Java code prefixes `!!` to outdated translations, which is a new development compared to other translation systems. This works well if you're translating the `.po` file by hand. If you import it into Crowdin though, that causes issues as it has it's own mechanisms for tracking updated source strings, matching them intelligently, and notifying translators of what actually needs translating.

The proposal here will leave the `!!` system as it is unless we need to interact with Crowdin, where we'll remove the outdated translations and let Crowdin's Translation Memory and other tools do their job.

### DIFF
--- a/.github/workflows/crowdin.yml
+++ b/.github/workflows/crowdin.yml
@@ -27,6 +27,33 @@ jobs:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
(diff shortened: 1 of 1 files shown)


dotasek (changes requested):
See earlier comments for details. The generated pull request is adding new comments/headers, and deleting existing translations.

PR #2230, opened 2025-10-30
R5 & R6 TurtleParser, ShExGenerator cross-version redirects 

tmprd:
Because `org.hl7.fhir.r5.elementmodel.TurtleParser` is used in cross-version contexts for R4, R5, and R6, it needs to handle each version according to the FHIR RDF spec for that version.

### Revert `org.hl7.fhir.r5.elementmodel.TurtleParser` to R5 only
This PR reverts recent changes intended for R6 that were added to `org.hl7.fhir.r5.elementmodel.TurtleParser`, based on the [commit history](https://github.com/hapifhir/org.hl7.fhir.core/commits/master/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/TurtleParser.java) and review with the [RDF working group](https://github.com/w3c/hcls-fhir-rdf/)
  * TurtleParser was reverted to the state as of [refactor xhtml handling to fix canonical issues](https://github.com/hapifhir/org.hl7.fhir.core/commit/99cec1c50bac6acee432464515bf7b931dffd859) June 24th, 2025 
  * Changes after that commit were removed, which were originally intended for R6
    * Beginning with ["Generate fhir:link triples for URI values."](https://github.com/hapifhir/org.hl7.fhir.core/commit/55b281908c0f13542fbd24bfd7e184a765a63e55)
    * and incuding ["more work on additional resources"](https://github.com/hapifhir/org.hl7.fhir.core/commit/92833603e7fb963b7b6205c2847fbaf183187019) 
    * Commits before then, going back to the R5 release (~2023-03-26), were intended for R5 (bug fixes)

### Redirect to TurtleParser and ShexGenerator R6 
This PR also adds a redirection to the R6 module `org.hl7.fhir.r5.elementmodel.TurtleParserR6` which can be used in R6 contexts. 
  * This module can moved in the future depending on how the R6+ namespaces are structured.
  * The corresponding R4 module in `org.hl7.fhir.r4.elementmodel` (or `org.hl7.fhir.r4.formats.RdfParser`) could also be added for redirection, but would require adding it as a depenency of `org.hl7.fhir.r5.elementmodel`.

R6 redirection for `org.hl7.fhir.r5.conformance.ShexGenerator` is similarly added, due to ShEx's dependency on the structure of TTL. 

### Add unit tests
This PR adds test XML files from R5 and R6, and their corresponding expected TTL files:
* https://hl7.org/fhir/R5/patient-example-f201-roel.xml.html
* https://build.fhir.org/patient-example-f201-roel.xml.html 
The TTL unit tests generate TTL from those example, and check whether the generated TTL matches the expected TTL.

For ShEx testing, the XML files for R5 and R6 are added:
* https://hl7.org/fhir/R5/patient.profile.xml.html
* https://build.fhir.org/patient.profile.xml.html
The ShEx unit tests generate ShEx from those profiles, and check whether the generated ShEx matches the expected ShEx.


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/TurtleParser.java
```diff
@@ -84,11 +84,24 @@ public class TurtleParser extends ParserBase {
...
     super(context);
+
+    this.r6Parser = new org.hl7.fhir.r5.elementmodel.TurtleParserR6(context);
   }
   @Override
   public List<ValidatedFragment> parse(InputStream inStream) throws IOException, FHIRException {
+    // Redirect cross-version parsing
+    String fhirVersion = context.getVersion();
+    if ( VersionUtilities.isR4Ver(fhirVersion) ) {
+        throw new FHIRException("Turtle parsing for R4 is not supported in this build. Use the R4 module.");
```
tmprd: @grahamegrieve I'm not sure how to redirect to R4 parsing/serialization without making the `org.hl7.fhir.r5` project depend on `org.hl7.fhir.r4`. Would throwing an exception here be ok for now? Or return an empy list?

dotasek (changes requested):
@tmprd how can we verify that this parser works as intended?

A base case of a ttl file and an expected resource should be fairly simple to implement.

If there are existing tests that use ttl, and this is just a refactor, please point them out and I will verify that the tests cover this code.

tmprd:
> @tmprd how can we verify that this parser works as intended?
> 
> A base case of a ttl file and an expected resource should be fairly simple to implement.
> 
> If there are existing tests that use ttl, and this is just a refactor, please point them out and I will verify that the tests cover this code.

Hi @dotasek , I added some tests for R5 and R6:
* These generate a TTL file from an XML file, and compare it with the expected TTL file.
* This is more relevant here than [TurtleTests.java](https://github.com/hapifhir/org.hl7.fhir.core/blob/master/org.hl7.fhir.utilities/src/test/java/org/hl7/fhir/utilities/turtle/TurtleTests.java), which tests parsing TTL, not generating it.
* The R6 tests and TurtleParserR6 can be moved to an R6 module when available. 

Do you know why SnapShotGenerationTests would be failing [here](https://github.com/hapifhir/org.hl7.fhir.core/blob/54b40f832c1bee5d40e9be04a9066d56a6924664/org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/SnapShotGenerationTests.java#L668) ? I can't see how my changes could be related. These pass for me on my machine, maybe due to a difference in packages.

Test output: https://dev.azure.com/fhir-pipelines/fhir-core-library/_build/results?buildId=29540&view=logs&j=e024f35c-af72-57d6-fcdb-0b5ae73d34cf&t=2f33f0d6-763f-5d62-8c02-969bc9d72807

- If you search for "snapshot test" in that log, it shows a difference in expected snapshots. 
- I formatted and compared some of these, and the only difference is that all the `targetProfile` canonical URLs have a version attached to the end
  - For example: Expected (t1) "http://hl7.org/fhir/StructureDefinition/Practitioner" vs Actual (t2) "http://hl7.org/fhir/StructureDefinition/Practitioner|5.0.0"

dbooth-boston:
> This PR also adds a redirection to the R6 module org.hl7.fhir.r5.elementmodel.TurtleParserR6 which can be used in R6 contexts.

Forgive me if this was discussed before, but it seems confusing that "r5" is in the name of the R6 parser.  Why is it organized that way?

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/TurtleGeneratorTestUtils.java
```diff
@@ -0,0 +1,266 @@
...
+        new Turtle().parse(turtle);
+        return normalizeTurtle(turtle);
+      }
+    }
+    
+    private static String normalizeTurtle(String turtle) {
+      return turtle.replaceAll("\\r\\n?", "\n").trim();
+    }
+
+    public Element parseGeneratedTurtleElement(String turtleFilePath) throws IOException {
```
dotasek, line 84: This method appears to be unused, and can be removed.

dotasek (changes requested):
@tmprd the tests look appropriately constructed. Thank you for adding them.

Two issues with this:

- I need someone who can review the expected ttl files. If there's anyone from a working group you can suggest, it would be appreciated.
- TurtleParserR6 and TurtleParser share a LOT of code. In previous iterations of the FHIR spec, we regenerated or copied code, but since R6 will be using R5 models and parsing moving forward, I think this copy-paste has a lot of potential to cause trouble (someone changing the original file might not know to change R6 and vice versa). I'd like to see some more core re-use, perhaps an abstract TurtleParser and a TurtleParserR5 and TurtleParserR6 that can insert the appropriate logic for each FHIR version.

dotasek:
@tmprd the failing SnapShotGenerationTests should resolve after a merge from master (they did for me).

tmprd:
> > This PR also adds a redirection to the R6 module org.hl7.fhir.r5.elementmodel.TurtleParserR6 which can be used in R6 contexts.
> 
> Forgive me if this was discussed before, but it seems confusing that "r5" is in the name of the R6 parser. Why is it organized that way?

There's no R6 module available, and Kindling is using this module for publishing the R6 FHIR spec (example TTL files). It sounds like the IG publisher will do the same. We can move this code to an R6 module in the future if needed.

> @tmprd the tests look appropriately constructed. Thank you for adding them.
> 
> Two issues with this:
> 
> * I need someone who can review the expected ttl files. If there's anyone from a working group you can suggest, it would be appreciated.
> * TurtleParserR6 and TurtleParser share a LOT of code. In previous iterations of the FHIR spec, we regenerated or copied code, but since R6 will be using R5 models and parsing moving forward, I think this copy-paste has a lot of potential to cause trouble (someone changing the original file might not know to change R6 and vice versa). I'd like to see some more core re-use, perhaps an abstract TurtleParser and a TurtleParserR5 and TurtleParserR6 that can insert the appropriate logic for each FHIR version.

Thank you for the feedback. The RDF working group will review the expected TTL files. 
I'll refactor to share code between the versions as you suggested. I wasn't sure if R5 would continue to be modified, but I guess it's always possible if a bug needs fixing.

tmprd:
@dotasek I refactored these into a shared TurtleParserBase class like you suggested. Now TurtleParser and TurtleParserR6 are much smaller https://github.com/hapifhir/org.hl7.fhir.core/pull/2230/changes#diff-bd407421afc5895cb2936314b9d06f0ea933f533d39e113780f9c3bb48625a53

I also remembered that ShExGenerator depends on the structure of each versioned TTL. So I introduced a new ShexGeneratorR6 with the same pattern of redirection, depending on the FHIR version in the "worker context". 
This also gave me an opportunity to add better unit tests for ShEx, so I added those. The working group will confirm all unit test example files for both TTL and ShEx are correct.

I would like ShExGenerator to similarly have a shared base class between versions, however I know @ericprud has some pending changes to ShExGenerator, so I would prefer to do that after this PR to avoid further complication. 

> the failing SnapShotGenerationTests should resolve after a merge from master (they did for me).
I fixed this but now see a new test failing: `ExternalTerminologyServiceTests.test:230 string property values differ at .parameter[1].resource.issue[0].details.text` https://dev.azure.com/fhir-pipelines/fhir-core-library/_build/results?buildId=29855&view=logs&j=a653de61-ab9c-5ebd-a143-e04ec0557a8f&t=f7d9eba4-2c95-54aa-40c6-3628ad042ae0&l=16423

ericprud (approved):
I haven't executed the ShEx, but the target R5 and R6 ttl look good to me.
Tim, can you easily check the following:? 
1. ShEx generator emits files `code.shex` and `Code.shex` in the R5 and R6 output respectively?
2. Al of `code`, `date`, `boolean`, `integer`, and `dateTime` are capitolized in R6?

I see new Patient.contact.{role,additional{Name,address}} in org.hl7.fhir.r5/src/test/resources/testUtilities/xml/examples/expected/R6/patient.profile.xml vs R5 so I believe the R5 an R6 ShEx and examples are aligned correctly.

dbooth-boston:
I also looked at the Turtle, and it looked okay to me.

dbooth-boston (commented):
The turtle looks okay to me.

tmprd:
> 1. ShEx generator emits files `code.shex` and `Code.shex` in the R5 and R6 output respectively?

Yes the expected "patient.shex" examples import those, so they're required to be generated to pass the unit test: https://github.com/hapifhir/org.hl7.fhir.core/pull/2230/changes#diff-a63be92d5918323584d1c62edfdb411cccb0cc3399eb75f2df9b628243824688

> 2. Al of `code`, `date`, `boolean`, `integer`, and `dateTime` are capitolized in R6?

Those are all used in "patient.shex" in both versions, so the test has to generate them with the same casing in order to match the string comparison.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/ShExGenerator.java
```diff
@@ -551,6 +591,15 @@ public String generate(HTMLLinkPolicy links, List<StructureDefinition> structure
     return shex_def.render();
   }
 
+  private void syncR6GeneratorState() {
```
tmprd: This is needed because of how ShExGenerator is called and then mutated in Kindling https://github.com/HL7/kindling/blob/main/src/main/java/org/hl7/fhir/tools/publisher/Publisher.java#L2943-L2946

dotasek: We should really refactor ShExGenerator to have these fields final and set via a builder.

That's a legacy issue though. I will make an issue for refactoring this.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/TurtleParser.java
```diff
@@ -36,628 +34,57 @@ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWIS
...
-    if (property.isList() && (e instanceof TTLList)) {
-      TTLList arr = (TTLList) e;
-      for (TTLObject am : arr.getList()) {
-        parseChildInstance(errors, src, npath, object, context, property, name, am);
-      }
-    } else {
-      parseChildInstance(errors, src, npath, object, context, property, name, e);
+    // Redirect cross-version parsing
+    String fhirVersion = context.getVersion();
+    if ( VersionUtilities.isR4Ver(fhirVersion) ) {
```
dotasek: I think the logic here should be:

```java
if ( VersionUtilities.isR6Ver(fhirVersion) ) {
  return r6Parser.parse(inStream); 
}

if (VersionUtilities.isR5Ver(fhirVersion)) {
   return super.parse(inStream);
}

 throw new FHIRException("Turtle parsing for versions under R5 is not supported in this module. Use the appropriate module. (R4, DSTU3, etc)");
```

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/ShExGeneratorR6.java
dotasek, line 1: This is a near copy of a lot of ShExGenerator. Can the also be refactored to share the same code?

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/ShexGeneratorTests.java
```diff
@@ -32,131 +41,308 @@
...
+
+  // ---------------------------------------------------------------------------
+  // Test helpers
+  // ---------------------------------------------------------------------------
+
+  private void doTestR5(String name, ShexGeneratorTestUtils.RESOURCE_CATEGORY cat) throws FileNotFoundException, IOException, FHIRException, UcumException {
+    doTest(name, cat, r5WorkerContext, expectedShexDirectory.resolve("R5"));
+  }
+
+  private void doTestR6(String name, ShexGeneratorTestUtils.RESOURCE_CATEGORY cat) throws FileNotFoundException, IOException, FHIRException, UcumException {
```
dotasek: cat is unused here. It should be deleted for cleaner code.

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/ShexGeneratorTests.java
```diff
@@ -32,131 +41,308 @@
...
+    doTest(name, cat, r6WorkerContext, expectedShexDirectory.resolve("R6"));
+  }
+
+  private void doTest(String name, ShexGeneratorTestUtils.RESOURCE_CATEGORY cat, IWorkerContext context, Path expectedDirectory) throws FileNotFoundException, IOException, FHIRException, UcumException {
```
dotasek: cat and expectedDirectory are unused here, and should be cleaned up.

dotasek: This actually contains no assertion. assertDoesNotThrow should at least be used to around the method under test if it's just intended to run without exception.

dotasek (changes requested):

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/ShexGeneratorTests.java
```diff
@@ -32,131 +43,314 @@
...
+  private void doTestR5(String name) throws FileNotFoundException, IOException, FHIRException, UcumException {
+    doTest(name, r5WorkerContext);
+  }
+
+  private void doTestR6(String name) throws FileNotFoundException, IOException, FHIRException, UcumException {
+    doTest(name, r6WorkerContext);
+  }
+
+  private void doTest(String name, IWorkerContext context) throws FileNotFoundException, IOException, FHIRException, UcumException {
+    assertDoesNotThrow(() -> {
```
tmprd, line 273: Added back assertDoesNotThrow here ✅

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/ShexGeneratorTests.java
```diff
@@ -32,131 +43,314 @@
...
 
-    // For Testing Schema Processing and Constraint Mapping related Development
-    // If you un-comment the following lines, please comment all other lines in this method.
-    // Test with processing constraints flag
-    // ----------------- Uncomment following to testing/Debugging -----
-    //    boolean processConstraints = false;
-    //    this.doTestSingleSD(name.toLowerCase(), cat, name,
-    //      false, ShExGenerator.ConstraintTranslationPolicy.ALL,
-    //      true, true, false, processConstraints);
+    assertCompleteModelR6MatchesDirectGenerator(generator, direct);
```
tmprd, line 83: This test ensures that the now deprecated way of using ShExGenerator by individually mutating its settings still works. ✅

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/ShExGenerator.java
```diff
@@ -3,1889 +3,136 @@
...
-
-  // List of URLs of selected Structure Definitions of Extensions from ShEx Schema generation.
-  // Extensions are Structure Definitions with type as "Extension".
-  private List<StructureDefinition> selectedExtensions;
-  private List<String> selectedExtensionUrls;
-
-  private List<String> imports;
-  private FHIRPathEngine fpe;
-
+  @Deprecated
```
tmprd, line 53: Deprecated constructing ShExGenerator without a config. This is because mutating ShExGenerator's global variable settings is also deprecated now. ✅

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/ShExGenerator.java
```diff
@@ -3,1889 +3,136 @@
...
-    element_def.add("card", card);
-    addComment(element_def, ed);
-
-    return element_def.render();
+    this(context, ShExGeneratorConfig.defaultConfig());
   }
 
-  private boolean changeShortName(StructureDefinition sd, ElementDefinition ed){
-      return shortIdException.contains(sd.getName());
+  public ShExGenerator(IWorkerContext context, ShExGeneratorConfig config) {
```
tmprd, line 58: New better way of constructing ShExGenerator which makes cross-version redirection easier ✅

org.hl7.fhir.r5/src/test/resources/testUtilities/xml/examples/codesystem-contact-point-use.xml
tmprd, line 1: This was added for previous TTL tests and is no longer used

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/ShExGenerator.java
tmprd: I added new code for that here. You can build ShExGenerator with a config instead, and the old way is deprecated.

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/TurtleGeneratorTests.java
```diff
@@ -1,199 +1,185 @@
...
  */
 public class TurtleGeneratorTests {
+  private static TurtleGeneratorTestUtils.ParserContext parsers;
 
-  private static IWorkerContext workerContext;
-  private static ResourceParser resourceParser;
-  private static XmlParser xmlParser;
-  private static TurtleParser turtleParser;
-
+  private static final Path ROOT_TEST_PATH = Paths.get("testUtilities");
```
dotasek, line 45: This test path is a little misleading, as it implies that it contains utilities .

I would prefer if these tests were just moved up one level, so that the structure was:

```
shex/expected
ttl/examples/expected
xml/examples
```

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/ShexGeneratorTests.java
dotasek, line 1: A side-effect of these changes is that ShexGeneratorTestUtils is no longer used. It should be deleted.

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/ShexGeneratorTests.java
```diff
@@ -32,131 +43,314 @@
...
+
+  private Path generateShex(String outputName, StructureDefinition sd, IWorkerContext context) throws IOException, UcumException {
+    Path outPath = FileSystems.getDefault().getPath(System.getProperty("java.io.tmpdir"), outputName + ".shex");
+    System.out.println("Generated ShEx to " + outPath.toString());
+    FileUtilities.stringToFile(new ShExGenerator(context).generate(ShExGeneratorBase.HTMLLinkPolicy.NONE, sd), outPath.toString());
+    return outPath;
+  }
+
+  private void compareShex(Path actualShexPath, Path expectedShexPath) throws IOException {
+    if (Files.exists(expectedShexPath)) {
```
dotasek, line 311: If the expected file does not exist, this will always succeed. It would be good to require it explicitly exist via:

```java
} else { 
      fail("File expected for comparison does not exist: " + expectedShexPath); 
    }
```

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/TurtleParser.java
```diff
@@ -36,628 +34,67 @@ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWIS
...
-  private String base;
+public class TurtleParser extends TurtleParserBase {
 
-  private OutputStyle style;
-
-  public static String FHIR_URI_BASE = "http://hl7.org/fhir/";
-  public static String FHIR_VERSION_BASE = "http://build.fhir.org/";
-  public static String FHIR_BASE_PREFIX = "fhir:";
+  // Cross-version usage
+  private TurtleParserR6 r6Parser;
```
dotasek: Should this also be made thread-safe the same way as in ShExGenerator?

dotasek (changes requested):

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/ShexGeneratorTests.java
tmprd: Ah right, that's from the unused parameter that I also removed. Both are being used in @ericprud 's branch we will PR after this. Anyway I'll delete it here for now.

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/TurtleGeneratorTests.java
tmprd: I'm not sure I understand. "testUtilities" already exists in master https://github.com/hapifhir/org.hl7.fhir.core/tree/master/org.hl7.fhir.r5/src/test/resources/testUtilities I don't see it explicitly referenced in other code, but it looks like it's used for other JSON and XML tests.

What's the full path you'd like to use?
Current: `org.hl7.fhir.r5/src/test/resources/testUtilities/ttl/examples/expected/R6/patient-example-f201-roel.ttl`

Change to this?: `org.hl7.fhir.r5/src/test/resources/ttl/examples/expected/R6/patient-example-f201-roel.ttl`

dotasek: Woops, you are right. That's our fault, not yours. Ignore the suggestion, I'll put refactoring that on my TODO list.

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/ShexGeneratorTests.java
dotasek: If I have you assurance that it will be re-used in another PR, that's fine. It will likely make the review for that one easier.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/TurtleParser.java
tmprd: ShExGenerator does this in getR6Generator() is because it can be constructed with settings that can later be mutated. That makes redirecting to the R6 version more complicated because the current settings have to be enumerated and copied. And so we require that only 1 thread at a time can do this, at least. But all of that is only needed for the existing constructor, which I deprecated. It's not needed for the new constructor which requires all the settings up front.

Looking into it more, I see TurtleParser currently inherits from [ParserBase](https://github.com/hapifhir/org.hl7.fhir.core/blob/master/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/formats/ParserBase.java) (used elsewhere), which does have some mutable settings fields. So to REALLY be safe here, I'll add the same for TurtleParser.

dotasek (approved):
Approved. Will merge either this release or the next, and pending pipeline checks.

PR #2235, opened 2025-11-05
Use param objects in CliTasks

dotasek:
Uses the new parameter objects and parameter parsing to replace ValidatorContext usage. 

This removes any usages of ValidatorContext except in places where backward compatibility is required. Usages of ValidatorContext in ValidatorService are deprecated.

Note that the new parameter objects in `org.hl7.fhir.validation.service.model` are marked as `@Deprecated`, as they are intended to be replaced by a much less verbose CLI system like Picocli. The only models that are not deprecated are those used for ValidationRequest, which are annotated for XML/JSON serialization and used by web projects such as validator-wrapper and Inferno.


grahamegrieve (approved):
ok

PR #2238, opened 2025-11-13
Ensure CI build server request time limits are always updated

dotasek:
This ensures that regardless of outcome, the last queried time for the CI server is always updated to ensure that queries are limited to once an hour by default.


grahamegrieve (approved):
ok

PR #2239, opened 2025-11-17
2025 11 gg tho

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/ContextUtilities.java
```diff
@@ -563,5 +563,15 @@ public String getDefinitionsName(Resource r) {
     return null;
   }
 
+  public NamingSystem fetchNamingSystem(String url) {
+    for (NamingSystem ns : context.fetchResourcesByType(NamingSystem.class)) {
+      for (NamingSystemUniqueIdComponent id : ns.getUniqueId()) {
+        if (url.equals(id.getValue())) {
```
dotasek, line 572: Add null check.

grahamegrieve: what could be null?

grahamegrieve: well, I checked for url being null

PR #2242, opened 2025-11-25
2025 11 gg lang regen

grahamegrieve:


dotasek:
Since some typos were fixed, this test cases PR might be required for tests to pass: https://github.com/FHIR/fhir-test-cases/pull/248

dotasek:
The additional changes added after @grahamegrieve's initial PR are to keep the language generation stable. 

The PO files inherited a series of `\` escapes that corrupted the format, as well as gradually increasing `!!` prefixes. As language regeneration was called, both of these would result in a worsening state.

The two changes that raise a few eyebrows but seem correct are to the following in Messages.properties:

HTA_LOINC_MESSAGE
HTA_SCT_MESSAGE

I removed escape characters there, as they don't appear to make sense. In the sample ig, these render fine without escapes:

Screenshot 2025-11-27 at 10 57 17 AM

What's odd is that these two were specifically involved in this PR, but have apparently rendered fine with the `\` escapes in place:

https://github.com/hapifhir/org.hl7.fhir.core/pull/2076

If these render fine as they appear above, I don't see a reason for PO files to have anything but a single escape of the format `\"` as in this fix, and for `.properties` files to  simply use `"` as is.

dotasek:
Approved by @grahamegrieve via a zulip message.

dotasek (approved):
Pipelines are passing here:

https://dev.azure.com/fhir-pipelines/fhir-core-library/_build/results?buildId=27679&view=results

PR #2202, opened 2025-10-09
Identifier system check

lmckenzi:
As per https://github.com/HL7/fhir-ig-publisher/issues/912, support cross-checking Identifier.system values against known identifier NamingSystems

PR #2244, opened 2025-11-26
Identifier system check2

lmckenzi:
As per https://github.com/HL7/fhir-ig-publisher/issues/912, support cross-checking Identifier.system values against known identifier NamingSystems

Replaces https://github.com/hapifhir/org.hl7.fhir.core/pull/2202 and makes changes requested by @grahamegrieve


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/ContextUtilities.java
```diff
@@ -576,5 +579,50 @@ public NamingSystem fetchNamingSystem(String url) {
     }
     return null;
   }
+  
+  public String preferredIdentifierSystem(String system) {
+    if (identifierSystemPreferredUris == null) {
+      identifierSystemPreferredUris = new HashMap<>();
+      // Yes, getNSUrlMap is deprecated, however I'm not supposed to change IWorkerContext, and
+      // context doesn't provide any other way of accessing the collection of loaded NamingSystems
```
grahamegrieve, line 587: well, you'd do fetchResourcesByType(NamingSystem.class)

org.hl7.fhir.utilities/src/main/resources/Messages.properties
```diff
@@ -986,6 +985,8 @@ TYPE_SPECIFIC_CHECKS_DT_IDENTIFIER_IETF_SYSTEM_VALUE = if identifier.system is '
 TYPE_SPECIFIC_CHECKS_DT_IDENTIFIER_IETF_SYSTEM_WRONG_1 = The URL ''https://tools.ietf.org/html/rfc4122'' is not correct for the identifier ''{0}''
 TYPE_SPECIFIC_CHECKS_DT_IDENTIFIER_IETF_SYSTEM_WRONG_2 = The URL ''https://tools.ietf.org/html/rfc4122'' should not be used with the UUID ''{0}''. Prefix the value with ''urn:uuid:'' and use the system value ''urn:ietf:rfc:3986'' instead
 TYPE_SPECIFIC_CHECKS_DT_IDENTIFIER_IETF_SYSTEM_WRONG_3 = The URL ''https://tools.ietf.org/html/rfc4122'' should not be used with the UUID ''{0}''. Use the system value ''urn:ietf:rfc:3986'' instead
+Type_Specific_Checks_DT_Identifier_System_Unknown = Identifier.system is not based on example.org or example.com and is not a preferred URI of any identifier system NamingSystems.  Ensure that the Identifier.system value is a properly defined system value
```
grahamegrieve, line 988: I think that you shouldn't say 'Ensure...' as an instruction for a hint. It's a bit strong

org.hl7.fhir.utilities/src/main/resources/Messages.properties
```diff
@@ -986,6 +985,8 @@ TYPE_SPECIFIC_CHECKS_DT_IDENTIFIER_IETF_SYSTEM_VALUE = if identifier.system is '
...
+Type_Specific_Checks_DT_Identifier_System_Not_Preferred = Identifier.system is set to ''{0}'' when the preferred URI for that identifier system is ''{1}''.  As a result, other implementations may not recognize the identifier system.
```
grahamegrieve, line 989: This one feels like it should be a warning not a hint

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
```diff
@@ -3070,6 +3078,14 @@ private boolean checkIdentifier(List<ValidationMessage> errors, String path, Ele
           ok = rule(errors, NO_RULE_DATE, IssueType.CODEINVALID, element.line(), element.col(), path, false, I18nConstants.TYPE_SPECIFIC_CHECKS_DT_IDENTIFIER_IETF_SYSTEM_WRONG_1, value) && ok;
         }
       }
+    } else if (isStrictIdentifierSystems() && system != null) {
```
grahamegrieve, line 3081: I'm not sure I follow the StrictIdentifierSystems logic part here - I would've thought that by default we should check these things, and StrictIdentifierSystems would increase the level of the issues?

Also, we should only allow example identifier systems if examples are allowed?

lmckenzi:
We're not checking identifier systems at all right now.  I *wouldn't* expect them to be checked in production systems because there will be all kinds of identifiers flying around and most systems won't recognize the majority of them.  That's normal and ok.  However, when we're publishing examples in specs, we (generally) don't want unrecognized systems that aren't clearly distinguished as examples.

grahamegrieve:
10000 commits?

PR #2248, opened 2025-12-02
2025 12 gg africa misc

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
```diff
@@ -858,7 +858,7 @@ public SystemSupportInformation getTxSupportInfo(String system, String version)
     synchronized (lock) {
       String vurl = CanonicalType.urlWithVersion(system, version);
       if (codeSystems.has(vurl) && codeSystems.get(vurl).getContent() != CodeSystemContentMode.NOTPRESENT) {
-        return new SystemSupportInformation(true, "internal", TerminologyClientContext.LATEST_VERSION);
+        return new SystemSupportInformation(true, "internal", TerminologyClientContext.LATEST_VERSION, "Content is not present");
```
dotasek: Doesn't  "Content is not present" directly contradict `codeSystems.get(vurl).getContent() != CodeSystemContentMode.NOTPRESENT`?

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/i18n/RenderingI18nContext.java
```diff
@@ -1122,6 +1122,7 @@ public class RenderingI18nContext extends I18nBase {
   public static final String VS_VERSION_NOTHING_TEXT = "VS_VERSION_NOTHING_TEXT";
   public static final String CS_VERSION_NOTHING_TEXT = "CS_VERSION_NOTHING_TEXT";
   public static final String STRUC_DEF_IMPLEMENTS = "STRUC_DEF_IMPLEMENTS";
+  public static final String PROV_REFERENCE = "PROV_REFERENCE";
```
dotasek, line 1125: Can't find a usage for this?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/expansion/PropertyFilter.java
```diff
@@ -28,7 +29,7 @@ public PropertyFilter(List<String> allErrors, ConceptSetFilterComponent fc, Prop
   public boolean includeConcept(CodeSystem cs, ConceptDefinitionComponent def) {
     ConceptPropertyComponent pc = getPropertyForConcept(def);
     if (pc != null) {
-      String v = pc.getValue().isPrimitive() ? pc.getValue().primitiveValue() : null;
+      String v = pc.getValue().isPrimitive() ? pc.getValue().primitiveValue() : pc.getValue() instanceof Coding ? ((Coding) pc.getValue()).getSystem()+"#"+((Coding) pc.getValue()).getCode() : null;
```
dotasek: The nesting here is getting a little out of hand for a single line.

Maybe this:

```java
private static String getStringValueFromProperty(ConceptPropertyComponent propertyForConcept) {
    if (propertyForConcept.getValue().isPrimitive()) {
      return propertyForConcept.getValue().primitiveValue();
    }
    if (propertyForConcept.getValue() instanceof Coding coding) {
      return coding.getSystem() + "#" + coding.getCode();
    }
    return null;
  }
```

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/utilities/ValidationResult.java
```diff
@@ -67,7 +67,23 @@ public ValidationResult(IssueSeverity severity, String message, List<OperationOu
...
-        if (issue.getSeverity() == OperationOutcome.IssueSeverity.ERROR || message == null) {
+        if (issue.getSeverity() == OperationOutcome.IssueSeverity.ERROR) {
+          String msg = issue.getDetails().getText();
+          if (!this.messages.contains(msg)) {
+            this.messages.add(msg);
+          }
+        }
+      }
+      for (OperationOutcomeIssueComponent issue : issues) {
+        if (issue.getSeverity() == OperationOutcome.IssueSeverity.WARNING && messages.isEmpty()) {
```
dotasek, line 78: Why the `isEmpty()` check here and below? 

I assume this is so that if there are errors, there are only errors in messages, and if no errors, then only warnings, and finally only adding info if there's nothing else of greater severity.

For testing purposes, I assume we should only be investigating issue lists then, and treating messages as a courtesy?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/validation/ValueSetValidator.java
```diff
@@ -1890,15 +1892,25 @@ private boolean codeInPropertyFilter(CodeSystem cs, ConceptSetFilterComponent f,
         return false;
       }
       DataType d = CodeSystemUtilities.getProperty(cs, code, f.getProperty());
-      return d != null && f.getValue().equals(d.primitiveValue());
+      if (d instanceof Coding) {
```
dotasek, line 1896: Nitpick: This equality seems similar to the suggested method in my other comment. Maybe it makes sense for a utility method?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/validation/ValueSetValidator.java
```diff
@@ -1890,15 +1892,25 @@ private boolean codeInPropertyFilter(CodeSystem cs, ConceptSetFilterComponent f,
...
+      }
     case EXISTS: 
       return CodeSystemUtilities.getProperty(cs, code, f.getProperty()) != null;
     case REGEX:
       if (f.getValue() == null) {
         return false;
       }
       d = CodeSystemUtilities.getProperty(cs, code, f.getProperty());
-      return d != null && d.primitiveValue() != null && d.primitiveValue().matches(f.getValue());
+      if (d instanceof Coding) {
```
dotasek, line 1908: Same nitpick here; repeated code regarding value.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/i18n/I18nConstants.java
```diff
@@ -1079,6 +1079,7 @@ public class I18nConstants {
   public static final String UNKNOWN_CODE_IN = "Unknown_Code_in";
   public static final String UNKNOWN_CODE_IN_FRAGMENT = "UNKNOWN_CODE_IN_FRAGMENT";
   public static final String UNKNOWN_CODE_IN_VERSION = "Unknown_Code_in_Version";
+  public static final String UNKNOWN_CODE_IN_VERSION_SCT = "Unknown_Code_in_Version_SCT";
```
dotasek, line 1082: Not used?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ValidatorSettings.java
```diff
@@ -131,4 +132,18 @@ public boolean isUseNewXVersionPackages() {
   public void setUseNewXVersionPackages(boolean useNewXVersionPackages) {
     this.useNewXVersionPackages = useNewXVersionPackages;
   }
+
+  private static Set<String> buildJadesHeaders() {
+    Set<String> stringSet = new HashSet<>();
+    // see https://github.com/hapifhir/org.hl7.fhir.core/issues/2209
+    stringSet.add("srCms");
```
dotasek, line 139: My understanding of this is from this [API doc](https://javadoc.io/doc/com.nimbusds/nimbus-jose-jwt/4.16/com/nimbusds/jose/crypto/RSASSAVerifier.html#RSASSAVerifier-java.security.interfaces.RSAPublicKey-java.util.Set-)

We're processing these headers ourselves in various places in the verifier, like `iat`, so that makes sense.

But, until I searched for that, I had no idea. And we don't manage "version" internally, or make reference to it at all.

Is it possible to pull this out into a class or util that handles the crit headers so it's clearer?

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/i18n/RenderingI18nContext.java
grahamegrieve: it's in the IG publisher

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
grahamegrieve: you're right

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/utilities/ValidationResult.java
grahamegrieve: no we care about both - the message is a summary of the errors if there are errors. Else warnings. Else hints. This code was disagreeing with the rest of the code, and leading to irregular message outcomes. fixed now

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/i18n/I18nConstants.java
grahamegrieve: Used in the terminology server

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ValidatorSettings.java
grahamegrieve: it's already in one place, here. How would making it its own class make it clearer?

dotasek: Maybe a class is overkill, but maybe constants?

For example, I have to search for usages of each of these keys in the codebase to see where they are actually produced in our signing, and then I have to check if they're actually used in signing, and not as some other key. 

But then, I find things like "sigD", which we do not use, and "sigT", which we do. So keeping things in constants would at least make finding these things a one-click process in an IDE, and less prone to typos, etc.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/expansion/PropertyFilter.java
grahamegrieve: ok fixed

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/validation/ValueSetValidator.java
grahamegrieve: ok fixed

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ValidatorSettings.java
grahamegrieve: ok I'll do that in the future

PR #2250, opened 2025-12-04
Necessary to ensure that simple extensions render correctly.  See the…

lmckenzi:
… fragment here: https://build.fhir.org/ig/HL7/davinci-crd/en/deviations.html#if-none-exist for an example


grahamegrieve (approved):
ok

PR #2254, opened 2025-12-08
Use Picocli for CLI

dotasek:
This will switch all of our CLI param/argument parsing to the Picocli library.

All existing CLI functionality will be preserved, but there is now a clearer distinction between parameters, options, and commands, and the CLI will warn users that some of their calls are deprecated. 

For example, the `snapshot` command is expected to be called like so:

```sh
java -jar validator.jar snapshot ./StructureDefinition-mypatient.json -outputSuffix .json ...
```

Users previously used `-snapshot`. This will still work, as there is code that translates the deprecated string to the new one, but there will be a warning printed in the output:

```
The option '-snapshot' is no longer in use. Using 'snapshot' 
```


dotasek:
@grahamegrieve 

I have only one question after a last minute fix I did: what does the following do?

```
catch (ENoDump e) {
      log.info(e.getMessage());
    }
```

I feel like that probably should go in the ValidateCommand instead, but thought I'd ask.

dotasek:
ENoDump is removed (upon inspection it was never thrown anyway, so changes no functionality). Will merge once pipelines pass.

PR #2273, opened 2026-01-05
Improve FilesystemPackageCacheManager multithread behaviour

dotasek:
This addresses intermittent failures in: https://github.com/hapifhir/org.hl7.fhir.core/blob/bb04abdb128e737aeb45bddbe7bd7530944f5198/org.hl7.fhir.utilities/src/test/java/org/hl7/fhir/utilities/npm/FilesystemPackageManagerTests.java#L366, on Windows systems.

In Windows, there was no actual file set for deletion when lockFile logic was cleaning up after performing write operations.

This PR also includes additional debugging logs, and clearer testing logic.


grahamegrieve (approved):
ok

PR #2278, opened 2026-01-09
Adding better support for extended operations in OpenAPI

jkiddo:
Adding minimal support for extended operations in the generated OpenAPI file


jkiddo:
@dotasek would this be something you can carve out time for?

jkiddo:
@dotasek 
> This complains about the following:
> 
> ```
> ./swagger-base.json: Error: https://build.fhir.org/OperationOutcome.xsd
> ```
Thats not introduced by this commit
> 
>  ... nor does it validate using openapi-spec-validator (https://github.com/python-openapi/openapi-spec-validator).

Would you like to have that added in the pipeline?
> In the Swagger UI, everything appears normal until I try to expand one of the operations. Then it freezes my browser without specific complaint.
> 
That sounds more like the browser is the problem - not the generated file

dotasek:
> Thats not introduced by this commit

> Would you like to have that added in the pipeline?

I understand that the OpenAPI code doesn't have much in the way of testing, and that this in fact improves on that.

> That sounds more like the browser is the problem - not the generated file

Could well be the case. 

For my review I was trying to find a way to test the generated OpenAPI without relying on my own eyes. Something like at least a successful parse with an OpenAPI library, even without validation, would be better than nothing.

However I'm OK with this code if @grahamegrieve is OK with it.

dotasek (approved):
See comment thread. I'm OK if Grahame is OK.

PR #2279, opened 2026-01-10
possible fix for #2138

### LINKED ISSUE #2138: InstanceValidator: Cross-version extensions cannot be used as discriminator in slices
Hi,

We wanted to use the cross version extension `http://hl7.org/fhir/5.0/StructureDefinition/extension-Observation.triggeredBy` as a discriminator for slicing and the InstanceValidator produced an error message.

The following part in the `getCriteriaForDiscriminator` method seems to produce the error:
https://github.com/hapifhir/org.hl7.fhir.core/blob/134a726e4f22658e4b68d77a316f129a6cb171f6/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java#L5113-L5118

It looks like only the context is used to determine the definition of the extension, the `isXverUrl` and `xverDefn` methods are not used. Also, there is a `fetcher` field, I am unsure what the exact purpose of this is. However, the `lookupProfileReference` method uses the `fetcher`, so this is an indication that this may also be a source for looking up profiles.

Best regards,
Max

oliveregger:
we got noted in matchbox https://github.com/ahdis/matchbox/issues/424 that r5 extensions in r4 are throwing an error when validating against a profile where the r5 extensions are listed as discriminators. after a second validation in the same context the problem disappeared. we tracked it down that the xversion where not considered inside the slicing, attached a fix which helped solved our problem and i assume it should also fix #2138

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java
@@ -4073,7 +4073,16 @@ public class ProfileUtilities {
     if (source != null && source.getSourcePackage() != null && source.getSourcePackage().isCore()) {
       source = null;
     }
-    return context.fetchResource(StructureDefinition.class, u, v, source);
+    StructureDefinition sd = context.fetchResource(StructureDefinition.class, u, v, source);
+    if (sd == null) {
+      if (makeXVer().matchingUrl(u) && xver.status(u) == XVerExtensionStatus.Valid) {
+        sd = xver.getDefinition(u);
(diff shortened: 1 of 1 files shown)


bvdh:
@grahamegrieve asked twice here for a test case in the `fhir-test-cases` validation tests, and this PR was closed when none arrived. That test case now exists: FHIR/fhir-test-cases#280.

I've opened #2599 against the same symptom. It is a different fix from this one — this PR addressed an xver fallback in `ProfileUtilities.findProfile`, whereas #2599 fixes `extension(url)` discriminator resolution in `FHIRPathEngine.evaluateDefinition()`. As far as I can tell the two are complementary rather than overlapping, so this may still be worth revisiting on its own merits.

PR #2281, opened 2026-01-13
Changes to support enhanced rendering of requirements statements

lmckenzi:
1. Fixed XHTMLToMarkdownConverter - it was removing spaces around bolded and similar elements.  (See updated test cases)
2. Added a removeLanguageTranslation method to ExtensionUtilities
3. Added hasClass method to XhtmlNode
4. Added constants, messages, and rendering phrases (with translations) to support enhanced rendering of requirements statements


lmckenzi:
See TestCase fix https://github.com/FHIR/fhir-test-cases/pull/250

Issue #2283, opened 2026-01-14
Enhancement - Add timeout setting to stop validation processing after a set number of seconds/milliseconds


rettema:
**From discussions during day 1 of the FHIR Connectathon 41 - Testing Track - the ability in core validation to specify a time out.**

_(Description by David Otasek after discussion with Richard Ettema and review of existing functionality coded by Richard in his local environment)_

If a resource being validated exceeds a set timeout value (seconds or milliseconds), the validator should:

- stop all validation functionality
- preserve any existing issues
- add an issue (warning, error, or fatal, which can be discussed) indicating that the timeout was reached and that the existing issue list may not be 100% correct.

This would actually help the online validator as well, as the user experience is expected to be for lighter weight validations, and a good default value would 'fail-fast' instead of resulting in client timeouts. Richard has coded his own implementation of this, which looks effective and would have to be correctly integrated into the command-line and service options.

**Follow on request by David Otasek to Richard Ettema**

Richard Ettema an online fork of either of these or better yet a draft PR that we can discuss on GitHub, that would be great.

PR #2287, opened 2026-01-15
Add support for validation processing time out

rettema:
### What does this pull request do?
This initial logic adds the ability to specify a validation processing time out to the Core Validation logic where, during the validation of a resource if the duration of that processing exceeds a set time out in milliseconds, the validator will:

- stop all validation functionality
- preserve any existing issues
- add a warning issue (message severity to be discussed) indicating that the time out was reached and that the existing issue list may not be 100% correct.

This would actually help the online validator as well, as the user experience is expected to be for lighter weight validations, and a good default value would 'fail-fast' instead of resulting in client timeouts

### Related issue(s)
- https://github.com/hapifhir/org.hl7.fhir.core/issues/2283

### To Dos
- Code review to determine appropriate and complete use of check for validation time out; currently only coded in the InstanceValidator class
- Implementation of Validator CLI to set time out value
- Current logic to add final warning message should be updated to leverage the message.properties and I18nConstants class
- Final warning message should be reviewed to determine if message severity is appropriate


dotasek:
After looking at the tracking code for a bit longer, maybe this needs a slight refactor to keep the TimeTracker from getting dragged into this too much.

org.hl7.fhir.utilities/src/main/resources/Messages.properties
```diff
@@ -1512,3 +1512,4 @@ CONFORMANCE_STATEMENT_WORD = The html source contains the word ''{0}'' but it is
 VALUESET_CODE_CONCEPT_HINT = {3}. Note that the display in the ValueSet does not have to match; this check exists to help check that it''s not accidentally the wrong code
 VALUESET_CODE_CONCEPT_HINT_VER ={3}. Note that the display in the ValueSet does not have to match; this check exists to help check that it''s not accidentally the wrong code
 TERMINOLOGY_TX_SYSTEM_UNSUPPORTED = The code cannot be checked because codeSystem ''{0}'' version ''{1}'' is not supported ({2})
+VALIDATION_TIMEOUT_EXCEEDED = Validation process exceeded maximum allowed time of {0}ms. Returned validation messages may be incomplete or inaccurate.
```
grahamegrieve: I think that this is a little ambiguous - who allowed it? It'll generate support work by being vague

dotasek:
> well, the relevance of the web site here is that the limit might not be user provided, but in the cli context it is. The message should be clear, or we'll get to support it....

I could make timeout its own object, with a settable field to indicate what set it. 

```java
TimeoutSettings(long timeoutMillis, String source) {/*...*/}

//...

timeoutSettings = new TimeoutSettings(options.timeoutSettings, "CLI option -validation-timeout");

// or 

timeoutSettings = new TimeoutSettings(manualValidationTimeout, "Web Validator Defaults")
```

Message.properties
```
Validation process exceeded maximum allowed time of {0}ms, set by {1}. Returned validation messages may be incomplete or inaccurate.
```

```java
formatMessage(TIMEOUT_MESSAGE, timeout.timeoutMillis, timeout.source);
```

@rettema @grahamegrieve is that an improvement?

PR #2288, opened 2026-01-16
Remove test code that has not run since initial git commit

dotasek:
Remove test code that has not run since initial git commit.

Initially, this code was flagged for having no assertion statements, but git history revealed that the majority of this code has been commented out or deprecated since the initial commit to git.

(this is the same as #2280 but rebased to master in order to fix issues with the git tree)


grahamegrieve (approved):
ok

PR #2291, opened 2026-01-18
Correct rendering of Permit and Deny code

JohnMoehrke:
Today the rendering is wrongly using opt-in and opt-out.
close https://github.com/HL7/fhir-ig-publisher/issues/1235


grahamegrieve:
this PR will not be accepted - it's not going to lead to a solution

PR #2302, opened 2026-01-27
2026 01 gg http server

grahamegrieve:


org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackage.java
```diff
@@ -216,6 +216,20 @@ public static boolean isValidVersion(String ver) {
     return ver.matches("^[0-9]+\\.[0-9]+\\.[0-9]+$");
   }
 
+  public static boolean isValidNameWithVersion(String name, boolean optionalVersion) {
```
dotasek, line 219: Used in IGPublisher?

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackage.java
```diff
@@ -216,6 +216,20 @@ public static boolean isValidVersion(String ver) {
...
+    if (name.contains("#")) {
```
dotasek, line 223: Could use `name != null &&` to prevent NullPointerException

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackage.java
```diff
@@ -216,6 +216,20 @@ public static boolean isValidVersion(String ver) {
...
+      String n = name.substring(0, name.indexOf("#"));
+      String v = name.substring(name.indexOf("#") + 1);
+      if (v == null || v.isEmpty()) {
+        return isValidName(n) && optionalVersion;
+      } else {
+        return isValidName(n) && (isValidVersion(v) || Utilities.existsInList(v, "current", "dev"));
+      }
+    } else {
+      return isValidName(name);
```
dotasek: Should this also check for 'optionalVersion'?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/OperationOutcomeUtilities.java
```diff
@@ -242,4 +242,17 @@ public static OperationOutcome outcomeFromTextError(String text) {
     return oo;
   }
 
+  public static OperationOutcome createSuccess(String text) {
```
dotasek: Is this not better to put alongside the other http classes, where there are similar helper methods?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/http/BaseHTTPHandler.java
```diff
@@ -0,0 +1,146 @@
...
+public class BaseHTTPHandler {
+
+  protected List<String> parseListParameter(String param) {
+    if (param == null || param.trim().isEmpty()) {
+      return new ArrayList<>();
+    }
+    return Arrays.asList(param.split(","));
+  }
+
+  protected Map<String, String> parseQueryParams(String query) {
```
dotasek, line 29: If I'm not mistaken, there is already some code in our codebase that uses a library for query param parsing. Maybe something we should rely on instead?

dotasek: We use URLEncodedUtils from org.apache.http.client.utils, for example, and I'm certain apache will be better at managing all sorts of weird edge cases with decoding.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/http/BaseHTTPHandler.java
```diff
@@ -0,0 +1,146 @@
...
+    if (param == null) return BestPracticeWarningLevel.Ignore;
+    return BestPracticeWarningLevel.valueOf(param);
+  }
+
+  protected  CheckDisplayOption parseCheckDisplayOption(String param) {
+    if (param == null) return CheckDisplayOption.Ignore;
+    return CheckDisplayOption.valueOf(param);
+  }
+
+  protected  OperationOutcome createSuccessOperationOutcome(String message) {
```
dotasek: This and the following seem like they could be a single `createOperationOutcome(message, severity, issueCode)` method.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/http/StopHTTPHandler.java
```diff
@@ -0,0 +1,68 @@
...
+  @Override
+  public void handle(HttpExchange exchange) throws IOException {
+    if (!"POST".equals(exchange.getRequestMethod())) {
+      sendResponse(exchange, 405, "Method not allowed", "text/plain");
+      return;
+    }
+
+    try {
+      OperationOutcome outcome = OperationOutcomeUtilities.createSuccess("Stopping");
+      sendOperationOutcome(exchange, 200, outcome, getAcceptHeader(exchange));
```
dotasek, line 52: Is there no concept of a callback or something we could rely on here instead of the delay below?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/http/ValidateResourceHTTPHandler.java
```diff
@@ -0,0 +1,73 @@
...
+      bpWarnings = parseBestPracticeWarningLevel(params.get("bpWarnings"));
+      displayOption = parseCheckDisplayOption(params.get("displayOption"));
+    } catch (Exception e) {
+      OperationOutcome outcome = createErrorOperationOutcome("Operation failed: " + e.getMessage());
+      sendOperationOutcome(exchange, 400, outcome, getAcceptHeader(exchange));
+      return;
+    }
+    try {
+      // Validate the resource using ValidationEngine
+      OperationOutcome outcome = fhirValidatorHttpService.getValidationEngine().validate("http-request", resourceBytes, format, profiles,
```
dotasek, line 64: This seems like a different entrypoint to validation than the ValidationService takes. Tracing this, I get to: https://github.com/hapifhir/org.hl7.fhir.core/blob/05ef26056fea874a0bced92d06681d859f4f4334/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ValidationEngine.java#L740

However, ValidationService, which the CLI normally uses, gets to https://github.com/hapifhir/org.hl7.fhir.core/blob/05ef26056fea874a0bced92d06681d859f4f4334/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ValidationEngine.java#L619

I wonder if this might cause some agony down the road if there are some subtle differences between the two.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/http/StopHTTPHandler.java
dotasek, line 1: Noting that these new HTTP Handlers don't appear to have any tests anywhere.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackage.java
grahamegrieve: yes

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackage.java
grahamegrieve: ok

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackage.java
grahamegrieve: yes

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/OperationOutcomeUtilities.java
grahamegrieve: this is not an http helper, it's an OperationOutcome helper.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/http/BaseHTTPHandler.java
grahamegrieve: well, this is a refactoring for me - just moving the code that we already had. I'm certainly happy to use a different parsing approach, but let's make it some other problem

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/http/BaseHTTPHandler.java
grahamegrieve: ok. they should be in OperationOutcomeUtilities, so I refactored that generally

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/http/StopHTTPHandler.java
grahamegrieve: not according to Claude. this is thread owned by the server we want to stop, and just closing it creates a deadlock.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/http/ValidateResourceHTTPHandler.java
grahamegrieve: actually, the real difference is between the validate functions on line 694 and 740, and that's really made me think. It could well be a source of trouble, but I'm not sure whether to even fix it or not. So I'm not taking it on in this PR since I'm just moving existing code around

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/http/StopHTTPHandler.java
grahamegrieve: no and nor did the existing http command. Problem for another day?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/PathSlicingParams.java
```diff
@@ -17,9 +20,18 @@ public class PathSlicingParams {
...
   final String path;
 
+  final List<ElementDefinition> slices = new ArrayList<>();
+
   public PathSlicingParams() {
     done = false;
     elementDefinition = null;
     path = null;
   }
+
```
dotasek, line 30: Is `slices` ever intended to contain ElementDefinitions that are not diffMatches? If it is exclusively diffMatches, it should be named as such.

dotasek (changes requested):
One renaming might be prudent; see comment.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/PathSlicingParams.java
grahamegrieve: diffMatches are not always slices, but these are always slices in this context - that's why I named them that way

PR #2307, opened 2026-01-29
Fix snapshot generation for nested slices with contentReference

glichtner:
# Summary

When resolving a `contentReference` inside a slice during snapshot generation, the `resolveContentReference` method incorrectly matches slices instead of the base element. This causes slice-specific constraints (like patterns) to be incorrectly propagated to nested elements.

# Reproduction

A minimal reproduction case is available at: https://github.com/glichtner/fhir-validator-nested-section-issue

The test case defines two semantically identical profile hierarchies:

## **FAIL case** - Slicing defined across two profiles (inheritance):

https://github.com/glichtner/fhir-validator-nested-section-issue/blob/main/input/fsh/composition-not-working.fsh

```
ParentProfile_FAIL [Parent: Composition]
* section contains introduction 0..1
* section[introduction]
  * code = #introduction

ChildProfile_FAIL [Parent: ParentProfile_FAIL]
* section[introduction]
  * section contains text 0..1
  * section[text]
    * code = #text
```

## **PASS case** - All slicing defined in one profile:

https://github.com/glichtner/fhir-validator-nested-section-issue/blob/main/input/fsh/composition-working.fsh

```
ParentProfile_PASS [Parent: Composition]

ChildProfile_PASS [Parent: ParentProfile_PASS]
* section contains introduction 0..1
* section[introduction]
  * code = #introduction
* section[introduction]
  * section contains text 0..1
  * section[text]
    * code = #text
```

Both define the same logical structure (simplified):
```
Composition
* section[introduction]
  * code = #introduction
  *  section[text]
    * code = #text
```

## Expected Behavior

Instances of `ChildProfile_PASS` and `ChildProfile_FAIL`  should both validate successfully since they conform to their respective (semantically identical) profiles.

# Actual Behavior (before fix)

The FAIL case produces a validation error:

```
-- fsh-generated/resources/Composition-ExampleInstance.FAIL.json ---------------------------------------------------------------------
*FAILURE*: 1 errors, 0 warnings, 0 notes
  Error @ Composition.section[0].section[0].code (line 26, col12): The pattern [system http://example.org/CodeSystem/my-code-system, code introduction, and display 'Introduction'] defined in the profile http://example.org/StructureDefinition/child-profile-fail not found. Issues: [ValidationMessage[level=ERROR,type=VALUE,location=Composition.section[0].section[0].code.coding.code,message=Value is 'text' but is fixed to 'introduction' in the profile http://example.org/StructureDefinition/child-profile-fail#Composition.section:introduction.section], ValidationMessage[level=ERROR,type=VALUE,location=Composition.section[0].section[0].code.coding.display,message=Value is 'Text' but is fixed to 'Introduction' in the profile http://example.org/StructureDefinition/child-profile-fail#Composition.section:introduction.section]]
--------------------------------------------------------------------------------------------------------------------------------------

-- fsh-generated/resources/Composition-ExampleInstance.PASS.json ---------------------------------------------------------------------
Success: 0 errors, 0 warnings, 1 notes
  Information: All OK
--------------------------------------------------------------------------------------------------------------------------------------
```

Called using
```bash
sushi && java -jar validator_cli.jar -ig fsh-generated/ -recurse -- fsh-generated/resources/Composition-ExampleInstance.FAIL.json fsh-generated/resources/Composition-ExampleInstance.PASS.json
```
> FHIR Validation tool Version 6.7.11 (Git# 82fb5ebf8e8f). Built 2026-01-20T17:32:33.991Z Java: 17.0.17 from /usr/lib/jvm/
> java-17-openjdk-amd64 on amd64 (64bit). 7952MB available

The validator incorrectly expects the nested `section.section.code` to have the value "introduction" (from the parent section) instead of "text".

# Root Cause

The problem seems to be in `ProfilePathProcessor.resolveContentReference()`. When expanding a `contentReference` (e.g., `Composition.section.section` references back to `#Composition.section`), the method walks backwards through the snapshot looking for an element with a matching path.

But slices have the same path as their base element, so when resolving the contentReference for `section:introduction.section`, the method finds `section:introduction` (which has path `Composition.section`) before finding the actual base `Composition.section`.

This causes the children of the slice (including `section:introduction.code` with the "introduction" pattern) to be copied into `section:introduction.section.*`, incorrectly applying the parent's constraints to nested elements.

# Suggested Fix

Add a check to ensure only base elements are matched (those without a slice name) when resolving content references:

```java
// Before:
if (path.equals(ed.getPath())) {
  return res;
}

// After:
if (path.equals(ed.getPath()) && !ed.hasSliceName()) {
  return res;
}
```

# Testing
Verified the fix with the reproduction case - both FAIL and PASS cases now validate successfully

```
-- fsh-generated/resources/Composition-ExampleInstance.FAIL.json ---------------------------------------------------------------------
Success: 0 errors, 0 warnings, 1 notes
  Information: All OK
--------------------------------------------------------------------------------------------------------------------------------------

-- fsh-generated/resources/Composition-ExampleInstance.PASS.json ---------------------------------------------------------------------
Success: 0 errors, 0 warnings, 1 notes
  Information: All OK
--------------------------------------------------------------------------------------------------------------------------------------
```


dotasek (changes requested):
The behaviour on the expected test case appears to match the requirements in the issue (both cases pass).

However, this is failing (at least for me) on the SnapshotGenerationTests here:

https://github.com/FHIR/fhir-test-cases/blob/master/r5/snapshot-generation/comp-nested-input.json
https://github.com/FHIR/fhir-test-cases/blob/master/r5/snapshot-generation/comp-nested-expected.xml
https://github.com/FHIR/fhir-test-cases/blob/master/r5/snapshot-generation/comp-nested-base.json

If this change is expected, there should be an additional PR on fhir-test-cases with an explanation of why it is valid.

This should also merge changes from the current master branch so all the other tests can also be run.

PR #2308, opened 2026-01-30
Fixed issues with missing translation property, allow use of markdown…

lmckenzi:
… to HTML by external classes


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/DataRenderer.java
```diff
@@ -117,7 +117,7 @@ protected void addMarkdown(XhtmlNode x, String text, String path) throws FHIRFor
     addMarkdown(x, processRelativeUrls(text, path)); 
   } 
 
-  protected void addMarkdown(XhtmlNode x, String text) throws FHIRFormatError, IOException, DefinitionException { 
+  public void addMarkdown(XhtmlNode x, String text) throws FHIRFormatError, IOException, DefinitionException {
```
grahamegrieve, line 122: when can this be used safely from something that's not a descedent class?

lmckenzi: The only reason this method isn't static is that it needs workerContext to look up resources.  You don't need to be a descendant to use the method.

grahamegrieve (approved):
ok

PR #2322, opened 2026-02-21
perf: O(1) drop and O(n) getList in CanonicalResourceManager

jmandel:
## Summary

Three performance improvements to `CanonicalResourceManager`:

1. **O(1) `drop()`**: Add a reverse index (`mapInverse`: value → keys) so that `drop()` can remove map entries in O(1) instead of scanning all entries.

2. **O(n) `getList()`**: Use an `IdentityHashMap`-backed Set for deduplication instead of `List.contains()`, which was O(n²).

3. **Cached comparator**: Reuse a single `MetadataResourceVersionComparator` instance instead of allocating a new one on every `see()` and `updateList()` call.

## Profiling Data

Profiled with JFR (`settings=profile`, 10ms sampling) against US Core 9.0.0.

- `getList()` inclusive CPU samples: **4,288 → 308 (-93%)**
- "Check profiles & code systems" phase: **26s → 15s (-44%)**

## A/B Benchmark

3 runs each, same warm terminology cache, same IG (US Core 9.0.0), same machine. This PR is one of several independent changes that together produce:

```
Upstream avg:  3:54
Optimized avg: 2:41  (31% faster)
```

Full benchmark details: https://github.com/jmandel/igpublisher-perf

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/CanonicalResourceManager.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/CanonicalResourceManager.java
@@ -266,8 +266,10 @@ public class CanonicalResourceManager<T extends CanonicalResource> {
   private Map<String, List<CachedCanonicalResource<T>>> listForId;
   private Map<String, List<CachedCanonicalResource<T>>> listForUrl;
   private Map<String, CachedCanonicalResource<T>> map;
+  private Map<CachedCanonicalResource<T>, Set<String>> mapInverse; // reverse index: value → keys for O(1) drop
   private Map<String, List<CachedCanonicalResource<T>>> supplements; // general index based on CodeSystem.supplements
   private String version; // for debugging purposes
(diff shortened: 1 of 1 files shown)


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/CanonicalResourceManager.java
```diff
@@ -575,11 +591,21 @@ private void addToSupplements(CanonicalResourceManager<T>.CachedCanonicalResourc
 
 
   public void drop(CachedCanonicalResource<T> cr) {
-    while (map.values().remove(cr)); 
-    while (listForId.values().remove(cr)); 
-    while (listForUrl.values().remove(cr)); 
+    // use reverse index to remove map entries in O(1) instead of scanning all values
+    Set<String> keys = mapInverse.remove(cr);
```
dotasek, line 595: This doesn't look like it's properly updating listForId or listForUrl according to the old logic. They are still being used in the code elsewhere assuming the old logic for drop.

PR #2325, opened 2026-02-21
perf: cache SemverParser results, getResourceNames, and JsonFactory

jmandel:
## Summary

Three small caching fixes:

1. **`SemverParser.parseSemver()`**: Add a `ConcurrentHashMap` cache for parse results. This method is called frequently from `CanonicalResourceManager` version comparisons via `VersionUtilities.isSemVer()`. (113 CPU samples in profiling)

2. **`SimpleWorkerContext.getResourceNames()`**: Cache the result with a lazy-init `volatile` field instead of rebuilding the list on every call. (Part of the `getResourceNames` 3% hotspot)

3. **`NpmPackageIndexBuilder`**: Share a single static `JsonFactory` instance instead of creating a new one per use.

## A/B Benchmark

Part of a set of independent changes producing 31% overall speedup. Full details: https://github.com/jmandel/igpublisher-perf


grahamegrieve:
see #2327 for a better way to do the resource names part

PR #2327, opened 2026-02-21
2026 02 gg definitions version count

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java
```diff
@@ -5637,29 +5646,28 @@ private List<Base> funcExists(ExecutionContext context, List<Base> focus, Expres
...
 
-
   private List<Base> funcResolve(ExecutionContext context, List<Base> focus, ExpressionNode exp) throws FHIRException {
     List<Base> result = new ArrayList<Base>();
-    Base refContext = null;
     for (Base item : focus) {
-      String s = convertToString(item);
+      Base refContext = item;
+      String url = null;
+      Identifier id = null;
```
dotasek, line 5667: This value is never set. I believe it should be at 5663?

grahamegrieve: it should indeed. thanks

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/i18n/POSource.java
```diff
@@ -131,7 +131,7 @@ private static POSource loadPOFile(String[] lines) throws FileNotFoundException,
         } else { // if (!Utilities.noString(s)) {
           poObject.getMsgstr().add(s);
         }
-      } else {
+      } else if (!line.startsWith("##")) {
```
dotasek: This is an odd one. I can't find any references to a `##` line in the PO docs and none in our collection of language files. 

Should this just be `#`?

In addtion, I think it would be better to do this as a debug and an else, like so:

```java
else if (line.startsWith("#")) {
   log.debug("Unprocessed PO comment: {}", line);
} else {
  log.warn("Unknown line: {}", line);
}
```

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackageIndexBuilder.java
```diff
@@ -68,7 +68,8 @@ public boolean seeFile(String name, byte[] content) {
        * We can then use a streaming parser to get the values of these fields instead of parsing the whole file and
        * allocating memory for everything in it.
        */
-      try (final var parser = new JsonFactory().createParser(content)) {
+      JsonFactory jsonFactory = new JsonFactory();
```
dotasek: It looks like this doesn't do much. Was there an additional step like cacheing that was meant for jsonFactory?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/validation/ValueSetValidator.java
```diff
@@ -1479,8 +1481,11 @@ private boolean scanForCodeInValueSetInclude(String code, Set<String> sys, List<
           ValidationResult vr = context.validateCode(options.withNoClient(), code, vsDummy);
           if (vr.isOk()) {
             sys.add(vsi.getSystem());
-          } else {
-            // ok, we'll try to expand this one then 
+          } else if (vr.IsNoService()) {
```
dotasek: Incidental nitpick: `IsNoService` should conform to method naming convention and be `isNoService`

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java
```diff
@@ -1644,6 +1654,19 @@ public static Quantity makeQuantity(Base base) {
     return null;
   }
 
+  public static Identifier makeIdentifier(Base base) {
```
dotasek, line 1657: Is the not more accurately `extractIdentifier` or `getIdentifierFromBase`?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/TerminologyClientManager.java
```diff
@@ -314,6 +317,37 @@ public TerminologyClientContext chooseServer(ValueSet vs, Set<String> systems, b
     }
   }
 
+  private void checkActuallySupports(ServerOptionList choice) {
+    boolean hasPrimary = false;
```
dotasek: Never used?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/TerminologyClientManager.java
```diff
@@ -433,8 +467,44 @@ private ServerOptionList decideWhichServer(String url) {
...
+  private boolean checkCSAvailable(String server, String canonical) {
+    if (isTxFhirOrg(server)) {
+      return true;
+    }
+    String key = server+"^"+canonical;
+    if (serverSupportMap.containsKey(key)) {
+      return serverSupportMap.get(key);
+    }
+    try {
+      var client = findClient(server, null, false);
```
dotasek, line 481: The following section looks like it does the following: search for a codesystem by canonical and version if available. 

If a bundle with a single entry is returned by the query, return that.

If not, try again with just the codesystem with no version, and if a bundle with a single entry is returned, use that.

Is that correct?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/expansion/ValueSetExpander.java
```diff
@@ -672,51 +673,80 @@ private void excludeCode(WorkingContext wc, String theSystem, String theCode) {
   private void excludeCodes(WorkingContext wc, ConceptSetComponent exc, Parameters expParams, ValueSetExpansionComponent exp, ValueSet vs, String vspath) throws FHIRException, FileNotFoundException, OperationIsTooCostly, IOException {
     opContext.deadCheck("excludeCodes");
     exc.checkNoModifiers("Compose.exclude", "expanding");
-    if (exc.hasSystem() && exc.getConcept().size() == 0 && exc.getFilter().size() == 0) {
+    if (exc.hasSystem() && exc.getConcept().size() == 0 && exc.getFilter().size() == 0 && exc.getValueSet().size() == 0) {
       wc.getExcludeSystems().add(exc.getSystem());
-    }
+    } else {
```
dotasek: I think this, and another following else, could be removed by replacing them with a guard clauses that return early.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/ObjectConverter.java
```diff
@@ -130,6 +120,12 @@ public DataType convertToType(Element element) throws FHIRException {
     return b;
   }
 
+  public DateTimeType readAsDateTime(Element e) {
+    DateTimeType b = (DateTimeType) new Factory().create("dateTime");
+    ((PrimitiveType) b).setValueAsString(e.primitiveValue());
```
dotasek: Is the cast necessary? IntelliJ seems happy without it.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/Utilities.java
```diff
@@ -191,6 +191,28 @@ public static String insertBreakingSpaces(String text, Set<Character> breakingCh
     return result.toString();
   }
 
+  public static String escapeUrl(String canonical) {
```
dotasek, line 194: This looks like more of a job for an apache http utility. Is there a particular reason it's hand-coded?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/TxTestData.java
```diff
@@ -104,7 +108,7 @@ public static TxTestData loadTestDataFromPackage(String source) throws IOExcepti
     Map<String, TxTestSetup> examples = new HashMap<String, TxTestSetup>();
     JsonObject manifest = org.hl7.fhir.utilities.json.parser.JsonParser.parseObject(contents);
     for (JsonObject suite : manifest.getJsonObjects("suites")) {
-      if (!"tx.fhir.org".equals(suite.asString("mode"))) {
+      if (!"tx.fhir.org".equals(suite.asString("mode")) && !omissions.contains(suite.asString("name"))) {
```
dotasek, line 111: Is there a way to align this with the new `included` / `excluded` filtering we're doing in validation tests? Maybe not in logic right away, but at least in nomenclature?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/model/TimeType.java
```diff
@@ -215,4 +215,12 @@ public String fpValue() {
     return "@T"+primitiveValue();
   }
 
+  public boolean before(TimeType other) {
```
dotasek, line 218: For both of these, primitiveValue in `this` and in `other` could be null and cause a NullPointerException

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
```diff
@@ -3908,4 +3921,29 @@ public IOIDServices oidServices() {
...
+    definitionsVersion++;
+    synchronized (lock) {
+      analyses.clear();
+      cachedResourceNames = null;
+      cachedResourceNameSet = null;
+    }
+  }
+  public void storeAnalysis(Class className, Object analysis) {
+    synchronized (lock) {
+      analyses.put(className.getName(), analysis);
```
dotasek, line 3939: I think we might be able to simplify this with a ConcurrentHashMap.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/i18n/POSource.java
grahamegrieve: it allows the source to have a line that starts with # that doesn't create a warning. Is that a problem?

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackageIndexBuilder.java
grahamegrieve: not as far as I knew - we talked about this one - this is what Josh's PR suggested

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/validation/ValueSetValidator.java
grahamegrieve: fixed

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java
grahamegrieve: well, maybe, but we already had makeQuantity, so I stuck with the same pattern

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/TerminologyClientManager.java
grahamegrieve: removed

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/TerminologyClientManager.java
grahamegrieve: Yes

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/expansion/ValueSetExpander.java
grahamegrieve: well, ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/ObjectConverter.java
grahamegrieve: leftover copy/paste thing. I removed it

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/Utilities.java
grahamegrieve: it's with all the other escape routines, it's straight-forward, and it was easier to write than chase down and figure out how to call it.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/TxTestData.java
grahamegrieve: not simple, at least. I'll think about it but it has to fit into the tx test framework

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
grahamegrieve: well, you can do that later.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/model/TimeType.java
grahamegrieve: its not clear what else they can do

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/i18n/POSource.java
dotasek: It can totally skip, sure, but I think it's clearer with an explicit match, like:

```java
else if (line.startsWith("#")) {
   // Do nothing for comments.
}
```

But the code in place is also checking for `##` not just `#`.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackageIndexBuilder.java
dotasek: Not exactly; his moved this to a static field: 

```java
private static final JsonFactory JSON_FACTORY = new JsonFactory(); // reuse across all seeFile calls
```

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java
dotasek: Could both match the new pattern? `makeXXX` implies it will mutate the argument.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/model/TimeType.java
dotasek: I didn't check, but what happens if these are null where this is used? Does it need a catch there?

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/i18n/POSource.java
grahamegrieve: well, I'll change it to #. I'm not perfectly sure that's correct though.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackageIndexBuilder.java
grahamegrieve: well, ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/model/TimeType.java
grahamegrieve: it has checks where it's used

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java
grahamegrieve: deferred to the future

PR #2328, opened 2026-02-22
Feat/resolve logical

jkiddo:
This pull request introduces improvements to the handling and testing of logical references in FHIRPath resolution. The main changes ensure that logical references can be resolved when `hostServices` are provided, and adds tests to verify this behavior as well as the fallback when `hostServices` are absent.

Enhancements to reference resolution:

* Updated `funcResolve` in `FHIRPathEngine.java` to attempt resolution of logical references using `hostServices` when available, instead of defaulting to `null.

Testing improvements:

* Added `testResolve_LogicalReferenceWithHostServices` to verify that logical references are resolved correctly when custom `hostServices` are set.
* Added `testResolve_LogicalReferenceWithoutHostServices` to confirm that logical references are not resolved when `hostServices` are not provided.

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java
@@ -5648,8 +5648,10 @@ private TimeType timeAdd(TimeType d, Quantity q, boolean negate, ExpressionNode 
         Property p = item.getChildByName("reference");
         if (p != null && p.hasValues()) {
           s = convertToString(p.getValues().get(0));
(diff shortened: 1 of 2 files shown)


dotasek (changes requested):
The change appears to make sense, but the logic is hard to follow, and this is mostly due to the state of the code before the change.

If I follow the logic:
1. Each item is, by default, converted to String s
2. If the item has the type Reference and a field 'reference', this s is mutated to the value of that reference.
3. If the 'reference' field doesn't exist, and hostServices DOES exist, leave s as it was in step 1. Later, hostServices will try to resolve it.

It appears to work (if I revert it, three of the new test cases fail), BUT I think it actually decreases the clarity of code that is already difficult to follow.

PR #2329, opened 2026-02-23
perf: deduplicate tx server calls for unsupported code systems

jmandel:
## Problem

When the terminology server reports a code system as unsupported (e.g., `http://www.genenames.org`), the validator currently makes a separate HTTP call for **every unique code** from that system. For mCODE, this means 26 sequential round-trips to tx.fhir.org for genenames alone — all returning the same "unsupported" verdict.

Three bugs prevented effective deduplication:

### Bug 1: `fetchCodeSystem` guard in `updateUnsupportedCodeSystems()`
The check `fetchCodeSystem(codeKey) == null` prevented systems like genenames.org from being added to the `unsupportedCodeSystems` dedup set, because they have a local CodeSystem loaded (even a stub/not-present one). But having a local CS doesn't mean the server can validate codes from it — the server's "unsupported" verdict is system-wide.

### Bug 2: Lost `errorClass` on localWarning replacement
When `localWarning` was set and the server returned `CODESYSTEM_UNSUPPORTED`, the replacement `ValidationResult` at line ~1584 didn't preserve the `CODESYSTEM_UNSUPPORTED` errorClass. This caused `updateUnsupportedCodeSystems()` to fail silently since it checks `res.getErrorClass()`.

### Bug 3: Bare early-return lost OperationOutcome issues
The `unsupportedCodeSystems` early-return produced a `ValidationResult` without the OperationOutcome issues that the server would have returned. `InstanceValidator.calculateSeverityForTxIssuesAndUpdateErrors()` uses those issues to generate warnings, so 38 warnings were silently dropped.

### Bug 4: No CC path dedup
`validateCode(CodeableConcept)` had no `unsupportedCodeSystems` check at all.

### Bug 5: TerminologyCache in-memory skip
`TerminologyCache.store()` skipped both in-memory and disk caching for `CODESYSTEM_UNSUPPORTED` results. Changed to skip only disk persistence (preserving the intentional "try again next build" behavior) while caching in memory for within-run dedup.

## Fix

- Remove the `fetchCodeSystem` guard — the server's unsupported verdict is authoritative
- Cache the first server result per unsupported system (`unsupportedSystemResults` map) and replay it for subsequent codes, preserving the full OperationOutcome issues
- Add CC-path dedup with `allUnsupported` check
- Preserve `errorClass` on localWarning replacement
- Cache UNSUPPORTED results in memory (not disk) in TerminologyCache

## Impact (mCODE IG, warm tx cache)

| Metric | Before | After | Change |
|--------|--------|-------|--------|
| TX server calls | 35 | 7 | **−80%** |
| genenames.org calls | 26 | 1 | **−96%** |
| Validation wall time | 27s | 9s | **−67%** |
| Non-CPU gap | 22.5s | 4.7s | **−79%** |
| Errors | 71 | 71 | identical |
| Warnings | 154 | 154–155 | identical |

The effect scales with the number of unsupported system codes in an IG. Any IG referencing external code systems not supported by the tx server benefits.

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
@@ -299,6 +299,7 @@ public abstract class BaseWorkerContext extends I18nBase implements IWorkerConte
   private CanonicalResourceManager<CodeSystem> codeSystems = new CanonicalResourceManager<CodeSystem>(false, minimalMemory);
   private final HashMap<String, SystemSupportInformation> supportedCodeSystems = new HashMap<>();
   private final Set<String> unsupportedCodeSystems = new HashSet<String>(); // know that the terminology server doesn't support them
+  private final Map<String, ValidationResult> unsupportedSystemResults = new HashMap<>(); // first server result per unsupported system, replayed for dedup
   private CanonicalResourceManager<ValueSet> valueSets = new CanonicalResourceManager<ValueSet>(false, minimalMemory);
   private CanonicalResourceManager<ConceptMap> maps = new CanonicalResourceManager<ConceptMap>(false, minimalMemory);
   protected CanonicalResourceManager<StructureMap> transforms = new CanonicalResourceManager<StructureMap>(false, minimalMemory);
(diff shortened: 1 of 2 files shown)


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/utilities/TerminologyCache.java
```diff
@@ -618,15 +618,16 @@ public void store(CacheToken cacheToken, boolean persistent, NamedCache nc, Cach
...
-      return;
-    }
+        && !cacheToken.hasVersion);
 
     boolean n = nc.map.containsKey(cacheToken.key);
     nc.map.put(cacheToken.key, e);
+    if (skipPersist) {
+      return;
+    }
     if (persistent) {
```
dotasek, line 631: Nitpick: while we're here, this if statement could also be a guard clause.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
dotasek, line 1: These changes seem reasonable, but there are test failures here related to error messages that all appear to follow the same pattern:

Expected: `A definition for CodeSystem 'urn:oid:2.51.1.1' could not be found, so the code cannot be validated`
Actual: `A definition for CodeSystem 'urn:oid:2.51.1.1' could not be found, so the code cannot be validated (from https://tx-dev.fhir.org/r4)`

I think the actuals are correct here.

There are also other failures which also appear in the master branch in ...ServiceTests which will need resolution and a merge to here before this can pass.

dotasek (changes requested):

PR #2333, opened 2026-03-02
Fix problem where comparing profiles gets into an infinite loop

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/DefinitionNavigator.java
```diff
@@ -173,6 +177,10 @@ public List<DefinitionNavigator> children() throws DefinitionException {
   }
 
   private void loadChildren() throws DefinitionException {
+    if (level > LEVEL_LIMIT) {
+      System.out.println("Level limit reached @ "+level+": "+globalPath);
```
dotasek: This should be an Slf4J error log.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/comparison/StructureDefinitionComparer.java
```diff
@@ -358,35 +358,39 @@ private boolean compareElements(ProfileComparison comp, StructuralMatch<ElementD
...
-      if (r == null) {
-        comp.getUnion().getSnapshot().getElement().add(l.current().copy());
-        res.getChildren().add(new StructuralMatch<ElementDefinitionNode>(new ElementDefinitionNode(l.getStructure(), l.current()), vmI(IssueSeverity.INFORMATION, "Removed this element", path)));
-      } else {
-        matchR.add(r);
-        StructuralMatch<ElementDefinitionNode> sm = new StructuralMatch<ElementDefinitionNode>(new ElementDefinitionNode(l.getStructure(), l.current()), new ElementDefinitionNode(r.getStructure(), r.current()));
-        res.getChildren().add(sm);
-        def = compareElements(comp, sm, l.globalPath(), null, l, r) || def;
+
+    if (!left.isChildrenFromReference() || !right.isChildrenFromReference()) {
```
dotasek: This could be a simpler guard clause:

```
if (left.isChildrenFromReference() && right.isChildrenFromReference()) {
      return true;
  }
```

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/QuestionnaireRenderer.java
```diff
@@ -72,14 +72,17 @@ public void buildNarrative(RenderingStatus status, XhtmlNode x, ResourceWrapper
       throw new Error("Unknown questionnaire Renderer Mode"); 
     } 
 
-    boolean first = true;
-    for (ResourceWrapper cont : q.children("contained")) {
-      if (first) {
-        x.h2().tx("Contained Resources");
-        first = false;
+    if (context.forValidResource()) {
```
dotasek, line 75: Doesn't this just not render in IG Publisher instead of skipping duplicates as the commit message suggests?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/DefinitionNavigator.java
```diff
@@ -305,6 +314,9 @@ private void loadChildren() throws DefinitionException {
         }
       }
     }
+    for (DefinitionNavigator dn : children) {
```
dotasek, line 316: This feels a little weird to have right here. I feel like it should be included in the instantiation of every DefinitionNavigator instead of having to have special treatment after all the other logic.

DefinitionNavigator itself already has an overly complex constructor though, so some refactoring would have to happen there.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
```diff
@@ -95,6 +96,7 @@ public enum ExpansionPackageGeneratorScope {
   private List<String> ignoreList = new ArrayList<>();
   private FHIRPathEngine pathEngine;
   private List<CanonicalResource> includeList = new ArrayList<>();
+  @Getter @Setter private boolean verbose = true;
```
dotasek, line 99: This just decides whether or not to print info level messages, I'm assuming for development purposes. If that's the case, why not just use log.debug level messages?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
```diff
@@ -579,27 +608,7 @@ private void produceZip() throws IOException {
     }
 
     for (CanonicalResource cr : resources) {
-      zip.addBytes(cr.fhirType()+"-"+cr.getIdBase()+(cr.hasVersion() ? "-"+tokenise(cr.getVersion()): "")+(json? ".json" : ".xml"), composeResource(cr), false);
-    }
-    if (modeParams.contains("expansions")) {
```
dotasek, line 584: In this and a few other places, expansions are no longer generated. This seems intentional, but not linked to any intent in the commits. Why are these being removed?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/DefinitionNavigator.java
grahamegrieve: oops meant to remove that line

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/comparison/StructureDefinitionComparer.java
grahamegrieve: ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/QuestionnaireRenderer.java
grahamegrieve: yes but that's only place where duplicates arise

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/DefinitionNavigator.java
grahamegrieve: another time then

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
grahamegrieve: because there's a context where I don't want messages, indepdendent of the level of the messages

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
grahamegrieve: I cleaned up the logic so it all happens in one place (line 296)

dotasek: I don't see that at all. 296-314 seem unchanged except for logging.

From the looks of this, there are two extra files that were generated that are not anymore.

```java
if (modeParams.contains("expansions")) {
      StringBuilder b = new StringBuilder();
      for (String n : Utilities.sorted(entries.keySet())) {
        TerminologyResourceEntry e = entries.get(n);
        String name = e.valueSet.getIdBase();
        int i = 0;
        while (names.contains(name)) {
          i++;
          name = e.valueSet.getIdBase()+i;
        }
        names.add(name);
        if (e.error == null) {
          b.append(name+","+n+", , "+CommaSeparatedStringBuilder.join(";",e.sources)+"\r\n");
        } else {
          b.append(name+","+n+", \""+Utilities.escapeCSV(e.error)+"\", "+CommaSeparatedStringBuilder.join(";",e.sources)+"\r\n");
        }
        zip.addBytes(name+".json", composeResource(e.valueSet), false);
      }
      zip.addBytes("valuesets.csv", b.toString().getBytes(StandardCharsets.UTF_8), false);
    }
    ```

grahamegrieve: I'm not sure what you want here. We were producing duplicate value sets. This code was what was producing them. We're not producing them anymore.

PR #2338, opened 2026-03-03
Replace mistaken uses of bitwise & comparison with boolean &&

dotasek:
Fixes instances of rule java:S2178 discovered by SonarQube


grahamegrieve (approved):
wow :-(

PR #2341, opened 2026-03-04
Rework ManagedWebAccess

dotasek:
Providing authentication through ManagedWebAccess was becoming very difficult to manage. This reworks our approach to allowing a user to implement their own IHTTPAuthenticationProvider, or if that is not set, relying on an auth provider linked to the fhir-settings file.


grahamegrieve (approved):
ok

PR #2343, opened 2026-03-05
2026 03 gg scoped packages

grahamegrieve:


org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/misc/npm/TemplatePackageWrangler.java
```diff
@@ -1,4 +1,91 @@
...
+        FileUtilities.createDirectory(tgt);
+        npm.unPack(tgt.getAbsolutePath());
+        File p = new File(Utilities.path(tgt, "package"));
+        JsonParser.compose(npm.getNpm(), new File(Utilities.path(tgt, "package.json")), true);
+        FileUtilities.copyFiles(p.getAbsolutePath(), tgt.getAbsolutePath());
+        FileUtilities.deleteAllFiles(p.getAbsolutePath());
+        p.delete();
+        sh.append("echo " + id + "#" + v + "\n");
+        sh.append("cd " + id + "#" + v + "\n");
+        sh.append("npm publish"+(v.contains("-") ? " --tag beta" : "")+"\n");
```
dotasek, line 92: Why the beta publish? Should this not be a production release for anything 1.0.0 and above?

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
```diff
@@ -602,9 +603,9 @@ public NpmPackage addPackageToCache(String id, final String version, final Input
       try {
         if (!sid.equals(extractedNpm.getNpm().asString("name")) || !version.equals(extractedNpm.getNpm().asString("version"))) {
```
dotasek: The block of code starting here still retains the old if statement and then an unnecessary nesting, which will no longer be needed with the code removed.

I refactored this locally by replacing this whole chunk with this, which is a bit more explicit:

```java
 checkVersionAndUpdateIfOutdated(version, extractedNpm, tempDir);

private static void checkVersionAndUpdateIfOutdated(String version, NpmPackage extractedNpm, String tempDir) throws IOException {
    if (!version.equals(extractedNpm.getNpm().asString("version"))) {
      extractedNpm.getNpm().add("original-version", extractedNpm.getNpm().asString("version"));
      extractedNpm.getNpm().remove("version");
      extractedNpm.getNpm().add("version", version);
      FileUtilities.stringToFile(JsonParser.compose(extractedNpm.getNpm(), true), Utilities.path(tempDir, "package", "package.json"));
    }
  }
```

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
```diff
@@ -742,6 +745,12 @@ public NpmPackage loadPackage(String id, String version) throws FHIRException, I
     return addPackageToCache(id, source.version, source.stream, source.url);
   }
 
+  private void checkNotBlacklisted(String id) {
+    if ("fhir.base.template".equals(id)) {
+      System.out.println("This content depends on fhir.base.template which is no longer considered secure to use. See notification on chat.fhir.org (tbd)");
```
dotasek: Use logging system error here.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/BasePackageCacheManager.java
```diff
@@ -213,4 +214,12 @@ protected String stripAlias(String id) {
       return id;
     }
   }
+
+  protected String fixPrefix(String id) {
```
dotasek: `encodeIdForFilesystem` feels more specific.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
```diff
@@ -494,6 +494,7 @@ public void removePackage(String id, String version) throws IOException {
   @Override
   public NpmPackage loadPackageFromCacheOnly(String id, @Nullable String version) throws IOException {
     id = stripAlias(id);
```
dotasek: I know that this is outI don't like that we're losing the original id through mutation here. 

Maybe in the signature it should be 'potentialId' and then
`final String id = stripAlias(potentialId)`

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
```diff
@@ -494,6 +494,7 @@ public void removePackage(String id, String version) throws IOException {
...
+    String fid = fixPrefix(id);
 
     if (!Utilities.noString(version) && version.startsWith("file:")) {
       return loadPackageFromFile(id, version.substring(5));
```
dotasek, line 500: No accountiing for `fid` here?

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
```diff
@@ -494,6 +494,7 @@ public void removePackage(String id, String version) throws IOException {
   @Override
   public NpmPackage loadPackageFromCacheOnly(String id, @Nullable String version) throws IOException {
     id = stripAlias(id);
+    String fid = fixPrefix(id);
```
dotasek: Could be `filesystemId` for clarity.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
```diff
@@ -580,7 +581,7 @@ private String findPackageFolder(String id, String version) throws IOException {
    */
   @Override
   public NpmPackage addPackageToCache(String id, final String version, final InputStream packageTgzInputStream, final String sourceDesc) throws IOException {
-    String sid = stripAlias(id);
+    String sid = fixPrefix(stripAlias(id));
```
dotasek: Same commentary as above. 

```java
final String id = stripAlias(potentialId);
    String filesystemId = encodeIdForFileSystem(id);
```

Then, the below no longer needs calls to stripAlias(id) in the lock code.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
```diff
@@ -589,9 +590,9 @@ public NpmPackage addPackageToCache(String id, final String version, final Input
       NpmPackage extractedNpm = NpmPackage.extractFromTgz(packageTgzInputStream, sourceDesc, tempDir, minimalMemory);
 
       log.info("");
-      log.info("Installing " + sid + "#" + version);
+      log.info("Installing " + stripAlias(id) + "#" + version);
```
dotasek: See above regarding `potentialId`; can remove this `stripAlias`

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
```diff
@@ -589,9 +590,9 @@ public NpmPackage addPackageToCache(String id, final String version, final Input
...
 
-      if ((extractedNpm.name() != null && sid != null && !sid.equalsIgnoreCase(extractedNpm.name()) && !sid.equalsIgnoreCase(extractedNpm.name()+"."+VersionUtilities.getNameForVersion(extractedNpm.fhirVersion())))) {
+      if ((extractedNpm.name() != null && sid != null && !sid.equalsIgnoreCase(fixPrefix(extractedNpm.name())) && !sid.equalsIgnoreCase(fixPrefix(extractedNpm.name())+"."+VersionUtilities.getNameForVersion(extractedNpm.fhirVersion())))) {
```
dotasek: This is getting a bit hairy, but with the change above, this could be:

```java
if (
        extractedNpm.name() != null
          && id != null
          && !id.equalsIgnoreCase(extractedNpm.name())
          && !id.equalsIgnoreCase(extractedNpm.name()+"."+VersionUtilities.getNameForVersion(extractedNpm.fhirVersion()))
        ) {
        if (!suppressErrors && (!id.equals("hl7.fhir.r5.core") && !id.equals("hl7.fhir.us.immds"))) {// temporary work around
          throw new IOException("Attempt to import a mis-identified package. Expected " + id + ", got " + extractedNpm.name());
        }
      }
```

Which I think is clearer.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
```diff
@@ -580,7 +581,7 @@ private String findPackageFolder(String id, String version) throws IOException {
    */
   @Override
   public NpmPackage addPackageToCache(String id, final String version, final InputStream packageTgzInputStream, final String sourceDesc) throws IOException {
-    String sid = stripAlias(id);
+    String sid = fixPrefix(stripAlias(id));
     checkValidVersionString(version, sid);
```
dotasek: Considering the error code in checkValidVersionString, maybe stick to the id instead of the filesystemId

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
```diff
@@ -812,13 +821,12 @@ private NpmPackage nullIfNotCurrentPackage(String id, NpmPackage npmPackage) {
   }
 
   // ----- the old way, from before package server, while everything gets onto the package server
-  private InputStreamWithSrc fetchTheOldWay(String id, String v) {
+  private InputStreamWithSrc fetchTheOldWay(String id, String v) throws FileNotFoundException {
```
dotasek: FileNotFoundException doesn't appear to be thrown, according to IntelliJ

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/FileUtilities.java
```diff
@@ -226,6 +226,11 @@ public static File createDirectory(String path) throws IOException {
     return ManagedFileAccess.file(path);
   }
 
+  public static File createDirectory(File path) throws IOException {
+    ManagedFileAccess.csfile(path.getAbsolutePath()).mkdirs();
```
dotasek, line 230: Needs to account for mkdirs() being false, which means the directories weren't created. Throwing an IOException with this info would work.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/FileUtilities.java
```diff
@@ -491,7 +496,19 @@ public static void deleteAllFiles(String folder, String type) throws IOException
...
+
+  }
+  public static void deleteAllFiles(String folder) throws IOException {
+    File src = ManagedFileAccess.file(folder);
+    String[] files = src.list();
+    for (String f : files) {
+      if (ManagedFileAccess.file(folder + File.separator + f).isDirectory()) {
+        deleteAllFiles(folder + File.separator + f);
+      } else {
+        ManagedFileAccess.file(folder + File.separator + f).delete();
```
dotasek, line 508: Same here; if delete() returns false, this should indicate failure with an IOException

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/misc/npm/TemplatePackageWrangler.java
grahamegrieve: npm rules.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
grahamegrieve: yes, since you don't know of a problem I'll delete it properly

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
grahamegrieve: yes

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/BasePackageCacheManager.java
grahamegrieve: ok

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
grahamegrieve: fixed all the places like that

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
grahamegrieve: it doesn't 't make sense anyway. it can never be equal to both of them at once

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
grahamegrieve: ok. refactored.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
grahamegrieve: ok

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/FileUtilities.java
grahamegrieve: actually, I don't think that's obvious. It's a match for the existing function, so why do something different?

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/FileUtilities.java
grahamegrieve: again, this has a matching function, and I don't think this is obvious

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/FilesystemPackageCacheManager.java
grahamegrieve: no it works, though I had to make a fix on testing

PR #2348, opened 2026-03-13
Add OAuth client_credentials grant support for FHIR server authentication

dionmcm:
## Purpose
The IG Publisher uses this code to authenticate to the configured FHIR terminology server. The existing code supports OAuth by providing a static token in configuration, however if the IG Publisher runs longer than the token expiry the IG Publisher process will fail. This was reported for users with a 30 minute token expiry policy they cannot change, unable to complete the IG Publisher run.

This code addresses this by adding configuration to supply client_id, client_secret, and token endpoint URL. The change uses that configuration to get a token instead of a statically configured token, monitors the token for expiry, and gets a new token when required.

## Summary
  - Add OAuth 2.0 `client_credentials` grant type as a new authentication mode for configured FHIR servers, with automatic token acquisition, caching, and transparent refresh
  - Servers can be configured in `fhir-settings.json` with `authenticationType: "client_credentials"` plus `clientId`, `clientSecret`, and `tokenEndpoint` fields, or programmatically via `withClientCredentials()`
  - On 401/403 responses, the cached token is automatically invalidated and the request retried once with a fresh token
  - Refactor auth header resolution into a single `resolveAuth()` method in `ManagedWebAccessorBase`, eliminating 6 duplicated auth dispatch blocks across `ManagedWebAccessor` and `ManagedFhirWebAccessor`
  - Encapsulated retry logic into a shared `executeWithTokenRetry()` template method in the base class
  - Extract `ParamNames` constants for the validation HTTP handler, replacing scattered string literals                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                   
## Test plan
  - [x] `HTTPTokenManagerTest` — 8 tests covering token fetch, caching, expiry, invalidation, and error handling
  - [x] `ManagedWebAccessAuthTests` — 14 tests covering all auth modes (basic, token, apikey, client_credentials) for both accessor types, including 401/403 retry
  - [x] `OAuthClientCredentialsIntegrationTest` — 7 end-to-end tests with MockWebServer covering the full OAuth flow, transparent token refresh on expiry, and retry on 401/403
  - [x] `ManagedFhirWebAccessorTests` — 2 tests for managed header construction
  - [x] `FhirSettingsTests` — 7 tests including validation of required client_credentials fields
  - [x] `FhirValidatorHttpServiceTests` — updated for `ParamNames` constant usage


org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/HTTPTokenManager.java
```diff
@@ -0,0 +1,231 @@
...
+    } else {
+      log.warn("Token endpoint {} response missing 'expires_in', defaulting to {}s", tokenEndpoint, DEFAULT_EXPIRES_IN);
+      expiresIn = DEFAULT_EXPIRES_IN;
+    }
+
+    long expiresAtMillis = System.currentTimeMillis() + (expiresIn * 1000L);
+    return new CachedToken(accessToken, expiresAtMillis);
+  }
+
+  private static String readResponseBody(HttpURLConnection conn) throws IOException {
```
dotasek: This deviates from our usage of org.hl7.fhir.utilities.FileUtilities#streamToBytes, which is used here for a similar purpose: https://github.com/hapifhir/org.hl7.fhir.core/blob/a1c8a8305b7e67d5a1e38ce579893950ade3ef34/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/SimpleHTTPClient.java#L95

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/ManagedWebAccessor.java
```diff
@@ -17,19 +15,11 @@ public class ManagedWebAccessor extends ManagedWebAccessorBase<ManagedWebAccesso
...
-      headers.put("Authorization", "Bearer " + getToken());
-    } else if (getAuthenticationMode() == HTTPAuthenticationMode.BASIC) {
-      String auth = getUsername() +":"+ getPassword();
-      byte[] encodedAuth = Base64.getEncoder().encode(auth.getBytes(StandardCharsets.UTF_8));
-      headers.put("Authorization", "Basic " + new String(encodedAuth));
-    } else if (getAuthenticationMode() == HTTPAuthenticationMode.APIKEY) {
-      headers.put("Api-Key", getToken());
-    }
 
+  private Map<String, String> newHeaders(String url) throws IOException {
```
dotasek: When adjusting for the recent changes in master, keep in mind that this should definitely not be here. Auth should only be managed by IHTTPAuthenticationProvider implementations for now on.

https://github.com/hapifhir/org.hl7.fhir.core/blob/3871cc69c0516628c19adf824e5e7b0a0531511b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/IHTTPAuthenticationProvider.java#L13

dotasek (changes requested):
See PR comment regarding breaking changes to master. Otherwise, there are a few things I noticed on this quick review that should be considered.

I will re-review once the changes from master have been included.

alexanderkiel:
In the MII in Germany, we would be happy to use this `client_credentials` auth method to connect the validator to a terminology server.

dionmcm:
> @dionmcm thank you for the contribution. The logic so far looks reasonable, but we recently merged breaking changes to ManagedWebAccess its related classes: #2341

Thanks @dotasek for the heads-up on #2341. I've merged master in and moved the resolveAuth logic into the provider implementations as you suggested - client_credentials now lives in ServerDetailsPOJOHTTPAuthProvider.

I scoped it to ServerDetailsPOJOHTTPAuthProvider only. I didn't add it to PackageServerHTTPAuthProvider since npm package servers aren't configured for client_credentials in this change, but happy to extend
it there if that's wanted. Full test suite is green against the merged master.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/HTTPTokenManager.java
dionmcm: Fixed - the hand-rolled BufferedReader/InputStreamReader loop is gone. Both the success and error paths now read through the shared org.hl7.fhir.utilities.FileUtilities helper.

I used FileUtilities.streamToString() rather than streamToBytes() because the token response is parsed immediately as a JSON string (whereas SimpleHTTPClient needs the raw bytes for its HTTPResult).

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/ManagedWebAccessor.java
dionmcm: Thanks - agreed, and done. I've merged master (so this branch now includes #2341) and rebuilt the feature on the new model.

- All auth header generation moved out of ManagedWebAccessor/newHeaders and into the providers. client_credentials is now just another case in ServerDetailsPOJOHTTPAuthProvider.getHeaders(), alongside TOKEN/BASIC/APIKEY. newHeaders() only delegates to getHttpAuthHeaderProvider().getHeaders(url) - no auth logic remains there.
- The shared IHTTPAuthenticationProvider interface is unchanged. The one thing the provider model didn't offer was a hook for the 401/403 to invalidate-token then retry behaviour that client_credentials needs
(the interface is a one-shot header supplier). Rather than touch the shared interface, I added a small optional ITokenInvalidatingAuthProvider interface (invalidateCachedCredentials(URL)).
ServerDetailsPOJOHTTPAuthProvider and HTTPAuthenticationProviderChain implement it, and the accessors invoke it only on 401/403 and only when a client_credentials token was actually invalidated. Non-client_credentials auth is completely unaffected.
- HTTPTokenManager does the token fetch/cache, the provider calls it from the CLIENT_CREDENTIALS case.

That was my best guess, let me know if you'd prefer the retry hook modelled differently.

dotasek:
@dionmcm sorry for the vast delay, we were dealing with more security issues in ManagedWebAccess that were finally resolved. There's been a flurry of security advisories submitted, and we have been trying to keep up.

See this change: https://github.com/hapifhir/org.hl7.fhir.core/pull/2526

They will require another merge conflict resolution in this PR. 

If you have any questions, please ask.

dionmcm:
@dotasek I've rebased onto master and reworked for the `ManagedHTTPClient` changes. Your earlier review items are in too
- the token request now goes through the managed stack (so access policy, protocol allow-listing and SSRF blocking apply to it), 
- the client secret is excluded from `toString`, 
- token-endpoint response bodies are no longer echoed into exceptions, 
- and an incomplete `client_credentials` entry is demoted to `none` with a log rather than throwing

**Scope change**
I've dropped the `-tx-client-id` / `-tx-client-secret` / `-tx-token-endpoint` CLI parameters and I'll close the paired IG Publisher PR (HL7/fhir-ig-publisher#1270).

@grahamegrieve questioned those parameters there and, on revisiting, he was right - everything they did is already possible in `fhir-settings.json`, which can hold several terminology servers at once, each with its own credentials, selected per run by `-tx`.

Supporting a second way to inject server configuration required a `ManagedWebAccess.addServerAuthDetail` that had to decide how CLI-supplied entries combined with settings-supplied ones which just complicated things and didn't deliver much. I was thinking mostly of CI pipelines wanting CLI parameters, but we can revisit that if really needed.

So the PR is now more compact/targetted.

I've tried this out against the Australian NCTS using client credentials manually as well as the added automated tests - seems to work fine. There are 6 failing tests but they appear to be failing on master too and not related to my changes.

@alexanderkiel this doesn't affect the validator use case you mentioned - configuration is just in `fhir-settings.json`.

Please let me know if there are any other issues @dotasek

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/HTTPAuthenticationMode.java
```diff
@@ -4,5 +4,6 @@ public enum HTTPAuthenticationMode {
   NONE,
   BASIC,
   TOKEN,
-  APIKEY
-}
\ No newline at end of file
+  APIKEY,
+  CLIENT_CREDENTIALS
```
dotasek, line 8: issue: this name should specifically mention OAuth. OAUTH_CLIENT_CREDENTIALS gets wordy, but CLIENT_CREDENTIALS is too generic.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/HTTPAuthenticationProviderChain.java
```diff
@@ -42,12 +43,26 @@ public boolean canProvideHeaders(URL url) {
...
         return p.getHeaders(url);
       }
     }
     return Map.of();
   }
+
+  @Override
+  public boolean invalidateCachedCredentials(URL url) {
+    for (IHTTPAuthenticationProvider p : providers) {
+      if (!p.canProvideHeaders(url)) {
```
dotasek, line 58: issue: this can be `if(p.canProvideHeaders(url)) { return p.invalidateCachedCredentials(url); }` same as all the other methods in the class.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/ManagedWebAccess.java
```diff
@@ -136,6 +136,10 @@ static boolean inAllowedPaths(String pathname) {
     return false;
   }
 
+  public static String getUserAgent() {
```
dotasek, line 139: issue: This is already provided by the lombok `@Getter` annotation above.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/HTTPTokenManager.java
```diff
@@ -0,0 +1,286 @@
...
+ * <ul>
+ *   <li>{@link #getToken} returns a cached token if it is still valid, or
+ *       fetches a new one from the token endpoint using a standard OAuth 2.0
+ *       client_credentials POST request.</li>
+ *   <li>Tokens are treated as expired {@value #EXPIRY_BUFFER_SECONDS} seconds before their
+ *       actual expiry, so an about-to-expire token is re-fetched on the next {@link #getToken}
+ *       call (rather than handed out and failing mid-request) — there is no background refresh.</li>
+ *   <li>{@link #invalidateToken} removes a cached token, forcing the next call to
+ *       re-fetch. This is used by the retry logic in
+ *       {@link ManagedWebAccessorBase#executeWithClientCredentialsRetry} (invoked via
```
dotasek, line 33: issue: this javadoc is pointing to a method that does not exist.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/ManagedHTTPClient.java
```diff
@@ -275,15 +278,32 @@ public HTTPResult options(String urlString, String acceptHeader, Iterable<HTTPHe
             location = URLDecoder.decode(location, StandardCharsets.UTF_8);
             uri = uri.resolve(location); // Deal with relative URLs, resolved against the current hop
           }
+          case 401, 403 -> {
```
dotasek, line 281: issue: I think this needs to implement a check here for whether or not the authProvider _can_ invalidate/refresh cached credentials. Falling back to the default implementation response of `false` is less clear than indicating that the authprovider can perform it.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/ManagedWebAccessor.java
```diff
@@ -58,11 +75,8 @@ public HTTPResult get(String url) throws IOException {
 
   public HTTPResult get(String url, String accept) throws IOException {
     return switch (ManagedWebAccess.getAccessPolicy()) {
-      case DIRECT -> {
-        ManagedHTTPClient client = setupManagedHTTPClient(url);
-        yield client.get(url, accept);
-      }
-      case MANAGED ->  ManagedWebAccess.getAccessor().get(getServerTypes(), url, accept, newHeaders(url));
+      case DIRECT -> setupManagedHTTPClient(url).get(url, accept);
```
dotasek, line 78: praise: the refactoring here is cleaner, thank you.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/HTTPTokenManager.java
```diff
@@ -0,0 +1,286 @@
...
+    }
+    if (isBlank(server.getClientId())) {
+      throw new IOException("Server " + server.getUrl() + " uses client_credentials but has no clientId");
+    }
+    if (isBlank(server.getClientSecret())) {
+      throw new IOException("Server " + server.getUrl() + " uses client_credentials but has no clientSecret");
+    }
+  }
+
+  private static boolean isBlank(String s) {
```
dotasek, line 167: issue: We have Apache commons. Replace this with StringUtils.isBlank to keep this a bit cleaner.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/HTTPTokenManager.java
```diff
@@ -0,0 +1,286 @@
...
+        || ("http".equals(url.getProtocol()) && Boolean.TRUE.equals(server.getAllowHttp()));
+    }
+
+    @Override
+    public boolean canProvideHeaders(java.net.URL url) {
+      return true;
+    }
+
+    @Override
+    public java.util.Map<String, String> getHeaders(java.net.URL url) throws IOException {
```
dotasek, line 200: issue: this doesn't throw IOException

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/settings/ServerDetailsPOJO.java
```diff
@@ -57,6 +57,13 @@ public class ServerDetailsPOJO {
 
     Map<String, String> headers;
 
+    String clientId;
```
dotasek, line 60: issue: I think this should be enclosed in its own object to prevent this from colliding with any future changes.

```json
"oauthClientCredentials" : {
  "clientId" : "xxxx",
  "clientSecret" : "yyy",
  "tokenEndpont" : "zzz"
}
```

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/HTTPTokenManager.java
dotasek, line 1: question: Should there be some separation between ServerDetailsPOJO and this class? I think we just need the `clientId`, `clientSecret`, and `tokenEndpont`, which should be encapsulated in their own object (see comments on ServerDetailsPOJO), and 'allowHttp` and `allowPrivateNetwork` for TokenEndpointPolicy. There's no reason for this to accept all the other server details. 

I should also check if we're doing this elsewhere...

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/settings/FhirSettings.java
```diff
@@ -233,7 +242,30 @@ private static void normalizeSettings(FhirSettingsPOJO settings) {
...
+            server.getUrl() != null ? server.getUrl() : "(unknown)", problem);
+          server.setAuthenticationType("none");
+        }
+      }
+    }
+  }
+
+  /** Returns a description of the first missing client_credentials field, or null if complete. */
+  private static String findClientCredentialsProblem(ServerDetailsPOJO server) {
+    if (Utilities.noString(server.getClientId())) {
```
dotasek, line 259: issue: apache `isBlank` is more robust here.

org.hl7.fhir.utilities/src/test/java/org/hl7/fhir/utilities/http/HTTPAuthenticationProviderChainTest.java
```diff
@@ -98,4 +106,84 @@ void getHeaders_routesDifferentUrlsToDifferentProviders() throws MalformedURLExc
     assertThat(chain.getHeaders(urlA)).containsEntry("Authorization", "Bearer for-a");
     assertThat(chain.getHeaders(urlB)).containsEntry("Authorization", "Bearer for-b");
   }
+
+  /** A provider that serves {@code url} and reports whether it discarded cached credentials. */
+  private static IHTTPAuthenticationProvider mockInvalidatingProvider(URL url, boolean invalidated) {
```
dotasek, line 111: issue: SonarQube indicates that invalidated is always true here.

org.hl7.fhir.utilities/src/test/java/org/hl7/fhir/utilities/http/HTTPTokenManagerTest.java
```diff
@@ -0,0 +1,333 @@
...
+      .addHeader("Content-Type", "application/json")
+      .setResponseCode(200));
+
+    assertThatThrownBy(() -> HTTPTokenManager.getToken(server))
+      .isInstanceOf(IOException.class)
+      .hasMessageContaining("missing 'access_token'");
+  }
+
+  @Test
+  void testMissingExpiresInDefaultsTo3600() throws Exception {
```
dotasek, line 172: nitpick: this is not actually checking that this defaults to 3600, but just that a default value covers the timespan of our two getToken calls.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/ManagedHTTPClient.java
dotasek, line 281: issue: I think the 403 response should be checking for reasons for the error. 403 may be indicating a lack of authorization even if the client id and secret are correct, and this instance I don't think we should be retrying.

dotasek (changes requested):
# Review

In general this looks good, but there are issues I'd like to see addressed.

Some of my questions may come from ignorance of the OAuth client credentials flow. Please cite relevant areas of the spec when answering these.

I also had Claude do a review, which is also important to discuss, included below. The regression tests Claude came up with are also attached in [OAuthClientCredentialsIntegrationTest.java](https://github.com/user-attachments/files/31316637/OAuthClientCredentialsIntegrationTest.java)

# Claude Review

## Scope and threat model

The PR adds an OAuth 2.0 `client_credentials` grant to `ManagedWebAccess` so the IG Publisher
can survive a terminology-server token expiring mid-run. Two things about it are new to this
codebase:

1. It is the first code path that puts a **long-lived secret in an HTTP request body**
   (`client_secret` in the token POST). Every existing protection in `ManagedHTTPClient` was
   designed around credentials that live in *headers*.
2. It is the first that performs **network I/O from inside an auth provider**, which is why
   `IHTTPAuthenticationProvider.getHeaders` had to be widened to `throws IOException`.

Threat model assumed: `fhir-settings.json` is trusted local config. The adversaries that
matter are

- (a) the configured token endpoint itself — compromised, misconfigured, or hosting an open
  redirect;
- (b) a network attacker on a link where `allowHttp` has been set;
- (c) the resource server, which can now steer client behaviour via its response codes.

The PR is unusually security-aware for its size — `clientSecret` excluded from the generated
`toString()`, `safeOAuthError()` for error bodies, a deliberate `withRetries(0)` on the
secret-bearing request, explicit SSRF reasoning in the javadoc. Most of what follows is where
those good intentions don't actually hold in practice.

Findings already raised in
[review 4987286667](https://github.com/hapifhir/org.hl7.fhir.core/pull/2348#pullrequestreview-4987286667)
(403-retry semantics, `CLIENT_CREDENTIALS` naming, stale javadoc reference, `isBlank`, POJO
encapsulation, chain simplification) are not repeated except where there is a distinct
security angle. Findings are ordered by severity.

---

## 1. `client_secret` is re-POSTed to redirect targets — HIGH

`ManagedHTTPClient.execute()` (`ManagedHTTPClient.java:273-280`) handles 301/302/307/308 by
mutating `uri` and looping, while `requestMethod` and `content` are loop-invariant locals. The
next `buildRequest()` therefore re-sends **the same POST body to the new host**, up to
`MAX_REDIRECTS` (5) hops. For every pre-existing caller the body was non-sensitive; for
`HTTPTokenManager.executeTokenRequest` (`HTTPTokenManager.java:216-227`) the body is
`grant_type=client_credentials&client_id=…&client_secret=…`.

Consequences:

- A token endpoint that 302s — compromise, open redirect, or a lazy "we moved" config — hands
  the client secret to an arbitrary third-party host.
- `TokenEndpointPolicy.canProvideHeaders()` returns `true` for **every** URL
  (`HTTPTokenManager.java:192-194`), so `authCanHandle` is true on redirect hops too. That means
  `isProtocolAllowed()` is consulted instead of
  `ManagedWebAccessUtils.throwExceptionIfNotAllowedScheme()` — and `isProtocolAllowed` permits
  plain `http` whenever the *resource* server has `allowHttp: true`. So the redirect target can
  be `http://attacker/`, i.e. cleartext secret on the wire.
- The design comment at `HTTPTokenManager.java:169-177` reasons carefully about granting the
  token endpoint exactly the trust configured for its server, but nothing pins that trust to the
  *configured* token endpoint URL. `isProtocolAllowed` / `isPrivateNetworkAllowed` answer for any
  URL they are handed.

Note the asymmetry, which the PR relies on elsewhere: the resource request **is** protected
against exactly this, because `ServerDetailsPOJOHTTPAuthProvider.canProvideHeaders()`
re-evaluates per hop and returns `false` off-origin, so the `Authorization` header is dropped.
The token request has no equivalent protection because its credential is in the body, which is
never re-evaluated.

**Fix.** RFC 6749 §3.2 token endpoints do not redirect. Do not follow redirects for the token
request at all — treat a 3xx from the token endpoint as an error. Options:

- Plumb a `followRedirects` / `maxRedirects(0)` flag through `ManagedWebAccessor` alongside the
  new `withRetries(0)`, and set it for the token request; or
- Have `TokenEndpointPolicy` scope itself to the configured endpoint —
  `isProtocolAllowed` / `isPrivateNetworkAllowed` / `canProvideHeaders` return `false` unless
  `ManagedWebAccessUtils.urlMatchesOrigin(url, server.getTokenEndpoint())`. Worth doing
  regardless (see #3), but on its own it still lets the secret reach a *same-origin* redirect
  target, so prefer both.

Regression test: MockWebServer token endpoint returns `302` pointing at a second MockWebServer →
assert server #2 receives no request and the call fails.

---

## 2. `withRetries(0)` does not actually prevent the secret being re-sent — MEDIUM

`HTTPTokenManager.java:218-222` sets `withRetries(0)` with the explicit rationale "a retry
re-sends the client secret". Two ways that guarantee fails:

**(a) `RetryInterceptor.java:63`.** After the do-while exits, the interceptor ends with

```java
return response != null ? response : chain.proceed(request);
```

With `maxRetry = 0` and a connection-level `IOException`, `response` is null and the loop exits
immediately (`retryCounter = 1 <= 0` is false) — so that trailing `chain.proceed(request)`
**sends the request a second time**. The secret goes out twice on any transport failure,
regardless of the retry setting. This is pre-existing code, but this PR is the first to depend
on `retries == 0` as a security property.

**(b) MANAGED access policy.** `ManagedWebAccessor.post()` (`ManagedWebAccessor.java:89-96`) only
consults `retries` on the `DIRECT` branch (via `setupManagedHTTPClient`, lines 55-57). Under
`WebAccessPolicy.MANAGED` the call is delegated to the host-supplied `IWebAccessor`, which never
sees the retry setting — and also never applies `isProtocolAllowed`, the SSRF address-range
checks, or `explicitlyBlockedHosts`. The javadoc at `HTTPTokenManager.java:210-214` asserts the
token request is "subject to the same controls as every other outbound request: the
`WebAccessPolicy` …, the allowed-paths policy, protocol allow-listing, and SSRF address-range
blocking." True for DIRECT; false for MANAGED.

**Fix.** Fix `RetryInterceptor` so it does not re-issue when the retry budget is exhausted
(return a synthetic failure, or rethrow the last `IOException`). Correct the `HTTPTokenManager`
javadoc to scope its claims to DIRECT, and state explicitly that under MANAGED the client secret
is handed to the host application's accessor.

---

## 3. Token requests skip the blocked-host allow-list — MEDIUM

Because `TokenEndpointPolicy.canProvideHeaders()` is unconditionally `true`, the `authCanHandle`
branch at `ManagedHTTPClient.java:251-257` runs and `throwExceptionIfNotAllowedScheme(uri)` never
does. That check is the only enforcement point for `ManagedWebAccessUtils.explicitlyBlockedHosts`
(`metadata.amazonaws.com`, `metadata.google.internal`) and for the "URL has no host" guard.
Token-endpoint requests — and, per #1, their redirect targets — therefore bypass the
cloud-metadata hostname blocklist entirely.

The IP-range checks still catch these when they resolve into link-local space, so this is
defence-in-depth rather than an open hole — **except** when the server sets
`allowPrivateNetwork: true`, which sets `skipNonPublicAddressCheck` and removes the DNS guard
too.

**Fix.** Scoping `TokenEndpointPolicy` to the configured `tokenEndpoint` origin (see #1) restores
the blanket check for everything else. Independently, hoist the `explicitlyBlockedHosts` and
empty-host checks out of `throwExceptionIfNotAllowedScheme` so they apply on both branches of
`ManagedHTTPClient.execute`.

---

## 4. `client_secret` may be sent over cleartext HTTP — MEDIUM

`TokenEndpointPolicy.isProtocolAllowed` (`HTTPTokenManager.java:186-189`) allows `http` for the
token endpoint whenever the resource server has `allowHttp: true`. `allowHttp` is documented on
`ServerDetailsPOJO.java:44-47` as being for "internal servers (e.g. Docker service names) that
don't support HTTPS" — a substantially narrower grant than "it is fine to transmit a long-lived
client secret in the clear". RFC 6749 §2.3.1 requires TLS for client authentication at the token
endpoint.

**Fix.** Preferred: require `https` for `tokenEndpoint` unconditionally. If a plaintext option is
genuinely needed for local development, gate it on its own explicit field (e.g.
`allowHttpTokenEndpoint`) and log a warning naming the risk, rather than inheriting `allowHttp`.

---

## 5. `"(body suppressed)"` is defeated by the chained cause — MEDIUM

`HTTPTokenManager.java:237-243`:

```java
throw new IOException("Token endpoint " + tokenEndpoint
  + " returned a non-JSON response (body suppressed)", e);
```

The chained `Exception e` is the Gson failure, and Gson embeds the offending content in its
message. `JsonElement.getAsJsonObject()` throws `IllegalStateException("Not a JSON Object: " +
this)` — so a token endpoint returning a bare JSON string, or any single-token body that
`JsonParser`'s lenient mode accepts, puts that body verbatim into the cause's message. Any
handler that logs the stack trace, or calls `ExceptionUtils.getRootCauseMessage`, prints exactly
what the message claims to suppress.

`safeOAuthError` (lines 273-281) is otherwise correct and is a good pattern — this is the one
place the pattern is undermined.

**Fix.** Do not chain `e`; log the class name only (`e.getClass().getSimpleName()`), or
`throw new IOException(msg)` with no cause. Add a test asserting no substring of a
sensitive-looking body appears in `ExceptionUtils.getStackTrace(thrown)`.

---

## 6. 401/403 retry replays non-idempotent bodies, and amplifies — MEDIUM

The new branch at `ManagedHTTPClient.java:281-292` applies to **all** HTTP methods. A `POST` or
`PUT` that draws a 403 is re-sent in full. The existing review already questions whether 403
should trigger a retry at all; three additional angles:

- **Duplicate writes.** Nothing distinguishes "credential stale" from "request accepted, then an
  authorization check failed downstream". Any side effect of the first attempt happens twice.
- **Multiplicative amplification with `RetryInterceptor`.** A 401 is neither `isSuccessful()` nor
  `isRedirect()`, so `RetryInterceptor` already retries it once — with a 2 s sleep — at the
  default `retries = 1`. The new hop-level retry then runs the whole thing again:
  **4 requests to the server plus 1 token fetch per logical request**, with 2 s of sleep. A
  resource server that blanket-403s turns every client request into that.
- **Cache-wide invalidation.** `HTTPTokenManager.invalidateToken` clears the entry keyed on
  `tokenEndpoint|clientId`, which is shared by every configured server using that client. One
  server's 403 discards the token all of them are using. The per-key lock and double-check in
  `getToken` (`HTTPTokenManager.java:99-109`) do prevent a fetch stampede, but concurrent threads
  still stall on that lock across a network round trip (up to the 15 s default timeout).

**Fix.** Restrict the retry to `401` (the authentication signal) and to idempotent methods
(`GET` / `HEAD` / `OPTIONS`), or require the caller to opt in per request. Teach `RetryInterceptor`
to leave 401/403 alone so the two mechanisms stop compounding.

---

## 7. Fail-open config normalization — LOW/MEDIUM

`FhirSettings.normalizeSettings` (`FhirSettings.java:245-253`) demotes an incomplete
`client_credentials` entry to `authenticationType: "none"`. The rationale in the javadoc — don't
discard the whole settings file over one bad entry, since `getInstance()` substitutes an empty
POJO on throw — is sound. But the chosen fallback means a typo in `clientSecret` results in
requests going to the configured server **with no credentials at all**, rather than failing. The
`allowHttp` and `allowPrivateNetwork` grants on that entry stay in force.

**Fix.** Prefer a fail-closed sentinel: leave `authenticationType` as `client_credentials` and let
`HTTPTokenManager.requireClientCredentialsFields` (`HTTPTokenManager.java:156-166`, already
present and correct) reject the request at point of use. That `IOException` is per-request, so it
preserves the "don't discard the file" property while failing closed.

---

## 8. Full request URL logged at WARN on every 401/403 — LOW

`ManagedHTTPClient.java:289` logs `url` (a `java.net.URL`) on every credential refresh.
`URL.toString()` includes the query string and any `userinfo`. FHIR terminology requests carry
codes and value-set canonicals in the query, some deployments pass API keys as query params, and
a configured server URL may embed `user:password@`.

**Fix.** Log scheme + host + path only, or reuse whatever redaction the project already applies in
`LoggingInterceptor`.

---

## 9. Unchecked exception escapes `getHeaders()`, contrary to its own contract — LOW

`IHTTPAuthenticationProvider.getHeaders` was widened to `throws IOException` specifically so that
"an unchecked exception here would instead escape those handlers"
(`IHTTPAuthenticationProvider.java:27-32`). Two paths violate that:

- `HTTPTokenManager.java:245` → `JsonUtilities.str` → `JsonElement.getAsString()`, which throws
  `UnsupportedOperationException` for a JSON object or array. A token endpoint returning
  `{"access_token": {}}` propagates an unchecked exception out of `getHeaders`, past
  `buildRequest`, past every `IOException` handler.
- The `catch (Exception)` at lines 239-243 covers only the parse call; line 245 sits outside it.

**Fix.** Validate `json.get("access_token").isJsonPrimitive()` before reading it and throw
`IOException` otherwise; extend the existing `catch (Exception)` to cover the whole parse block.

---

## 10. Hardening notes — LOW

- **Unbounded `expires_in`.** `HTTPTokenManager.java:250-266` accepts any positive `int`, so
  `expires_in: 2147483647` pins a token in cache for ~68 years. Clamp to a sane ceiling (24 h).
- **Unbounded token response.** `ManagedHTTPClient.toResult` reads `response.body().bytes()` with
  no cap, then hands the whole thing to Gson. A hostile token endpoint can exhaust heap.
  Pre-existing for all callers; worth a size limit on the token path at least.
- **Cache survives settings reload.** `cache` and `locks` are JVM-global statics
  (`HTTPTokenManager.java:51-54`) and `ManagedWebAccess.applySettings` replaces
  `serverDetailsList` / `defaultAuthenticationProvider` without evicting them. A rotated or
  revoked `clientSecret` keeps working from cache until the old token expires. Call
  `HTTPTokenManager.clearCache()` from `applySettings`.
- **Cache key omits the secret.** `getCacheKey` (`HTTPTokenManager.java:129-132`) is
  `tokenEndpoint|clientId`. Two entries sharing those but differing in `clientSecret` share a
  token — whichever fetched first wins. The existing comment already flags scope/audience; add
  the secret (hashed) or the server URL.
- **`clearCache()` is public static** with no access control — any code on the classpath can force
  a global token refetch. Consider package-private plus `@VisibleForTesting`.
- **Non-monotonic clock.** `CachedToken.isExpiringSoon` uses `System.currentTimeMillis()`; an NTP
  step can make an expired token look valid, or expire a good one early. `System.nanoTime()` is
  monotonic.
- **`@ToString.Exclude` coverage.** `ServerDetailsPOJO.java:62` correctly excludes `clientSecret`
  — good — but `password`, `token`, and `apikey` are still in the generated `toString()`.
  Pre-existing, but worth fixing in the same pass since the class is already open.
- **Static `headers` can override `Authorization`.** `ServerDetailsPOJOHTTPAuthProvider` applies
  `serverDetails.getHeaders()` *after* the auth switch (lines 82-84), so a configured static
  header silently replaces the freshly-minted bearer token. Pre-existing; now applies to the new
  mode too.

---

## Things checked and found sound

- `Utilities.URLEncode` delegates to `URLEncoder.encode(s, "UTF-8")`, which correctly escapes `&`
  and `=`. No form-parameter injection via a secret containing those characters.
- `HTTPAuthenticationProviderChain.invalidateCachedCredentials` stops at the first provider whose
  `canProvideHeaders` matches, consistent with `getHeaders` — correct, and the reasoning in the
  comment is right.
- `invalidateToken` returning `false` when nothing was cached correctly prevents an
  invalidate-refetch-retry loop against a server that rejects freshly-minted tokens.
- `credentialsRefreshed` is a single boolean per `execute()`, and the `redirects--` compensation
  cannot be repeated, so the redirect/refresh interleaving is bounded at 6 hops. No infinite loop.
- Widening `ManagedWebAccessUtils.urlMatchesOrigin`'s catch to `IllegalArgumentException` fails
  closed (returns "no match" → no headers). Safe.
- `requireClientCredentialsFields` correctly covers the `loadFromSettings` /
  `loadFromFHIRSettings(overrides)` paths that bypass `normalizeSettings`.
- Server entries from `loadFromFHIRSettings(overrides)` are *appended*, and
  `ManagedWebAccessUtils.getServer` returns the first match — so `fhir-settings.json` entries win
  over override-supplied ones. That ordering is protective; worth keeping deliberately.
- Nothing logs the access token or the client secret.

---

## Suggested regression tests

The existing `OAuthClientCredentialsIntegrationTest` already has the MockWebServer harness.

1. Token endpoint returns `302` to a second MockWebServer → assert server #2 receives **no**
   request, and the call throws. (#1)
2. Resource server with `allowHttp: true` and an `http://` `tokenEndpoint` → assert rejected.
   (#4)
3. Token endpoint returns a bare JSON string body → assert
   `ExceptionUtils.getStackTrace(thrown)` contains no substring of that body. (#5)
4. Token endpoint returns `{"access_token": {}}` → assert `IOException`, not
   `UnsupportedOperationException`. (#9)
5. Resource server returns `403` to a `POST` → assert the body is submitted exactly once. (#6)
6. `ManagedWebAccess.loadFromSettings` after a token has been cached → assert a fresh token fetch
   occurs. (#10)
7. Token endpoint closes the connection (transport `IOException`) with `withRetries(0)` → assert
   exactly one request was received. (#2a)

Manual: point the IG Publisher at a Keycloak or Hydra instance with a short token lifetime,
confirm the refresh path works, and grep the run log for the secret.

PR #2350, opened 2026-03-13
Remove FTPClient, tests, and supporting dependencies

dotasek:
The FTPClient class is no longer used and causes intermittent test failures. This removes it.


grahamegrieve (approved):
great

PR #2352, opened 2026-03-16
2026 03 gg snapshot fix

grahamegrieve:


dotasek (approved):
Looks OK to me. This should pass tests before it is merged, though.

PR #2358, opened 2026-03-20
Skip non-existent files in ZipGenerator.addFolder to handle broken sy…

costateixeira:
…mlinks

File.list() can return entries for broken symlinks that don't resolve to actual files, causing FileNotFoundException when zipping IG build directories.

Fix authored by Claude - please review

### DIFF
--- a/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/ZipGenerator.java
+++ b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/ZipGenerator.java
@@ -103,7 +103,11 @@ public class ZipGenerator {
 	  String files[] = fd.list();
 	  for (String f : files) {
 	    if (!".DS_Store".equals(f)) {
-	      if (ManagedFileAccess.csfile(Utilities.path(actualDir, f)).isDirectory())
+	      File cf = ManagedFileAccess.csfile(Utilities.path(actualDir, f));
+	      if (!cf.exists()) {
+	        continue;
+	      }
+	      if (cf.isDirectory())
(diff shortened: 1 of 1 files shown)


dotasek (changes requested):
Since the intent for this PR is to handle symlinks, I would like it if we actually verified that the file in question is a symlink in this logic:

```
Files.isSymbolicLink(Path)
```

It could also benefit from a `log.warn` message when this happens, or a descriptive error if the file isn't a symbolic link and doesn't exist (unless that's somehow an expected and valid thing to be encountered).

costateixeira:
I think this is not intended to start handling symlinks. The issue is that when there is a symlink and it is broken, there is an npe. So i thought symlinks that are broken are just not to have an npe, to unblock builds. And then we could see what we want for supporting symlinks or not

dotasek:
OK, if we just want to unblock builds, the logic is fine, but I would still like to see a log.warn indicating what was skipped because it did not exist.

grahamegrieve:
Can you point me at the build that failed — a log or stack trace, and ideally what the folder had in it?

I can't work out where the symlink would come from. Two of the three callers zip
`pf.outputDir`, which the publisher creates and fills itself, and the third zips a
`FileUtils.copyDirectory` clone of the IG source — and that copy opens a stream per file,
so a broken symlink would kill the clone well before the zip. Neither core nor the IG
publisher ever creates a symlink. And `FileUtilities.copyDirectory` in core has the same
`list()` / `csfile()` / `isDirectory()` shape as `addFolder`, so anything with a broken
symlink in it should have failed on the way *into* the output directory.

given the lack of obvious solution, I'd like to understand the source better

Issue #2360, opened 2026-03-21
StructureMapUtilities fails for transitive imports and `value[x]` assignment in anonymous backbone supplementation


jurmous:
While investigating hapifhir/hapi-fhir#7690, the underlying problem turned out to be in the upstream R4/R4B/R5 StructureMap runtime, not just in HAPI.

There are two related bugs:

1. `StructureMapUtilities.resolveGroupReference(...)` only resolves named groups from direct imports, not transitive imports.
2. `StructureMapUtilities.processTarget(...)` fails for element-model targets when assigning a non-primitive datatype into a choice element like `Extension.value[x]` during anonymous backbone supplementation.

## Symptoms

- A dependent group such as `CopyHumanName` is not found when it exists only in an imported map of an imported map.
- A rule such as:

```text
src -> tgt.value = create('Reference') as t then CopyReference(src, t)
```

fails when `tgt` is an `Extension` being materialized under an anonymous/supplemented backbone target.

## Expected

- Named groups should resolve through the full transitive import graph.
- Assigning `Reference` into `Extension.value[x]` should materialize correctly in the element model.

PR #2361, opened 2026-03-21
Fix StructureMapUtilities failures  by updating the StructureMap runtime in R4, R4B, and R5.

jurmous:
This addresses two related runtime problems:
- named groups were only resolved from direct imports, not through transitive imports
- element-model targets could fail when assigning a non-primitive datatype into a choice element such as `Extension.value[x]` during anonymous backbone supplementation

This issue was identified while investigating the downstream HAPI report `hapifhir/hapi-fhir#7690`.

## Problem
`StructureMapUtilities.resolveGroupReference(...)` stopped at the first import layer. That meant a map could not invoke a group that existed only in an imported map of an imported map.

`StructureMapUtilities.processTarget(...)` also mishandled some element-model assignments when the produced value was a non-primitive datatype. In practice this broke cases such as assigning a `Reference` into `Extension.value[x]` while materializing an extension under an anonymously supplemented backbone element.

## Changes
- add recursive group lookup across the full import graph, with cycle protection and duplicate-match detection preserved
- normalize converted non-primitive datatype values before calling `setProperty(...)` on element-model targets
- apply the fix consistently in:
  - `org.hl7.fhir.r4`
  - `org.hl7.fhir.r4b`
  - `org.hl7.fhir.r5`
- add regression tests for:
  - transitive imported group resolution
  - anonymous backbone supplementation with `Extension.valueReference`

## Impact
After this change:
- dependent groups such as `CopyHumanName` resolve correctly through transitive imports
- assignments like `src -> tgt.value = create('Reference') as t then CopyReference(src, t)` materialize correctly in the element model for anonymous backbone supplementation scenarios

## Testing
Added focused regression coverage for R4, R4B, and R5.

## References
- `hapifhir/org.hl7.fhir.core#2360`
- `hapifhir/hapi-fhir#7690`

### DIFF
--- a/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/StructureMapUtilities.java
+++ b/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/StructureMapUtilities.java
@@ -53,6 +53,7 @@ import org.hl7.fhir.r4.conformance.ProfileUtilities.ProfileKnowledgeProvider;
 import org.hl7.fhir.r4.context.IWorkerContext;
 import org.hl7.fhir.r4.context.IWorkerContext.ValidationResult;
 import org.hl7.fhir.r4.elementmodel.Element;
+import org.hl7.fhir.r4.elementmodel.ObjectConverter;
 import org.hl7.fhir.r4.elementmodel.Property;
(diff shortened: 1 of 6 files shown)


jurmous:
How to get feedback on this? Is there something I can provide to help in the review?

PR #2364, opened 2026-03-23
Match servers using URL objects instead of string logic

dotasek:
Uses java.net.URL for matching servers.

### DIFF
--- a/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/PackageServerHTTPAuthProvider.java
+++ b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/PackageServerHTTPAuthProvider.java
@@ -26,7 +26,7 @@ public class PackageServerHTTPAuthProvider implements IHTTPAuthenticationProvide
   public Map<String, String> getHeaders(URL url) {
     if (!ManagedWebAccessUtils.urlMatchesOrigin( url, this.url)) {
       // We should not get here unless there is an error in the client using this provider
-      throw new IllegalArgumentException(("Calling code is attempting to access headers for " + this.url + " with a request to " + url));
+      throw new IllegalArgumentException(("Unexpected request for access headers for " + this.url + " with a request for " + url));
     }
     Map<String, String> headers = new HashMap<>();
     switch (server.getAuthenticationMode()) {


org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/ManagedWebAccessUtils.java
```diff
@@ -23,14 +25,31 @@ public static ServerDetailsPOJO getServer(Iterable<String> serverTypes, String u
...
+      return urlMatchesOrigin(requestUrl, serverUrl);
+    } catch (MalformedURLException e) {
+      return false;
+    }
+  }
+
+  public static boolean urlMatchesOrigin(URL requestUrl, URL serverUrl) {
+    return requestUrl.getProtocol().equals(serverUrl.getProtocol())
+      && requestUrl.getHost().equals(serverUrl.getHost())
+      && requestUrl.getPort() == serverUrl.getPort()
```
grahamegrieve: I'm troubled by the port and path bit here. What if server doesn't have a path? I have to specify the password differently for SSL and not SSL?

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/PackageServerHTTPAuthProvider.java
```diff
@@ -19,12 +19,14 @@ public PackageServerHTTPAuthProvider(PackageServer server) throws MalformedURLEx
...
   public boolean canProvideHeaders(URL url) {
-    return this.url.getProtocol().equals(url.getProtocol())
-      && this.url.getHost().equals(url.getHost());
+    return ManagedWebAccessUtils.urlMatchesOrigin( url, this.url);
   }
 
   @Override
   public Map<String, String> getHeaders(URL url) {
+    if (!ManagedWebAccessUtils.urlMatchesOrigin( url, this.url)) {
+      throw new SecurityException("Requested authorization headers for " + this.url + " for unmatched URL: " + url);
```
grahamegrieve: so when does this happen? if you request headers, you just don't get any? When/why is it an error?

dotasek: I do not see this happening the way we presently use it. It is there defensively however, because the packageClient can inherit configured servers, and we want strictness around which of these it gets auth details from.

grahamegrieve: Then the error message should reflect that this is defensive and unexpected, not a user error

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/ManagedWebAccessUtils.java
dotasek: I expanded the test cases to include some examples of paths / no paths. I think the logic there is pretty reasonable.

For http/https I certainly think we should have very explicit settings, per server and protocol. It's less complex in the code and requires more config if you do something suspect, and I think we want that. 

I have, however, included some inference rules for ports and protocols in the last PR, which I think eases a few common cases.

PR #2367, opened 2026-03-24
Ensure that conformance is only inferred for CapabilityStatements if …

lmckenzi:
…a flag is set.


grahamegrieve:
but the tests fail now?

PR #2371, opened 2026-03-25
Fix maven build in VSCode

dotasek:
Some part of the maven configuration is not being loaded correctly in VSCode, resulting in the project always attempting to build with Java 8.

The pom.xml change reverts changes that were appear to be an unfinished attempt to use kotlin in this project:

https://github.com/hapifhir/org.hl7.fhir.core/pull/490


lmckenzi:
This seems to work on Eclipse for me

PR #2398, opened 2026-04-13
Increase testCachedCanonicalResourceGetWithMultiThread timeout

dotasek:
The timeout on testCachedCanonicalResourceGetWithMultiThread is currently being hit on our pipelines and causing test failures. It still executes in a reasonable timespan on local machines (1.0s in most cases), so this seems likely to be a Azure VM issue.


dotasek:
Fixes #2402

grahamegrieve (approved):
ok

grahamegrieve:
Seems like whatever it was fixed itself, but might as well merge this anyway

PR #2399, opened 2026-04-13
Fix sqlite native access warnings

dotasek:
When run in later Java versions, sqlite will produce warnings regarding native access.

The following thread includes instructions from xerial on how to rectify this in the config, which will be applied here: https://github.com/xerial/sqlite-jdbc/issues/1289

Additional note:

These changes also introduce new testing for Java 25. This revealed that DateTimeUtil.toHumanDisplay no longer functions with the deprecated COMPAT provider for Java locale and now produces different output for different versions of Java. Since 'toHumanDisplay' is defined in the API as "using the system local format" this is expected, and the tests are now Java version specific.


dotasek:
This is a fix for https://github.com/hapifhir/org.hl7.fhir.core/issues/2401

grahamegrieve (approved):
ok

PR #2403, opened 2026-04-14
Add RegexTimeout utilities class

dotasek:
This utility class executes common regular expression methods and times out if processing takes longer than expected.

FHIRPathEngine in all versions will now use this with the default timeout.


grahamegrieve (approved):
ok

PR #2405, opened 2026-04-18
Fix missing assets and malformed icon URLs in comparison output

costateixeira:
Two independent one-liners in the comparison/hierarchical-table path that together caused broken icons and 404s in the ig-comparison output of the validator's `compare` command:

- HierarchicalTableGenerator.srcFor: use Utilities.pathURL instead of raw string concat so an imagePath without a trailing slash (as passed by the comparison renderers) does not yield `ig-comparisonicon.gif`.

- Utilities.isProhibitedBinaryFile: predicate was inverted. It returned true for everything NOT in EXCLUDED_FILES, so ComparisonRenderer.dumpBinaries skipped every binary except the ones meant to be excluded (fhir.css, tbl_*.png, icon_*.png were dropped while togaf.png / xver-paths-*.json were the only files copied).

Before: 
image

after:
image


grahamegrieve (approved):
thanks

PR #2406, opened 2026-04-20
Fhir path html checks

lmckenzi:
https://jira.hl7.org/browse/FHIR-56312 - Allow htmlChecks to be invoked on string, not just xhtml.  Also, fixed issue where htmlChecks was returning 'false' when invoked on unsupported types.  According to the spec, it's supposed to return empty.


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java
```diff
@@ -4813,9 +4814,20 @@ private List<Base> funcJoin(ExecutionContext context, List<Base> focus, Expressi
   private List<Base> funcHtmlChecks1(ExecutionContext context, List<Base> focus, ExpressionNode exp) throws FHIRException {
     // todo: actually check the HTML
```
dotasek: It looks to me like this todo is now done.

dotasek (changes requested):
This appears to work for the FHIRPathEngine, but:

1. It requires a bump of the fhir-test-cases version to 1.7.60-SNAPSHOT
2. It doesn't implement anything for liquid-tests.json, which has been modified in the fhir-test-cases and fails for this PR.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java
dotasek: Resolved with merge from master and https://github.com/hapifhir/org.hl7.fhir.core/pull/2406/commits/3bed4a34f7ae13ed59111887b954af47df1f76f8

dotasek (approved):

dotasek (changes requested):
This now fails tests because some of our test resources:

ValidationEngineTests
ValidationTests

As far as I can tell, these are all due to your change:

> Unexpected Issue found: error/invariant @ StructureDefinition.text.div [6,76]: Constraint failed: txt-1: 'The narrative SHALL contain only the basic html formatting elements and attributes described in chapters 7-11 (except section 4 of chapter 9) and 15 of the HTML 4.0 standard, <a> elements (either name or href), images and internally contained style attributes' (defined in http://hl7.org/fhir/StructureDefinition/Narrative) (context: http://hl7.org/fhir/StructureDefinition/StructureDefinition|5.0.0)

Could you take a look at these and decide if the new validation logic is faulty, or if these are actual failures?

For example, `validator/ext-stds-status-normative.json` has what I believe is valid HTML content:

```json
"text" : {
    "status" : "extensions",
    "div" : "<div xmlns=\"http://www.w3.org/1999/xhtml\">Placeholder</div>"
  }
```

dotasek (approved):
Looks like the latest changes get passing tests. Thanks.

PR #2409, opened 2026-04-22
2026 04 gg preferred bindings

grahamegrieve:


org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
```diff
@@ -1843,32 +1841,39 @@ else if (!noExtensibleWarnings)
...
-//        (Utilities.listValueStartsWith("http://hl7.org/fhir", systems) || Utilities.listValueStartsWith("http://terminology.hl7.org", systems))) {
-//      return org.hl7.fhir.r5.model.OperationOutcome.IssueSeverity.ERROR;
-//    } else if (strength == BindingStrength.REQUIRED || strength == BindingStrength.EXTENSIBLE) {
-//      return org.hl7.fhir.r5.model.OperationOutcome.IssueSeverity.WARNING;
-//    } else {
-//      return org.hl7.fhir.r5.model.OperationOutcome.IssueSeverity.INFORMATION;      
-//    }
-//  }
+  // We don't want to create issues about preferred bindings from the base spec, only from derived profiles
+  private boolean isBaseBinding(StructureDefinition profile, ElementDefinition definition) {
```
dotasek, line 1845: profile is unused here. If that's OK, it should be removed from the method.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
```diff
@@ -1843,32 +1841,39 @@ else if (!noExtensibleWarnings)
...
+    if (definition == null) {
+      // this means it's an additional binding, and the base spec never puts preferred bindings in additional bindings
+      return false;
+    }
+    String basePath = definition.getPath();
+    String baseType = basePath.contains(".") ? basePath.substring(0, basePath.indexOf(".")) : basePath;
+    StructureDefinition baseProfile = context.fetchResource(StructureDefinition.class, baseType);
```
dotasek: fetchResource is deprecated, we should use the one with VersionResolutionRules.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
```diff
@@ -2143,7 +2148,7 @@ else if (strength == BindingStrength.EXTENSIBLE) {
               else if (!noExtensibleWarnings)
                 txWarningForLaterRemoval(element, errors, NO_RULE_DATE, vr.getTxLink(), vr.getDiagnostics(), IssueType.CODEINVALID, element.line(), element.col(), path, false, I18nConstants.TERMINOLOGY_TX_CONFIRM_5, describeReference(vsRef.primitiveValue(), valueset, bc, usageNote));
             } else if (strength == BindingStrength.PREFERRED) {
-              if (baseOnly) {
+              if (isBaseBinding(profile, definition) || settings.isForPublication()) {
```
dotasek: This is different from the other uses, and based on the comment for `isBaseBinding` shouldn't it be !isBaseBinding?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
```diff
@@ -2155,7 +2160,7 @@ else if (strength == BindingStrength.EXTENSIBLE) {
             else
               txWarning(errors, NO_RULE_DATE, vr.getTxLink(), vr.getDiagnostics(), IssueType.CODEINVALID, element.line(), element.col(), path, false, I18nConstants.TERMINOLOGY_TX_NOVALID_5, describeReference(vsRef.primitiveValue(), valueset, bc, usageNote), (vr.getMessage() != null ? " (error message = " + vr.getMessage() + ")" : ""), system+"#"+code);
           } else if (strength == BindingStrength.PREFERRED) {
-            if (baseOnly) {
+            if (isBaseBinding(profile, definition) || settings.isForPublication()) {
```
dotasek: Same comment as above, shouldn't this be `!isBaseBinding`?

org.hl7.fhir.validation/src/test/java/org/hl7/fhir/terminology/tests/ExternalTerminologyServiceTests.java
```diff
@@ -31,17 +35,57 @@
...
+        @Override
+        public void schedule(Runnable child) {
+          pool.submit(child);
+        }
+
+        @Override
+        public void finished() {
+          pool.shutdown();
+          try {
+            pool.awaitTermination(1, TimeUnit.HOURS);
```
dotasek: Maybe a shorter timeout than an hour? This whole thing runs in 10 seconds for me. A few minutes would be generous and still keep our pipelines responsive.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/TxTester.java
```diff
@@ -48,20 +49,32 @@
...
+      this.result = result;
+    }
+
+    public ResultInformation(String message) {
+      this.message = message;
+      this.result = false;
+    }
+  }
+
   public static class IntHolder {
```
dotasek: Nitpick: this looks like it could be an AtomicInteger instead of implementing our own counter.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/TxTester.java
```diff
@@ -408,43 +449,76 @@ private ITerminologyClient connectToServer(Set<String> modes) throws URISyntaxEx
...
+   * Eagerly perform the one-shot, not-thread-safe setup: connect to the server
+   * from the calling thread, fetch the CapabilityStatement and
+   * TerminologyCapabilities, and stash the modes so worker threads can
+   * lazily construct their own clients on first dispatch. Call this once on
+   * the main thread before issuing any concurrent executeTest calls.
+   * Idempotent — subsequent calls only refresh the stored modes.
+   */
+  public void initialise(Set<String> modes) throws URISyntaxException, IOException {
+    this.clientModes = modes;
+    if (capabilityStatement == null) {
```
dotasek, line 468: ThreadLocal has a safer way of doing this using withInitial(), which will create the value on the first call of get().

If we use this client() can just be replaced with terminologyClient.get()

```java
// Modes passed to the first initialise() call, cached so worker threads can
  // lazily construct their own clients with the same connection config.
  private volatile Set<String> clientModes;

  // Per-thread ITerminologyClient. Every operation mutates client state
  // (setAcceptLanguage, setClientHeaders) before dispatch; giving each worker
  // its own client makes those mutations private to that worker and removes
  // the need for cross-thread synchronisation around the dispatch itself.
  // Lazily created on first use per thread via client(); the one-time
  // capability-statement fetch runs against the main-thread bootstrap client
  // inside initialise().
  private final ThreadLocal<ITerminologyClient> terminologyClient = ThreadLocal.withInitial(() -> {
    Set<String> modes = clientModes;
    if (modes == null) {
      throw new IllegalStateException("TxTester not initialised — call initialise(modes) before executeTest()");
    }
    try {
      return connectToServer(modes);
    } catch (URISyntaxException | IOException e) {
      throw new RuntimeException("Unable to initialize client");
    }
  });
```

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
```diff
@@ -1843,32 +1841,39 @@ else if (!noExtensibleWarnings)
...
+  private boolean isBaseBinding(ElementDefinition base, ElementDefinition definition) {
+    if (base == null || !base.hasBinding()) {
+      return false;
+    }
+    ElementDefinitionBindingComponent baseBinding = base.getBinding();
+    ElementDefinitionBindingComponent definitionBinding = definition.getBinding();
+    if (baseBinding.getStrength() != definitionBinding.getStrength()) {
+      return false;
+    }
+    ValueSet baseVS = context.fetchResource(ValueSet.class, baseBinding.getValueSet());
```
dotasek: Same deprecation warnings for this and the following line.

dotasek (changes requested):

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
grahamegrieve: well, it's unclear. I rewrote it several times, and 2 of three implementations used profile. I'm not convinced that I'm done with that method yet either

grahamegrieve: no I do use it

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
grahamegrieve: fixed

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
grahamegrieve: fixed

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
grahamegrieve: fixed

org.hl7.fhir.validation/src/test/java/org/hl7/fhir/terminology/tests/ExternalTerminologyServiceTests.java
grahamegrieve: ok

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/TxTester.java
grahamegrieve: wasn't a nitpick. needed to be thread safe now

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/TxTester.java
grahamegrieve: maybe but I don't care in testing code. it works, so I'm not fixing it.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
grahamegrieve: also fixed

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
dotasek: Where? IntelliJ finds no references in that method. Only baseProfile shows a partial text match in that method.

grahamegrieve: did you get my updates?

dotasek: I see no updates since 17 hours ago on this branch, here or via git pulls.

grahamegrieve: I failed to hit enter

dotasek (approved):

PR #2410, opened 2026-04-22
Add _include/_revinclude parameters to OpenAPI search operations

mrunibe:
Add `_include` and `_revinclude` parameters to the generated OpenAPI spec based on `searchInclude/searchRevInclude` in the CapabilityStatement. This makes the OAS spec more explicit about which include targets a server supports, giving API consumers a machine-readable contract.

A capability statement with:
```
...
"searchInclude" : ["ServiceRequest:requester",
                   "ServiceRequest:subject"]
...
```

will result in a corresponding `_include` query parameter with the allowed values as an enum (and equivalently for `_revinclude`):
```
{
  "name": "_include",
  "in": "query",
  "description": "Request that the server include additional resources related to the primary search results",
  "schema": {
    "type": "array",
    "items": {
      "type": "string",
      "enum": [
        "ServiceRequest:requester",
        "ServiceRequest:subject"
      ]
    }
  }
}
```
swagger


dotasek (changes requested):
This loads in SwaggerUI with some modification, and appears to produce valid OpenAPI with `r5/openapi/cs-base.json`.

However, if I load this OpenAPI spec in SwaggerUI (with some modifications to get it to load) and try to build a query using the new '_include' param, I get the following:

```
http://0.0.0.0:8080/Account?type=patient&_include=Account.owner
```

But the syntax for _include/_revinclude should use `:`, not `.` between `Account` and `owner`: https://build.fhir.org/search.html#_revinclude

This seems wrong. Or am I misreading the spec?

Also, if you could find it in your heart to include in this PR, in both the r4b and r5 OpenApiGenerator, the following line results in a 404:

`resp.content("application/fhir+xml").schemaRef(specRef()+"/CapabilityStatement.xsd");`

It should just be:

`resp.content("application/fhir+xml").schemaRef(specRef()+"/capabilitystatement.xsd");`

mrunibe:
Thanks for looking into this, @dotasek !

> But the syntax for _include/_revinclude should use `:`, not `.` between `Account` and `owner`: https://build.fhir.org/search.html#_revinclude
> 
> This seems wrong. Or am I misreading the spec?

Yes, the correct query should indeed be `&_include=Account:owner`. However, I think that is because the input cs-base.json you used for testing is not correct in fhir-test-cases:

https://github.com/FHIR/fhir-test-cases/blob/e4365d37ad7875e8584badf9d8933048336cfb5e/r5/openapi/cs-base.json#L84:

`"searchInclude" : ["Account.owner",`

I would assume this should be `"searchInclude" : ["Account:owner",` instead.
Should we raise a separate issue in fhir-test to get that changed? I think it would be wrong to work around that in the OpenApiGenerator.

When I run a test using swagger with the example quoted in the PR, I get:

swagger-req


Note: I added a fix for the lowercase `/capabilitystatement.xsd` to this PR: https://github.com/hapifhir/org.hl7.fhir.core/pull/2410/changes/ea36bb31186b99dbdb4230e05243a7bc8cf68e73

dotasek:
@mrunibe It appears that you are correct. The docs for capabilitystatement even include an example with the correct syntax: https://build.fhir.org/capabilitystatement-definitions.html#CapabilityStatement.rest.resource.searchInclude

> These values would be listed as part of capabilities for "CarePlan":

> `"searchInclude" : ["CarePlan:activity-reference:DeviceRequest","DeviceRequest:device"]`

I can set up a PR on fhir-test-cases myself and test that it will work correctly with this PR.

dotasek:
@mrunibe I have created the following PR: https://github.com/FHIR/fhir-test-cases/pull/273

With that in place, I now get the following formatting: `http://0.0.0.0:8080/Account?_include=Account%3Aowner`

This format is parsed correctly by HAPI's test server.

Once that is merged, please update this PR to make sure it uses the most recent version of the fhir-test-cases, and I will approve and merge.

mrunibe:
> Once that is merged, please update this PR to make sure it uses the most recent version of the fhir-test-cases, and I will approve and merge.

Thanks @dotasek

Updated PR:
- Rebased onto master, so the PR now picks up the latest 1.7.63-SNAPSHOT of fhir-test-cases (incl. FHIR/fhir-test-cases#273)
- Repointed the r4b OpenAPI test to load cs-r4.json from r4b/openapi (per the rename in FHIR/fhir-test-cases#273)
- **NOTE**: I also removed the local copy of cs-r4.json under src/test/resources — a leftover from #2278 that was likely missed when its r5 sibling was cleaned up in 0960f2d

PR #2412, opened 2026-04-24
2026 04 gg coding rendering

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/CodeSystemRenderer.java
```diff
@@ -589,8 +589,9 @@ private void addDefineRowToTable(RenderingStatus status, XhtmlNode t, ConceptDef
         for (ConceptPropertyComponent pcv : pcvl) {
           if (pcv.hasValue()) {
             if (first) first = false; else td.addText(", ");
-            if (pcv.hasValueCoding()) { 
-              td.addText(pcv.getValueCoding().getCode());
+            if (pcv.hasValueCoding()) {
+              ResourceWrapper cc = ResourceWrapper.forType(context.getContextUtilities(), pcv.getValueCoding());
+              renderCoding(status, td, cc, false);
```
dotasek, line 594: Not directly related to this PR, but `status` in renderCoding is never used.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/DataRenderer.java
```diff
@@ -357,11 +357,19 @@ protected String describeLang(String lang) {
         for (ConceptReferenceDesignationComponent cd : l.getDesignation()) { 
           if (cd.getLanguage().equals(lang)) 
             nativelang = cd.getValue(); 
-        } 
-        if (nativelang == null) 
-          return en+" ("+lang+")"; 
-        else 
-          return nativelang+" ("+en+", "+lang+")"; 
+        }
+        if (lang != null) {
```
dotasek, line 361: Lang can't be null here; is it possible that we need to check for isEmpty or isBlank?

org.hl7.fhir.validation.cli/src/main/java/org/hl7/fhir/validation/cli/logging/LogColorConverter.java
```diff
@@ -7,6 +7,14 @@
 
 public class LogColorConverter  extends ForegroundCompositeConverterBase<ILoggingEvent> {
 
+  @Override
```
dotasek: The commit for this change seems to have gone in before this fix to tty:

https://github.com/hapifhir/org.hl7.fhir.core/pull/2386/changes

The PR above is the recommended fix for terminal colour compatibility. I suspect this change is no longer doing anything, and might introduce some hard to debug logging issues down the road.

dotasek (changes requested):

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/CodeSystemRenderer.java
grahamegrieve: it's a standard api for everything in that class. I might get around to using it in the future.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/DataRenderer.java
grahamegrieve: why can't it be null? Just because if the check above? I'm not sure which line is in error

org.hl7.fhir.validation.cli/src/main/java/org/hl7/fhir/validation/cli/logging/LogColorConverter.java
grahamegrieve: fixed

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/DataRenderer.java
dotasek: IntelliJ is saying it can't be null at that point.

Were it null, line 353 would throw a NullPointerException before 361.

dotasek: Yeah, that's it. If I do a `lang != null && lang.contains("-")` at 334 and 353, it no longer complains.

dotasek (approved):
Approved from a code-review perspective.

This should pass tests correctly before merge, though.

PR #2413, opened 2026-04-24
Manage encoding 

dotasek:
This makes the encoding for the parent pom and all modules explicitly UTF-8.

It also revises some pluginManagement to ensure uniform usage of plugin versions without parameters.


dotasek:
Closing, as this is included in https://github.com/hapifhir/org.hl7.fhir.core/pull/2433

PR #2415, opened 2026-04-25
Fix SHCParser related test failures on some OSs

dotasek:
In addition, this fixes an fhir.org related issue that was due to a new https certificate and includes some fixes for SonarQube reported issues (no functional changes, refactor only)


grahamegrieve (approved):
ok

PR #2418, opened 2026-04-29
Liquid enhancements2

lmckenzi:
Enhanced the Liquid engine to support the full Liquid spec (with the exception of the Where and Map filters).  See also https://github.com/FHIR/fhir-test-cases/pull/265 which has complete tests for all of the new functionality (aligned with the tests in the Liquid spec).


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/liquid/LiquidEngine.java
```diff
@@ -303,69 +449,741 @@ private class LiquidStatement extends LiquidNode {
...
+      try {
+        return Double.parseDouble(s.trim());
+      } catch (NumberFormatException e) {
+        return 0;
+      }
+    }
+
+    private boolean isInteger(String s) {
+      try {
+        Long.parseLong(s.trim());
```
dotasek, line 891: This evaluation conflicts with the method name. Liquid appears to consider Integer to be the same as a Java int, so should we not check for that instead?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/liquid/LiquidEngine.java
```diff
@@ -303,69 +449,741 @@ private class LiquidStatement extends LiquidNode {
...
+          t = liquify(ctxt, engine.evaluate(ctxt, resource, resource, resource, i.expression));
         } else switch (i.filter) {
+
+        // ── String filters ──────────────────────────────────────────────
+
+        case APPEND:
+          t = List.of(singleString(t) + liquifySingle(ctxt, engine.evaluate(ctxt, resource, resource, resource, i.expression)));
+          break;
+        case BASE64_DECODE:
+          t = List.of(new String(java.util.Base64.getDecoder().decode(singleString(t)), java.nio.charset.StandardCharsets.UTF_8));
```
dotasek, line 506: I get a warning here, and for all subsequent references to `t`, that indicate it could be null and cause an exception. There must be some point above where we can enforce it having a value.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/liquid/LiquidEngine.java
```diff
@@ -513,15 +1335,15 @@ public void evaluate(StringBuilder b, Base resource, LiquidEngineContext ctxt) t
             break;
           }          
           LiquidforloopObject forloop = new LiquidforloopObject(list.size(), i, offset, limit, parentLoop);
-          lctxt.globalVars.put("forloop", forloop);
+          lctxt.globalVars.put("forloop", List.of(forloop));
           if (lctxt.globalVars.containsKey(varName)) {
```
dotasek: I think the containsKey isn't needed here anymore; the remove will work regardless.

dotasek (changes requested):
See notes.

All tests pass at this point (when run with the appropriate test branch), and the new code is covered with the exception of the new markdown implementation. Is that intended to also be under test?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/liquid/LiquidEngine.java
lmckenzi: It should probably be renamed to isNumber.  The Liquid syntax refers to "Integer" some places, but the examples always include at least one float.  Do you want me to rename or do you want to?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/liquid/LiquidEngine.java
lmckenzi: Ok.  I didn't change that bit, but I'm fine for the check to be removed if it's no longer necessary.  Personally, I think from a readability perspective, the check acknowledges that "this might not be here", while a straight removal implies we're always expecting it to exist.  Does it make much difference to performance?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/liquid/LiquidEngine.java
lmckenzi: How can it be null?  We're creating a String right here.  Does `new String(null) return null?  The only situation singleString(t) could return null is if the previous t collection contained a null item, and that shouldn't be possible.  We could throw an exception in singleString if that happens I guess.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/liquid/LiquidEngine.java
dotasek: Probably a tiny improvement to performance. It's a complaint from Sonarqube, which I've been trying to pay attention to more.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/liquid/LiquidEngine.java
dotasek: I will rename it shortly and review what that does while I do it.

dotasek: Well, hold on if it's 'isNumber', shouldn't we also change this do Double.parseDouble(...)?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/liquid/LiquidEngine.java
dotasek: It's because singleString(t) has no null check, and will fail due to a null pointer.

lmckenzi: Ok.  I'm happy to throw an exception if the input is null.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/liquid/LiquidEngine.java
lmckenzi: Umm.  It seems like the answer should be "yes", though I'm curious how the Liquid tests are passing with the code as it is...

dotasek: > Umm. It seems like the answer should be "yes", though I'm curious how the Liquid tests are passing with the code as it is...

We're lucky?

Another library implemented it like so (https://github.com/bkiers/Liqp/blob/1f0c47e87053f937fde5448cab0963f3379ce4a3/src/main/java/liqp/LValue.java#L421):

```java
// valid Long?
        if(String.valueOf(value).trim().matches("\\d+")) {
            return true;
        }

        try {
            // valid Double?
            Double.parseDouble(String.valueOf(value).trim());
        } catch(Exception e) {
            return false;
        }
```

That doesn't look too shabby. I'd have removed the regex for a parseLong, but it prevents having to catch an exception, and I can live with that.

lmckenzi: Doesn't the next clause catch the exception anyhow?  I'm not seeing a return in the second clause though.  Presume there's a default return true; after the quoted code.

In any event, I'm not overly fussed.

dotasek: Nope. We're explicitly checking to see if this is a Long. See:

https://github.com/hapifhir/org.hl7.fhir.core/blob/97f07178ea723ff9620d10fa58c488bfab3205bc/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/liquid/LiquidEngine.java#L704

There and other places, this is just misnamed. I updated, and added a null check while I was there.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/liquid/LiquidEngine.java
dotasek: Maybe it's more graceful if singleString(null) returns "", just like an empty list?

What's our desired behaviour here for for this and 'isLong'? Throw a FHIRException, or return "" and `false` respectively?

lmckenzi: We should throw an exception.  If there's null in there, it means something's gone badly wrong.

dotasek: I'm going to use LIQUID_SYNTAX_UNEXPECTED for the message.

dotasek: Or LIQUID_UNKNOWN_SYNTAX?

dotasek (approved):
Approved pending @lmckenzi 's review of the new exceptions I added.

lmckenzi:
I'm fine with those changes

PR #2419, opened 2026-04-30
Profile-based test data generation: requiredOnly + literal mapping values + GITB-aligned REST API for ITB integration

costateixeira:
This PR has two parts:

A. Adds a **`requiredOnly`** generation mode and richer mapping inputs to `ProfileBasedFactory` so it can produce minimal-but-valid resources from any (profiled) StructureDefinition.

B. A set of GITB-faithful REST endpoints under `/itb/*` so the FHIR validator can be used as a backend for the [Interoperability Test Bed (ITB)](https://www.itb.ec.europa.eu/) test engine.

## A. `requiredOnly` generation mode + richer mappings (`ProfileBasedFactory`)

The factory previously generated every optional element it had base data for — producing kitchen-sink resources to assert against and sometimes invalid (e.g. truncated Base64 photos). This PR adds a minimal-generation mode and a richer mapping format.

1. **`requiredOnly` flag on `ProfileBasedFactory`.** Only required elements (`min ≥ 1`) are populated at the resource top level. Children of required complex elements (CodeableConcept, Coding, Reference, …) are still populated normally — otherwise the parent ends up empty and gets pruned, breaking its `min=1` cardinality.

2. **`hasMappingFor` keeps optional elements that have a descendant mapping.** Without this, a caller who maps `clinicalStatus.coding` would also have to map the `clinicalStatus` parent (an optional CodeableConcept) just to keep `requiredOnly` from skipping it. Instead, any mapping under an element's path is enough to keep that element.

3. **Mapping entries accept a literal `value` field** alongside the existing FHIRPath `expression`. Applies to primitive entries and `parts` items. Why a separate field?
   - FHIRPath uses single quotes for string literals, so callers passing a plain URL have to write `"expression": "'http://...'"` — knowing FHIRPath syntax for a value that isn't FHIRPath-y.
   - Hosts that also use single quotes as their string delimiter (e.g. the GITB/ITB TDL engine) collide with the FHIRPath quoting and reject the mappings outright.

   `value` is used as-is, no evaluation, no quoting. `expression` is still preferred when the value is genuinely computed.

### Files

- `org.hl7.fhir.r5/.../testfactory/ProfileBasedFactory.java` — feature implementation
- `org.hl7.fhir.validation/.../ValidationEngine.java` — `generateTestData()` gains a `requiredOnly` parameter that threads through to the factory
- `org.hl7.fhir.validation/src/test/.../ProfileBasedFactoryTests.java` *(new)* — JUnit 5 tests that go through `ValidationEngine.generateTestData()`:
  - `requiredOnly` skips optional elements (Patient → just `resourceType`)
  - `requiredOnly` populates required References (`AllergyIntolerance.patient` has a `.reference`)
  - Literal `value` field on a primitive mapping is used as-is
  - Mapping for a descendant path (`...clinicalStatus.coding`) keeps the optional parent element

## B. GITB-aligned REST API (`/itb/*`)

A new HTTP layer mounted under `/itb/*` exposes the validator as a set of GITB validation/processing services (TAR-shaped responses, AnyContent inputs, `getModuleDefinition` discovery). Test cases written in GITB TDL can call the validator directly without a custom adapter. The TestDataGenerator endpoint exposes the `requiredOnly` flag from (A) as an optional input.

### Endpoints

| Service | Path | Operation(s) |
|---|---|---|
| FHIRValidator | `/itb/fhir/validate` | validate resource against base + profiles |
| MatchetypeValidator | `/itb/matchetype/validate` | structural pattern match |
| FHIRPathAssertion | `/itb/fhirPathAssertion/validate` | boolean FHIRPath assertion |
| FHIRPathProcessor | `/itb/fhirPath/process` | evaluate FHIRPath, return value |
| TestDataGenerator | `/itb/testdata/process` | generate / generateBundle from a profile (with `requiredOnly`) |
| ValidationResultsProcessor | `/itb/validationResults/process` | summarize / filterBySeverity / filterByText on an OperationOutcome |
| IGManager | `/itb/igManager/process` | loadIG (`hl7.fhir.x.y#vN.M`) |

### New code (under `org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/http/`)

- `GitbServiceHandler` — common base (request decoding, AnyContent helpers, OpenAPI-style `getModuleDefinition`)
- `GitbValidationServiceHandler`, `GitbProcessingServiceHandler` — base classes for the two GITB service shapes
- `GitbFhirHandler`, `GitbMatchetypeHandler`, `GitbFhirPathAssertionHandler` — validation services
- `GitbFhirPathHandler`, `GitbTestDataHandler`, `GitbIgManagerHandler`, `GitbValidationResultsHandler` — processing services
- `GitbTarBuilder` — TAR (Test Assertion Report) construction, including hidden context items (`errorCount`, `warningCount`, `informationCount`, `severity`) and the OperationOutcome embedded as `forReport=false` AnyContent so callers can post-process

### Spec & docs

- `documentation/itb-rest-spec.md` — full request/response contract for each endpoint, including TAR shape, AnyContent semantics, IG loading, and worked examples


costateixeira:
@grahamegrieve I have tested this in an ITB instance. Some new parts of this will surely need iteration or review, like the package generation. They are working - but I am sure we will add features and corrections.
This supports ITB integration and allows further testing, which is the main goal of this PR.

costateixeira:
Splitting this into reviewable pieces rather than asking anyone to read it as one change. Converting to a draft in the meantime — the branch behind it is still where I work, so it will keep moving.

The library changes are now separate PRs, each independently mergeable and each with its own tests:

| | |
| --- | --- |
| #2606 | `ProfileBasedManipulator` — apply set/add/remove operations to a resource |
| #2607 | Literal `value` in test data mappings, and reject invalid Base64 |
| #2608 | `requiredOnly` test data generation mode *(after #2607)* |
| #2609 | Fix inverted binding strength in `QuestionnaireBuilder` answer constraints |
| #2610 | `GET /questionnaire` |
| #2611 | `GET /package`, the CRMI `$package` operation |
| #2612 | `POST /fml` |
| #2613 | `POST /loadResource` *(after #2612)* |
| #2614 | StructureMap `dependent` conversion test, R4 ↔ R5 |

The `/itb` GITB REST surface — the largest part of this PR — is #2615. It depends on six of the
above, so it is a draft until they land.

Two things worth pulling out of the original diff, since they are behaviour changes rather than
additions and are easy to miss at this size:

- **#2609 fixes an inverted condition.** `constraintTypeForBinding` had `!=` where it wanted `==`,
  so a *required* binding produced `optionsOrType` and an *example* binding produced `optionsOnly`
  — backwards in both directions. This changes the output of every Questionnaire generated from a
  profile with bindings.
- **#2611 lets `ResourceDependencyWalker` walk a `StructureMap`.** Previously that threw
  `Error("Resource StructureMap not Processed yet")`, so this replaces a failure rather than
  altering behaviour that worked.

Nothing else in the split changes existing behaviour.

costateixeira:
@dotasek https://github.com/hapifhir/org.hl7.fhir.core/pull/2419#issuecomment-5532339926 this shows why this is draft and not to be reviewed

PR #2420, opened 2026-05-01
Regex Checkstyle checking

dotasek:
Add checkstyle entries to checkstyle.xml that scan for:

* Usages of Pattern.compile(...) and Pattern.matches(...) (plus static imports of these)
* Usages of String.matches(...), String.replaceAll(...), String.replaceFirst(...), and String.split(...), which implicitly use the aforementioned methods.

As a result, MANY regex usages were identified and analyzed using Claude Sonnet 4.6, and checkstyle suppressions were added with commentary.

A few user-supplied regexs were also spotted through this analysis, and changed to use the RegexTimeout utility, notably:

https://github.com/hapifhir/org.hl7.fhir.core/blob/5beab54186836f0d9a89ac6bbe561f025c90747e/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/advisor/RulesDrivenPolicyAdvisor.java#L124
https://github.com/hapifhir/org.hl7.fhir.core/blob/5b61c71eb5668c7d81322c8784814ff0e07c02a0/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/scoring/ScoringEngine.java#L188
https://github.com/hapifhir/org.hl7.fhir.core/blob/5beab54186836f0d9a89ac6bbe561f025c90747e/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java#L1026


grahamegrieve (approved):
ok

PR #2421, opened 2026-05-03
toDateTime() fhirpath implementation

oliveregger:
I would like to use the toDateTime() fhirpath function as specified in https://hl7.org/fhirpath/#todatetime-datetime, however it is not implemented in java (it is in the javascript fhir.js library).

attached the pr for the java implementation, test-cases for the functionality have been created in https://github.com/FHIR/fhir-test-cases/pull/269


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java
```diff
@@ -5219,11 +5219,51 @@ private List<Base> funcToQuantity(ExecutionContext context, List<Base> focus, Ex
...
+      result.add(item);
+    } else if (item instanceof DateType) {
+      // a date is a valid datetime
+      result.add(new DateTimeType(((DateType) item).getValueAsString()).noExtensions());
+    } else if (item instanceof StringType) {
+      String s = convertToString(item);
+      // Same FHIRPath/FHIR datetime regex as funcIsDateTime so that
+      // toDateTime() and convertsToDateTime() agree on what is convertible.
+      if (s != null && s.matches(
+          "([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)"
```
dotasek: This regex is repeated multiple times in FHIRPathEngine.

Since we now collect regexs in org.hl7.fhir.utilities.regex.RegexConstants, I would like to see this, and the two other instances of the constant moved there:

```java
public static final String DATE_TIME_REGEX =
      "([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|1[0-2])(-(0[1-9]|[1-2][0-9]|3[0-1])(T([01][0-9]|2[0-3])(:[0-5][0-9](:([0-5][0-9]|60))?)?(\\.[0-9]+)?(Z|(\\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00))?)?)?)?";
```

dotasek (changes requested):
Aside from the one comment regarding moving the regex to a constant, this looks good to me. The tests are appropriate.

dotasek (approved):
Looks good, I will merge after tests complete. Thank you!

PR #2422, opened 2026-05-04
Fix all maven warnings in build

dotasek:
Our maven build contained numerous warnings regarding misconfigurations and empty jars. 

This gives us a clean build (with the exception of animal-sniffer related signature errors)


grahamegrieve (approved):
ok

PR #2425, opened 2026-05-06
Check for terminologyClientManager being null in CompliesWithChecker

dotasek:
The new getTerminologyClientManager() method forces IWorkerContext implementations to use or extend TerminologyClientManager, which is our own implementation.

The only use of getTerminologyClientManager() is in code that explicitly fails with an Error().

Until this is better thought out, we should allow a `null` value provided by `getTerminologyClientManager()` to skip the unimplemented code, and note its incompleteness.


grahamegrieve (approved):
ok

PR #2426, opened 2026-05-06
2026 05 gg version changes

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/TerminologyClientManager.java
```diff
@@ -69,7 +69,7 @@ public void replace(String src, String dst) {
 
     @Override
     public String toString() {
-      return "auth = " + CommaSeparatedStringBuilder.join("|", authoritative)+ ", candidates=" + CommaSeparatedStringBuilder.join("|", candidates);
+      return "["+url+ "]: auth = " + CommaSeparatedStringBuilder.join("|", authoritative)+ ", candidates=" + CommaSeparatedStringBuilder.join("|", candidates);
```
dotasek: Can we make this `url: authoritative = `?

Seeing `auth` and a String return looks like we're logging authentication, which looks scary in review.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/model/PackageInformation.java
```diff
@@ -18,6 +20,13 @@ public class PackageInformation {
   private List<String> dependencies = new ArrayList<>();
   private String fhirVersion;
 
+  /**
+   * this marks the package as the very first package loaded, the master definitions
+   * these are the ones that provide type definitions, and if there's a version-less request
+   * for any of these, the master ones override any other version.
+   */
+  @Getter @Setter private boolean isMaster;
```
dotasek: This doesn't look right to me. Every other parameter in this class is private and only set in the constructor, and `isMaster` should follow that pattern.

Considering that there's 9 fields, it this class would ideally be a builder, so something like this would happen:

```java
PackageInformation packageInfo = PackageInformation.builder().fromRequiredFields(id, date, fhirVersion).canonical(canonical).asMaster().build()
```

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/IWorkerContextManager.java
dotasek, line 1: The method overloads here are multiplying, which makes this class harder to change later on. I think it is worth considering moving the responsibility of 'isMaster' onto something like a MasterNpmPackage decorator, which would extend NpmPackage. Then, most of the methods will stay the same with the exception of those that just use combinations of id and version, like int loadPackage(String idAndVer, boolean isMaster).

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/IWorkerContextManager.java
```diff
@@ -94,30 +91,43 @@ interface ICanonicalResourceLocator {
    * note that the package system may use lazy loading; the loader will be called later when the classes that use the context need the relevant resource
    *
    * @param pi - the package to load
-   * @param loader - an implemenation of IContextResourceLoader that knows how to read the resources in the package (e.g. for the appropriate version).
+   * @param loader - an implementation of IContextResourceLoader that knows how to read the resources in the package (e.g. for the appropriate version).
+   * @param isMaster - marks that the package being loaded is the very first package loaded, the master definitions
    * @return the number of resources loaded
    */
-  int loadFromPackage(NpmPackage pi, IContextResourceLoader loader) throws FileNotFoundException, IOException, FHIRException;
+  int loadFromPackage(NpmPackage pi, IContextResourceLoader loader, boolean isMaster) throws IOException, FHIRException;
```
dotasek: `pi` should be `npmPackage`

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/IWorkerContextManager.java
```diff
@@ -94,30 +91,43 @@ interface ICanonicalResourceLocator {
...
    *
    * note that the package system may use lazy loading; the loader will be called later when the classes that use the context need the relevant resource
    *
    * @param pi - the package to load
+   * @param isMaster - marks that the package being loaded is the very first package loaded, the master definitions
    * @return the number of resources loaded
    */
-  int loadPackage(NpmPackage pi) throws FileNotFoundException, IOException, FHIRException;
+  int loadPackage(NpmPackage pi, boolean isMaster) throws IOException, FHIRException;
+  default int loadPackage(NpmPackage pi) throws IOException, FHIRException {
```
dotasek: `pi` should be `npmPackage`

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/IWorkerContextManager.java
```diff
@@ -128,10 +138,12 @@ interface ICanonicalResourceLocator {
    * This method also loads all the packages that the package depends on (recursively)
    *
    * @param pi - the package to load
-   * @param loader - an implemenation of IContextResourceLoader that knows how to read the resources in the package (e.g. for the appropriate version).
+   * @param loader - an implementation of IContextResourceLoader that knows how to read the resources in the package (e.g. for the appropriate version).
    * @param pcm - used to find and load additional dependencies
    * @return the number of resources loaded
    */
-  int loadFromPackageAndDependencies(NpmPackage pi, IContextResourceLoader loader, BasePackageCacheManager pcm) throws FileNotFoundException, IOException, FHIRException;
+  int loadFromPackageAndDependencies(NpmPackage pi, IContextResourceLoader loader, BasePackageCacheManager pcm) throws IOException, FHIRException;
```
dotasek: `pi` should be `npmPackage`

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/HTMLClientLogger.java
```diff
@@ -82,7 +82,7 @@ public void logRequest(String method, String url, List<String> headers, byte[] b
       }
     }
     if (body != null) {
-      if (body.length > 100000) {
+      if (body.length > 1000000) {
         body = Arrays.copyOf(body, 100000);
```
dotasek: Should we not copy the same length we're checking for above?

dotasek (changes requested):

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/HTMLClientLogger.java
grahamegrieve: we should

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/IWorkerContextManager.java
grahamegrieve: I don't understand why that would be good, or why it would be easier to change, just because the overloads are elsewhere

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/IWorkerContextManager.java
grahamegrieve: ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/model/PackageInformation.java
grahamegrieve: ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/TerminologyClientManager.java
grahamegrieve: ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/IWorkerContextManager.java
dotasek: OK, I checked for the impact in HAPI for example, and it doesn't seem to have as much impact as I thought it would.

dotasek (approved):

PR #2427, opened 2026-05-08
SonarQube: Fix repetition of XML constant + SonarQube naming issues in OpenEHR

dotasek:
SonarQube pointed out that the generated method names in OpenehrXmlParser could potentially be confused with those in its parent class.

This fix documents the potential issue (and suppresses SonarQube rule matching for the class)

It also replaces several constants that were used as literals that were contained in XmlConstants.


grahamegrieve:
how does this have anything to do with generated method names?

dotasek:
> how does this have anything to do with generated method names?

A bunch of methods have names that only differ in capitalization from similar methods in the class.

For example, OpenehrXmlParser#parseADDRESS, which is either generated automatically or otherwise constructed from the actual OpenEHR spec, looks too similar to XmlParser#parseAddress for SonarQube's liking. One extends the other, so both methods exist in OpenehrXmlParser. This just comments those methods to emphasize care around using them and adds suppression to keep SonarQube from continuing to complain.

grahamegrieve (approved):
ok

PR #2433, opened 2026-05-13
Check file encoding for NPM tar entries

dotasek:
Some legacy FHIR packages use outdated single-bit file encodings (such as ISO_8859_1) for files. 

This attempts to use UTF-8, the preferred encoding, but falls back to ISO_8859_1 when necessary.

This fixes issues where single bit encodings would be automatically replaced with `?` characters in some Windows environments.


grahamegrieve (approved):
ok

PR #2437, opened 2026-05-14
FHIRPath execution limits

dotasek:
This adds execution/iteration limits to two methods in FHIRPathEngine across all modules: `execute` and `funcRepeat`.

These are set to defaults of 50 in this PR. This limit should be evaluated in review.


dotasek:
Temporarily closing after discussion with @grahamegrieve . Will re-open if still applicable, or create another PR.

PR #2441, opened 2026-05-18
Whitelist all valid fields within ViewDefinition validation

### LINKED ISSUE #2440: SQL on FHIR Validator rejects ViewDefinition fields defined by the spec's logical model
The SQL on FHIR `Validator` (`org.hl7.fhir.r5.utils.sql.Validator` and its r4 mirror) uses an explicit allow-list per `checkProperties(...)` call. The allow-list is missing properties defined by the [ViewDefinition logical model](https://sql-on-fhir.org/ig/StructureDefinition-ViewDefinition.html) ([FSH source](https://github.com/FHIR/sql-on-fhir-v2/blob/master/input/fsh/models.fsh)), so valid ViewDefinitions generated by SUSHI from the spec's own examples fail with `Unknown JSON property ...`.

Missing properties:

- Top level (`Validator.java:93`, r4 at line 91): `profile`, `fhirVersion`, plus `id`, `meta`, `text`, `language`, `implicitRules`, `contained`, `modifierExtension` (inherited from `CanonicalResource` -> `DomainResource` -> `Resource`; SUSHI emits `id` on every Instance).
- Select (`Validator.java:160`): `repeat`.

The column, constant, and where allow-lists appear correct.

**To reproduce:** build the [SQL on FHIR IG](https://github.com/FHIR/sql-on-fhir-v2) with `viewDefinition` parameters registered in `sushi-config.yaml`. Every registered view fails with `Unknown JSON property id` and `View Definition is not valid`, leaving downstream `{% sql %}` Liquid tags reporting `no such table: <view_name>`.

johngrimes:
The SQL on FHIR Validator's `checkProperties` allow-lists (r4 and r5) were rejecting valid ViewDefinition fields declared by the spec's logical model - including `profile`, `fhirVersion`, `repeat`, and the Resource/DomainResource/CanonicalResource inheritance SUSHI emits on every Instance. IGs built from the SQL on FHIR v2 examples failed with `Unknown JSON property ...`, leaving `{% sql %}` queries running against an empty database.

Each of the five `checkProperties` calls is now aligned with the [edge ViewDefinition StructureDefinition](https://build.fhir.org/ig/FHIR/sql-on-fhir-v2/StructureDefinition-ViewDefinition.json). New `ValidatorTests` cover each newly-allowed property plus a negative case for genuinely unknown ones.

Fixes #2440

### DIFF
--- a/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/sql/Validator.java
+++ b/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/sql/Validator.java
@@ -88,7 +88,7 @@ public class Validator {
 
 
   public void checkViewDefinition(String path, JsonObject viewDefinition) {    
-    checkProperties(viewDefinition, path, "resourceType", "url", "identifier", "name", "version", "title", "status", "experimental", "date", "publisher", "contact", "description", "useContext", "copyright", "resource", "constant", "select", "where");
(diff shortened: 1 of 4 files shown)


johngrimes:
Superseded by #2453

PR #2444, opened 2026-05-19
Implement SQL on FHIR repeat directive

### LINKED ISSUE #2443: Add support for the `repeat` directive in the SQL on FHIR ViewDefinition runner
The SQL on FHIR spec defines a `repeat` directive on `select` for recursively flattening unbounded nesting (e.g. `QuestionnaireResponse.item.item...`, `CodeSystem.concept.concept...`). The IG Publisher's runner does not appear to implement it: child column paths under a `repeat` are evaluated against the root resource type, not the type yielded by the repeat expression.

Spec link: https://build.fhir.org/ig/FHIR/sql-on-fhir-v2/StructureDefinition-ViewDefinition.html#recursive-traversal-with-repeat

johngrimes:
Implements the recursive traversal `repeat` directive from the SQL on FHIR spec as a third iteration mode on `select`, alongside `forEach` and `forEachOrNull`. Closes #2443.

Each entry in `repeat` is a FHIRPath expression applied to the focus and recursively to every descendant yielded by any sibling expression, depth-first pre-order. Mutually exclusive with `forEach`/`forEachOrNull`. Type analysis is lenient because a path may be invalid against the starting type but valid against a descendant yielded by a sibling (e.g. `item` with `answer.item`). Applied to R4, R5, and the element-model `ViewDefinitionValidator`. Row comparison in the test runner is also switched to multiset since SoF does not mandate row order.

## Test plan
- [x] R5 `SqlOnFhirRunnerTests`: 137/137 pass, including all 19 `repeat.json` cases
- [x] Existing R4/R5 SoF fixtures not regressed
- [x] Validation module: failures are pre-existing network/terminology connectivity issues, unrelated


johngrimes:
Superseded by #2453

PR #2453, opened 2026-05-20
Enhancements and fixes to SQL on FHIR view runner

### LINKED ISSUE #2440: SQL on FHIR Validator rejects ViewDefinition fields defined by the spec's logical model

### LINKED ISSUE #2442: SQL on FHIR runner reports `dateTime` columns as integer
A `ViewDefinition` column whose path resolves to a `dateTime` aborts with:

```
Error Processing ViewDefinition: Attempt to add a type dateTime to an integer column for column <name>
```

Explicit `type = "dateTime"` on the column does not help.

Likely site: `org.hl7.fhir.r5.utils.sql.Runner.genValue()` case `Integer`. `Validator.checkColumn` -> `kindForType()` appears to set `ColumnKind` to `Integer` for these paths, and the explicit `type` declaration does not override it.

### LINKED ISSUE #2443: Add support for the `repeat` directive in the SQL on FHIR ViewDefinition runner

### LINKED ISSUE #2445: ViewDefinition validator emits false-positive collection warnings for columns under forEach
`org.hl7.fhir.r5.utils.sql.Validator` warns on every 0..1 column under a `forEach` whose expression yields a collection, even though each iteration sees a single value.

## Reproduction

```json
{
  "resourceType": "ViewDefinition",
  "status": "draft",
  "resource": "Patient",
  "select": [{
    "forEach": "address",
    "column": [
      { "path": "use", "name": "use" },
      { "path": "city", "name": "city" }
    ]
  }]
}
```

Each column warns:

> This column is not defined as a collection, but the path statement 'use' might return multiple values for the column 'use' for some inputs

`Address.use` is 0..1, so per row the value is single.

## Cause

In `Validator.checkSelect`, the `TypeDetails` returned by `checkForEach` is passed straight to `checkColumn`. That `TypeDetails` carries `CollectionStatus != SINGLETON` from the parent expression (`address` is 0..* on Patient), and the status propagates into the column's `TypeDetails`, tripping the warning at `Validator.java:369`.

Fix: after `checkForEach`/`checkForEachOrNull`, downgrade the result's `CollectionStatus` to `SINGLETON` before passing it to nested column/select checks - the iteration consumes the parent's collection-ness.

## Side note

For each such warning, `ValidatorUtils.messagesToOutcome` (line 151) also logs `Internal error in location ... unexpected token "$"`, because it feeds resource-file locations like `Binary-Foo.json$.select[0].column[0]` into `fpe.parse(...)`. Independent issue, but doubles the noise.

## Impact

The standard `forEach` unnesting pattern can't be validated cleanly. CI gates that promote warnings to failures (e.g. the SQL on FHIR IG build) break without spurious `collection: true` workarounds.

### LINKED ISSUE #2452: Implement %rowIndex in SQL on FHIR ViewDefinition runner
The SQL on FHIR ViewDefinition runner in `org.hl7.fhir.core` does not appear to support the `%rowIndex` environment variable, so views that reference it cannot be evaluated.

Spec reference: https://build.fhir.org/ig/FHIR/sql-on-fhir-v2/StructureDefinition-ViewDefinition.html#rowindex

Could support for `%rowIndex` be added?

johngrimes:
SQL on FHIR runner and validator updates across r4, r5 and R6 (org.hl7.fhir.services), bundled to avoid inter-PR dependencies.

Runner
- Implement the `repeat` directive as a third iteration mode on `select`.
- Implement `%rowIndex` with an immutable execution context; empty `forEachOrNull` emits a null row so direct columns still evaluate.
- Implement StorageSqlite3 key resolution; fix element-model dateTime column values; strip trailing zeros from r4 StorageJson decimals.
- Fix reference key resolution, type checks, integer and null handling; `join()` returns empty on an empty collection (FHIRPathEngine).

Validator (utils/sql Validator and ViewDefinitionValidator)
- Fix false-positive "might return multiple values" warnings for 0..1 columns under `forEach`.
- Port repeat, unionAll and type-family checks to ViewDefinitionValidator; report unionAll column mismatches at the offending branch.
- Accept same-family primitive types in column checks; report valueInteger64 constants as unsupported in r4.
- Harden input and column checks; whitelist all fields permitted by the ViewDefinition logical model (profile, fhirVersion, repeat, resourceDefinition, inherited CanonicalResource, MetadataResource and DomainResource fields) so valid instances are no longer rejected as "Unknown JSON property".

Tests
- Read the SoF suite from fhir-test-cases instead of per-module `sof/*.json` copies; align scaffolding across the three modules, honour result comparisons, compare rows order-insensitively and numbers by value.
- Add Runner, Validator, StorageJson, StorageSqlite3 and ViewDefinitionValidator unit tests.

Supersedes #2444 and #2441.
Fixes #2440, fixes #2442, fixes #2443, fixes #2445, fixes #2452.

Depends on FHIR/fhir-test-cases#282, which must be merged first.


johngrimes:
@grahamegrieve Would you mind taking a look at this PR?

We need this to finalise the SQL on FHIR IG for the upcoming ballot.

org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/sql/Runner.java
```diff
@@ -190,49 +193,74 @@ public void finish(WorkContext wc) {
...
-    //  } else if (select.has("unionAll")) {
-    //    focus.addAll(executeUnion(select, b));
-
     List<List<Cell>> tempRows = new ArrayList<>();
     tempRows.addAll(rows);
     rows.clear();
 
-    for (Base f : focus) {
-      List<List<Cell>> rowsToAdd = cloneRows(tempRows);  
+    // Save the enclosing %rowIndex so nested iteration scopes do not leak the
```
grahamegrieve: Is this right? it feels like a weak recursion strategy that'll get broken somewhere soon?

johngrimes: Hi @grahamegrieve, I've made some updates to improve this. The code now uses an immutable context that gets passed along with the recursive call.

johngrimes:
Hold off on reviewing/merging this - we've identified another issue that we need to fix that may alter this implementation.

https://github.com/HL7/sql-on-fhir/issues/395

grahamegrieve:
where is this now?

johngrimes:
Ok, I have taken another look at this - this is what I plan to do:

- Rebase onto master (looks like you may have added an R6 implementation, and some of the changes on main partially supersede these?)
- Refresh sof/*.json from sql-on-fhir.js
- Add hasChildren("repeat"), toSingleton() to both checkRepeats
- Whitelist additions related to inheriting from R6 DomainResource and MetadataResource (resourceDefinition, approvalDate, lastReviewDate, effectivePeriod, topic, author, editor, reviewer, endorser, relatedArtifact)
- funcJoin empty handling
- Port checkUnion + familyMatches into ViewDefinitionValidator

johngrimes:
My assumption is that we need the r4, r5 and r6 parallel implementations, so I will make sure that they are all consistent.

I will also rationalise so that all the implementations point to `fhir-test-cases` exclusively for the SQL on FHIR test cases.

johngrimes:
@grahamegrieve Ok, I think I am done with updating this.

Let me know what you think.

grahamegrieve:
Does this deal with any of the open issues?

johngrimes:
Yes - it closes all five open issues with the `sql` label:

- #2440 - Validator rejects ViewDefinition fields defined by the logical model (whitelist additions)
- #2442 - `dateTime` columns reported as integer
- #2443 - `repeat` directive in the runner
- #2445 - false-positive collection warnings for columns under `forEach`
- #2452 - `%rowIndex` in the runner

The description only listed the first three as fixes; I've added #2443 and #2452 so they close on merge. Each is implemented across r4, r5 and R6 (`org.hl7.fhir.services`), and covered by the fhir-test-cases SoF suite (FHIR/fhir-test-cases#282) plus the new unit tests.

grahamegrieve:
The new repeat support in Runner.expandRepeat (r4, r5 and services) recurses with no depth limit and no record of what it has already visited. So a ViewDefinition with "repeat": ["$this"] or ["%resource"] never stops and ends in a StackOverflowError. That's an Error, so it gets past the publisher's catch (Exception e) in generateViewDefinitions and kills the whole build instead of reporting a problem with the view. Something like ["descendants()"] does stop, but its output grows exponentially with nesting depth. Please add a limit: a cap on recursion depth and/or total rows, which throws a FHIRException with a clear message when it's hit. 

Separately, FHIRPath's own repeat() in FHIRPathEngine drops items it has already seen (equalsDeep) and works in a loop rather than recursively, which is what makes it safe. Does the SQL on FHIR spec say anything about whether repeat should also drop items already seen? If it does, doing the same here would fix the cycle problem directly. If it's deliberately silent, or wants duplicates kept, then the limit is the only protection, and it would be worth getting the spec to say so explicitly.

johngrimes:
I've added repeat and total node limits to mitigate this problem. I've checked that the runner can process data approaching these limits within a negligible amount of memory.

I've also created this issue to add wording to the spec to guide implementers to consider limits when implementing this: https://github.com/HL7/sql-on-fhir/issues/404

The repeat directive in SQL on FHIR intentionally does not de-duplicate nodes, as nodes can be legitimately identical and de-duplicating would lead to things like incorrect counts.

PR #2461, opened 2026-05-28
Spotbugs initial work

dotasek:
# SpotBugs PR #1 — Fix always-true / always-false evaluations

This PR addresses a specific subset of SpotBugs High-severity findings: structurally-guaranteed always-true or always-false comparisons due to type mismatches or self-comparisons. Bug categories covered: `EC_UNRELATED_TYPES`, `GC_UNRELATED_TYPES`, `SA_LOCAL_SELF_COMPARISON`, `QBA_QUESTIONABLE_BOOLEAN_ASSIGNMENT`, and `UC_USELESS_CONDITION`.

Starting count of this subset: **115+**. After this PR: **9 remaining** (all deferred with TODO comments, see additional PR comments).


dotasek:
---

## Dependency changes

Added `spotbugs-annotations` (provided scope) to `org.hl7.fhir.r4/pom.xml` and `org.hl7.fhir.validation/pom.xml`. These modules now use `@SuppressFBWarnings` on deprecated classes. The annotation is already in `dependencyManagement` in the root pom; this just declares it in the affected modules.

---

## Batch A — StringBuilder.equals(String) in Turtle.java

**EC_UNRELATED_TYPES**

**Files:**
- `org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/turtle/Turtle.java` (lines 889, 910)
- `org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/formats/Turtle.java` (lines 836, 855) — **deprecated class, suppressed**

`b.equals("\"\"")` and `b.equals("''")` where `b` is a `StringBuilder`. `StringBuilder` inherits `Object.equals()` which compares by identity, not content — always false. Effect: the empty string literal (`""`) and empty single-quoted string (`''`) detection in the Turtle lexer never fired; those boundary cases were skipped.

**Fix (utilities):** `b.equals(...)` → `b.toString().equals(...)`

**Fix (r4 deprecated):** Class-level `@SuppressFBWarnings("EC_UNRELATED_TYPES")` added to the `@Deprecated` class to avoid touching legacy code.

---

## Batch B — String.equals(XhtmlNode) in FHIRPathEngine.checkHtmlNames

**EC_UNRELATED_TYPES**

**Files:**
- `org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/fhirpath/FHIRPathEngine.java` (line 4680)
- `org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java` (line 4972)

`!"p".equals(c)` where `c` is an `XhtmlNode`. `String.equals(XhtmlNode)` is always false, so `!"p".equals(c)` is always true — the `block` parameter was never cleared to `false` when recursing into a `<p>` element. Effect: every HTML descendant was treated as block-level context regardless of whether it was inside a `<p>`.

**Fix:** `!"p".equals(c)` → `!"p".equals(c.getName())`

---

## Batch C — String.equals(StringType) in CodeSystemUtilities.isInactive

**EC_UNRELATED_TYPES**

**Files:**
- `org.hl7.fhir.dstu3/src/main/java/org/hl7/fhir/dstu3/utils/CodeSystemUtilities.java` (line 155)
- `org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/terminologies/CodeSystemUtilities.java` (line 173)
- `org.hl7.fhir.r4b/src/main/java/org/hl7/fhir/r4b/terminologies/CodeSystemUtilities.java` (line 268)

`"inactive".equals(p.getValueStringType())` where `getValueStringType()` returns a FHIR `StringType` object. Comparing a `String` to a `StringType` is always false. Effect: `isInactive()` always returned `false` — every concept appeared active regardless of its `status` property.

**Fix:** `p.getValueStringType()` → `p.getValueStringType().asStringValue()`

---

## Batch D — CommaSeparatedStringBuilder.equals(String) in NarrativeGenerator / CompartmentDefinitionRenderer

**EC_UNRELATED_TYPES**

**Files (non-deprecated):**
- `org.hl7.fhir.dstu3.support/src/main/java/org/hl7/fhir/dstu3/support/utils/NarrativeGenerator.java` (line 3579)
- `org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/NarrativeGenerator.java` (line 4380)
- `org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/CompartmentDefinitionRenderer.java` (line 51)

**File (deprecated):**
- `org.hl7.fhir.r4b/src/main/java/org/hl7/fhir/r4b/renderers/CompartmentDefinitionRenderer.java` (line 50)

`!rules.equals("{def}")` where `rules` is a `CommaSeparatedStringBuilder`. `CommaSeparatedStringBuilder` inherits `Object.equals()` — always false, so `!rules.equals(...)` is always true. The `rules` object is empty at the point of the check (params are appended inside the else-if body), so changing to `rules.toString().equals("{def}")` preserves existing behavior (still always true at that point).

**Fix (non-deprecated):** `!rules.equals("{def}")` → `!rules.toString().equals("{def}")`

**Fix (r4b deprecated):** Class-level `@SuppressFBWarnings("EC_UNRELATED_TYPES")` on the `@Deprecated` class.

---

## Batch E — String.equals(char '1') in QuestionnaireBuilder.buildQuestion

**EC_UNRELATED_TYPES**

**Files:**
- `org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/QuestionnaireBuilder.java` (line 361)
- `org.hl7.fhir.r4b/src/main/java/org/hl7/fhir/r4b/utils/QuestionnaireBuilder.java` (line 369)
- `org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/QuestionnaireBuilder.java` (line 379)

`element.getMax().equals('1')` — `element.getMax()` returns a `String`; `'1'` is a `char` literal autoboxed to `Character`, which can never equal a `String`. Effect: `setRepeats()` was always `true` — every questionnaire item was marked as repeating regardless of max cardinality.

**Fix:** `'1'` → `"1"`

---

## Batch F — String.equals(XhtmlNode) in XhtmlParser.parseElementInner

**EC_UNRELATED_TYPES**

**File:** `org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/XhtmlParser.java` (line 636)

`parents.get(i).getName().equals(n)` where `n` is an `XhtmlNode`. Same pattern as Batch B. Effect: when malformed XHTML has a closing tag that doesn't match the current open tag, the ancestor search to find the right unwind point always failed — `unwindPoint` was never set, and the parser could not recover from mismatched closing tags.

**Fix:** `.equals(n)` → `.equals(n.getName())`

---

## Batch G — Date.equals(String) in QuestionnaireValidator.checkDateOption

**EC_UNRELATED_TYPES**

**File:** `org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/QuestionnaireValidator.java` (line 1514)

`item.getValue().equals(v.primitiveValue())` — `DateType.getValue()` returns `java.util.Date`; `v.primitiveValue()` returns a FHIR date string. Comparing them is always false. Effect: every date value in a questionnaire answer was flagged as invalid regardless of whether it matched an allowed date option.

Note: the second occurrence of the same pattern at line 1547 uses `TimeType` where `getValue()` returns `String` — that one is correct and was left unchanged.

**Fix:** `item.getValue()` → `item.getValueAsString()`

---

## Batch H — vsu.equals(vsu) self-comparison in CodeSystemValidator.validateCodeSystem

**SA_LOCAL_SELF_COMPARISON**

**File:** `org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/CodeSystemValidator.java` (line 139)

`vsu.equals(vsu)` — a variable compared to itself is always `true`. Effect: the "supplement's declared valueSet URL doesn't match the base CS's valueSet URL" warning fired unconditionally for every supplement that has a valueSet, rather than only when the URLs differ.

**Fix:** `vsu.equals(vsu)` → `vsu.equals(csB.getValueSet())`

---

## Batch I — int key on Map\<String,\_\> in ICFImporter.doImport

**GC_UNRELATED_TYPES**

**File:** `org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/misc/ICFImporter.java` (lines 94–95)

`codes.containsKey(i)` / `codes.remove(i)` where `i` is `int` and `codes` is `Map<String, ConceptDefinitionComponent>`. An autoboxed `Integer` key never matches a `String` key — the parent-concept cleanup loop was always a no-op. Note: even after the fix, the cleanup is semantically wrong (ICF codes are strings like `"b110"`, not `"4"` or `"5"`); a TODO comment documents this for future correction.

**Fix:** `codes.containsKey(i)` + `codes.remove(i)` → `codes.remove(String.valueOf(i))` (combined, `containsKey` eliminated as redundant)

---

## Batch J — Map.values().remove(T) in CanonicalResourceManager.drop

**GC_UNRELATED_TYPES — deferred, TODO comment only**

**File:** `org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/CanonicalResourceManager.java` (lines 611–612)

`listForId.values().remove(cr)` and `listForUrl.values().remove(cr)` are no-ops: `values()` returns `Collection<List<CachedCanonicalResource<T>>>`, while `cr` is `CachedCanonicalResource<T>` — type mismatch, remove always returns false. The `listForUrl` cleanup is already handled correctly at lines 618–621. TODO comments added; the actual fix (`listForId.get(cr.getId()).remove(cr)`) is deferred.

---

## Batch K — List.indexOf(String) on List\<ColumnSpecifier\> in ValueSetDataProvider.cell

**GC_UNRELATED_TYPES**

**File:** `org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/testfactory/dataprovider/ValueSetDataProvider.java` (line 144)

`columns.indexOf(name)` where `columns` is `List<ColumnSpecifier>` and `name` is a `String`. `indexOf(String)` always returns -1 — so `cell()` always returned `null`. The parallel `columnNames` list (`List<String>`) has the column names in the same order and should be used.

**Fix:** `columns.indexOf(name)` → `columnNames.indexOf(name)`

---

## Batch L — QBA: if(firstA = true) in ProfileUtilities.genTypes

**QBA_QUESTIONABLE_BOOLEAN_ASSIGNMENT**

**Files:**
- `org.hl7.fhir.dstu3.support/src/main/java/org/hl7/fhir/dstu3/support/conformance/ProfileUtilities.java` (lines 1831, 1862)
- `org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/conformance/ProfileUtilities.java` (line 2578)
- `org.hl7.fhir.r4b/src/main/java/org/hl7/fhir/r4b/conformance/ProfileUtilities.java` (line 4120)

`if (firstA = true)` — assignment-in-condition always evaluates to `true`. The `else` branch (which inserts a comma separator between aggregation modes) was never executed. Effect: multiple aggregation modes were displayed as a run-on string without separating commas.

**Fix:** `if (firstA = true)` → `if (firstA)`

---

## Generated code exclusion — dstu2 and dstu2016may model classes

**EC_UNRELATED_TYPES — excluded via filter**

**Modules:** `org.hl7.fhir.dstu2`, `org.hl7.fhir.dstu2016may`

85 EC_UNRELATED_TYPES findings in generated `model.*` classes. The pattern (comparing a `String` argument to a list element in `has*()` methods) is endemic to the generated JAXB/FHIR model code and cannot be fixed without regenerating the model. Added `spotbugs-exclude.xml` filter files (one per module) that exclude `EC_UNRELATED_TYPES` for all classes in the `org.hl7.fhir.dstu2.model.*` and `org.hl7.fhir.dstu2016may.model.*` packages. Wired the filter into each module's `pom.xml` via a `<build>/<plugins>` block.

---

## UC_USELESS_CONDITION — deferred, TODO comments only (9 remaining)

These findings were annotated with `// TODO: UC_USELESS_CONDITION` comments for resolution in a future PR:

| File | Line | Note |
|------|------|------|
| `r4/fhirpath/FHIRLexer.java` | 254 | `} else if (ch == '` `` ` `` `) {` — condition is always false |
| `r4b/fhirpath/FHIRLexer.java` | 254 | Same |
| `r5/fhirpath/FHIRLexer.java` | 254 | Same |
| `r4b/conformance/ProfileUtilities.java` | 4775 | `!hasDef` is always false inside `if (hasDef && ...)` |
| `r5/renderers/StructureDefinitionRenderer.java` | 1325 | Same pattern |
| `r5/renderers/StructureDefinitionRenderer.java` | 4308 | `mode==GEN_MODE_DIFF` flagged as always true/false in context |
| `r5/renderers/StructureDefinitionRenderer.java` | 4916 | `"1".equals(d.getMax())` flagged as always true/false in context |
| `r5/context/CanonicalResourceManager.java` | 611 | GC_UNRELATED_TYPES no-op remove (see Batch J) |
| `validation/instance/InstanceValidator.java` | 1449 | `pattern` flagged as always true/false in context |

grahamegrieve (approved):
ok

PR #2465, opened 2026-06-01
Fix illegal windows filenames in TestDataFactory / TestDataProvider 

dotasek:
File paths generated by processing test data names containing cell ranges (A1:B5, etc) would crash in Windows environments.

This parses the test data beforehand to only handle the path portion of the test as a filename.


grahamegrieve (approved):
ok

PR #2466, opened 2026-06-01
Fix test inclusion for rX/narrative

dotasek:
rX/narrative test cases weren't being included in maven test runs.

The test will fail until https://github.com/FHIR/fhir-test-cases/pull/268 is merged to address changes in an expected outcome.


grahamegrieve (approved):
ok

PR #2407, opened 2026-04-20
2026 04 gg terminology fixes

grahamegrieve:

PR #2467, opened 2026-06-02
Fix: restore binary compatibility of IValidationPolicyAdvisor

bassspieler:
### Problem

PR #2407 changed the signature of IValidationPolicyAdvisor.isSuppressMessageId from

```
boolean isSuppressMessageId(String path, String messageId);
```
to

```
boolean isSuppressMessageId(String path, String messageId, Object... theMessageArguments);
```

In Java, varargs produce a distinct method descriptor. This removes the original method from the interface contract, breaking all existing implementations — they either fail to compile or throw `AbstractMethodError` at runtime.

This is a confirmed cause of the incompatibility between `hl7.fhir.core 6.9.9` and `hapi-fhir 8.10.0`. Affected classes include `ValidatorPolicyAdvisor`, `FhirDefaultPolicyAdvisor`, and `BaseJpaR4Test.ValidationPolicyAdvisor.`

### Fix

Add a `default` bridge method with the old signature that delegates to the new one. Existing implementations continue to work without modification; implementations that want access to `theMessageArguments` can override the new method.

Related to #2407

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/validation/IValidationPolicyAdvisor.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/validation/IValidationPolicyAdvisor.java
@@ -284,5 +284,7 @@ public interface IValidationPolicyAdvisor {
    * or replaced by some arbitrary placeholder (e.g. "XXX") to allow for automated tested frameworks
    * @return a value to use, or null to use the real value
    */
-  String relativeDatePlaceHolder();
+  default String relativeDatePlaceHolder() {
+    return null;
+  }
 }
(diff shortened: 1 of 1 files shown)


grahamegrieve:
Is this really a problem? You should not be ignoring the parameters

bassspieler:
Sorry, the fix is wrong. But the problem is real. 

The class `FhirDefaultPolicyAdvisor` of HAPI 8.10.0 validation module breaks because it is forced to implement the new method.

Your comment "You should not be ignoring the parameters" is okay for clients calling the method `isSuppressMessageId` but all classes outside of hl7-fhir that are implementing `IValidationPolicyAdvisor` are broken and do not compile.

I'll think again and hopefully come up with a real fix soon.

dotasek:
@bassspieler @grahamegrieve there is already a PR in HAPI that is approved and awaiting merge that addresses this:

https://github.com/hapifhir/hapi-fhir/pull/7962

bassspieler:
Thank you @dotasek

PR #2468, opened 2026-06-03
Fix: preferred binding violations incorrectly reported as ERROR (#2269)

### LINKED ISSUE #2269: Binding reported as ERROR by validation although binding strength is only preferred
We are currently using the validation to verify if the fhir-structures we produce are compliant to the profile.
While upgrading ca.uhn.hapi.fhir:hapi-fhir-validation from 8.2.0 to 8.6.0 we start getting validation failures which were not present before.
The last version without the additional findings seems to be 8.4.2.

It seems to us, that the failure is a false positive as non-compliance with a binding of strength preferred is reported as Error.
To our understanding, it is totally valid to not comply with a binding of strength preferred.
The profile we intend to implement and verify actually requires to not follow the preference.
The resource in question can be found here: https://simplifier.net/packages/de.gematik.vsdm2/1.0.0-rc5/files/2973452
There, the attribute Adress inherits from a base profile, but also changes semantics of the binding.
An example of the desired binding is also provided by the authors of the profile: https://simplifier.net/packages/de.gematik.vsdm2/1.0.0-rc5/files/2973424/~json

As the address.country is not an iso-3166 country code as preferred, but contains the localized country name instead, this triggers the validation issue.
This seems to be some kind of unintentional regression in the validation logic.
If we have missed some kind of new setting for the new version, we would be grateful for advice.

bassspieler:
## Description

Fixes a regression where terminology issues produced during binding validation were not consistently capped by binding strength. This caused violations of a `preferred` binding to be reported as `ERROR`, although the FHIR specification explicitly allows non-conformance with preferred bindings.

Fixes #2269

## Changes

- Ensured that terminology validation issues are consistently limited to non-fatal severity when the binding strength is `preferred`
- Behavior for `required` bindings is unchanged

## Testing

- Added a unit test that reproduces the regression scenario: a `preferred` binding violation that previously produced `ERROR` now correctly produces `WARNING`
- Verified locally against a real-world FHIR R4 profile (`Address.country` with preferred ISO 3166 binding from the German base profile `fhir.de/StructureDefinition/address-de-basis`)
- Tested end-to-end in a production-like validation service — the `OperationOutcome.issue.severity` correctly changed from `error` to `warning`

## References

- Reproducing test case in fhir-test-cases: https://github.com/FHIR/fhir-test-cases/pull/264
- FHIR spec on binding strength: https://www.hl7.org/fhir/terminologies.html#strength

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/validation/IValidationPolicyAdvisor.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/validation/IValidationPolicyAdvisor.java
@@ -284,5 +284,7 @@ public interface IValidationPolicyAdvisor {
    * or replaced by some arbitrary placeholder (e.g. "XXX") to allow for automated tested frameworks
    * @return a value to use, or null to use the real value
    */
-  String relativeDatePlaceHolder();
+  default String relativeDatePlaceHolder() {
+    return null;
+  }
 }
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/validation/IValidationPolicyAdvisor.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/validation/IValidationPolicyAdvisor.java
@@ -72,7 +72,17 @@ public interface IValidationPolicyAdvisor {
    * @param messageId - the message id (from messages.properties)
    * @return true if the validator should ignore the message
    */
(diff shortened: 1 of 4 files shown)


grahamegrieve (commented):
This causes 18 validation tests to fail, and I think they should fail. If you really think that there's a problem you need to do more work here to figure out under what conditions there's a problem

bassspieler:
I fail to figure out how to locally run the integration tests. It's a shame. Coding professionally since 2009 and unable to execute Java stuff -.-

I'll open a new PR once I learned how to work on this project the right way.

PR #2469, opened 2026-06-04
FHIR Mapping Language parse/render fixes + enhancements

brianpos:
Many fixes to the FML parser and renderer including: *(and one fhirpath `toString()` bug)*
* Detecting Auto-generated Rule names to omit from rendering to FML text (backtic escaped fields impacted)
* Support for description markdown metadata element parsed in from triple quoted content `""" description"""`
* String encoding in FHIRPath isn't jsonEscaped - single not double quotes. Impacted FML rendering from StructureMap when the content had a single quote in a string it wasn't escaped e.g. `'Single  \'quotes\' in a string'`
* Implement the Simple Form: Identity Transform parse/render
    e.g. `src -> tgt: type, subtype, action, recorded;`
    This round trips, but only when outputing the same parsed instance - as uses UserData to retain the list, otherwise the long form is rendered.
* Use a common parsing function for all the fhirpath in the FML text `check`, `log`, `where`
     Better error handling with the required `()`
* Correct parsing the share modes - some legacy values were still in there
* `share` target list mode requires a `listRuleId` to follow it
* FHIRLexer processConstant now inline with processConstantString from the FHIRPathEngine's parser
* Escape the content in the FML metadata while rendering them
* parse/render the StructureMap `experimental` property to metadata
* Add support for parsing/rendering FML constants
* The FML default is a fhirpath expression, so it doesn't need escaping
* Add support for backtics in source/target element names
* Add support for `queried` and `produced` structure map model modes
* Title sets the title, not the description, Description sets description
* Only render the description if it is different to the title
* When outputting single quoted content, using the correct escaping routine (not escapeJava which does double quotes)

All this is to the R5 StructureMap/FhirPath engine.

The second batch of refinements to this also cover the element model based FmlParser
* Embeded ConceptMap status has same status as StructureMap
* Uses and Rule documentation blocks made consistent (pre in documentation, post comments in UserData)
* Add a comment to the ValidateRuleTarget that the "evaluate" transform requires a single parameter is incorrect - so can be fixed later
* The type of the output from the translate operation is defined by the 3rd parameter - which can be `code`, `Coding` or `CodeableConcept`.
* FmlParser (element model) - parse const/let statements and correct fhirpath expression handling (where, log, check, default)
* FmlParser (element model) - parse source/target context/element with support for backtick processing
* Parameter constants should be typed, not all strings
* Cleanup metadata parsing to be consistent with the StructureMapUtilities class, and include support for the markdown """ processing for description
* The Create transform requires a single parameter - the type it is going to create (check size of the collection before just referencing it)
   *This was causing an index range error in some maps*

And also apply all these changes to the Narrative StructureMapRenderer class too


grahamegrieve (approved):
ok

PR #2472, opened 2026-06-04
2026 06 gg txcache

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/utilities/TerminologyCache.java
dotasek, line 1: There is a lot of work done for synchronizing the CacheEntry list and map in NamedCache.

It may be better to use a single map with something like the following implementation (Original here: https://stackoverflow.com/a/16989040):

```java
import java.util.LinkedHashMap;
import java.util.Map;

public class CircularFifoMap<K, V> extends LinkedHashMap<K, V> {
    private final int maxSize;

    public CircularFifoMap(int maxSize) {
        // 'true' initializes access-order if you want LRU; 
        // leave it out (or use false) for strict FIFO insertion order
        super(maxSize, 0.75f, false); 
        this.maxSize = maxSize;
    }

    @Override
    protected boolean removeEldestEntry(Map.Entry<K, V> eldest) {
        // Automatically removes the oldest inserted element when size exceeds limit
        return size() > maxSize;
    }
}
```

dotasek (changes requested):
After some additional tests, it appears as though refactoring jsonHash will only result in performance degradation.

I did some minor refactors and renaming in the process of this review, included as a commit.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/utilities/TerminologyCache.java
grahamegrieve: I think that this is a thing for another PR

dotasek (approved):

PR #2473, opened 2026-06-05
Make using the TerminologyCache after unload() a hard error

jmandel:
## What

`TerminologyCache.unload()` flushes the complete cache to disk and then clears it — it is meant to be the **last** terminology operation. Nothing currently enforces that. If a caller uses the cache after `unload()`, the later (partial) entries get written back over the complete `.cache` files, silently corrupting the persisted cache.

This adds the guard @grahamegrieve asked for: `unload()` sets an `unloaded` flag, and the four cache entry points (`getExpansion`, `cacheExpansion`, `getValidation`, `cacheValidation`) throw if used afterward. Plus a unit test.

## Why an `AssertionError` (not a `RuntimeException`)

This is a programming-error invariant on the caller's **code ordering** — no IG input/content can trigger it. Downstream consumers (e.g. the IG Publisher) wrap terminology operations in `catch (Exception)`, so a `RuntimeException` gets swallowed there and the misuse continues silently (degraded output, "successful" build). An `AssertionError` escapes `catch (Exception)` and fails the build, surfacing the bug immediately.

## How it was found / companion fix

The IG Publisher called `context.unload()` mid-build and then ran `genCombinedPackage()`, which expands value sets → used the tx cache after `unload()` → overwrote the persisted cache with partial data → warm builds re-queried everything already computed. With this guard, that ordering now hard-fails with a stack that pinpoints the culprit:

```
java.lang.AssertionError: Terminology cache used after unload()...
  at TerminologyCache.getExpansion(...)
  at BaseWorkerContext.expandVS(...)
  at PackageReGenerator.generateExpansionPackageInner(...)
  at PublisherGenerator.genCombinedPackage(...)
  at Publisher.createIg(...)
```

The publisher-side rewrite that satisfies this guard (moves `unload()` to the end of `createIg()`, after all terminology consumers) is **HL7/fhir-ig-publisher#1318**.

**Sequencing:** the publisher should adopt this guard together with HL7/fhir-ig-publisher#1318 — on its own it would hard-fail every terminology-bearing build until the publisher's ordering is corrected. Opening as a draft for that reason.

## Test

`TerminologyCacheTests.testCacheUseAfterUnloadThrows`: pre-`unload()` use works; all four entry points throw after `unload()`. The existing 42 `TerminologyCache` tests still pass (verified on master).

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/utilities/TerminologyCache.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/utilities/TerminologyCache.java
@@ -382,13 +382,28 @@ public class TerminologyCache {
     return id;
   }
   
+  private boolean unloaded = false;
+
   public void unload() {
     // not useable after this is called — flush any pending writes first so we don't lose
     // entries that were waiting out the SAVE_DELAY_MS coalescing window.
     save();
+    unloaded = true;
     caches.clear();
     vsCache.clear();
     csCache.clear();
+  }
+
+  // Using the terminology cache after unload() is a programming error in the caller: unload()
+  // has already persisted the complete cache, so any later entry would overwrite it on disk with
+  // partial state. This is an invariant on the *code* (no IG content can trigger it — callers must
+  // finish all terminology work before unload()), so fail hard with an AssertionError that is not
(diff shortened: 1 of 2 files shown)


grahamegrieve:
see https://github.com/hapifhir/org.hl7.fhir.core/pull/2472 - already proposed

PR #2475, opened 2026-06-08
R4 TurtleParser cross-version redirect fix

tmprd:
Follow-up to #2230 . This makes the R5 TurtleParser handle R4 syntax, without creating a dependency on the R4 module. 

I also fixed the TTL as HTML rendering -- the one that adds links to the syntax-highlighted version shown in IGs.


grahamegrieve (approved):
ok

PR #2476, opened 2026-06-09
Fix "the the" typo in cardinality tooltip text

devvratbhardwaj:
The "Card." column tooltip had "the the" in the text across R4, R4B, and R3. 
Also fixed a missing space ("itemcan") in the same string in the Questionnaire renderers.


grahamegrieve (approved):
ok

PR #2480, opened 2026-06-12
2026 06 gg fix npe

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/expansion/ValueSetExpander.java
```diff
@@ -246,8 +246,12 @@ private ValueSetExpansionContainsComponent addCode(WorkingContext wc, String sys
     } else {
       if (designations == null) {
         designations = new ArrayList<>();
+      } else {
+        designations = new ArrayList<>(designations); // copy, so we don't mutate the source code system
```
dotasek, line 250: I like that this is not mutating, but it would make things even safer and more explicit if a new `designations` variable was created. 

Say, something like: `designationsWithMaintCode`, so we know what the later code is processing.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/validation/ValueSetValidator.java
```diff
@@ -1219,6 +1219,13 @@ private ValidationResult validateCode(String path, Coding code, CodeSystem cs, C
     
     for (ConceptDefinitionDesignationComponent ds : cc.getDesignation()) {
       opContext.deadCheck("validateCode1 "+ds.toString());
+      if (!isOkLanguage(ds.getLanguage()) && (!ds.hasLanguage() || ds.getLanguage().equals(cs.getLanguage()))
+          && code.getDisplay().equalsIgnoreCase(ds.getValue())) {
+        // the display matches a designation in the code system's default language; that's
+        // acceptable if there are no displays in the requested language(s) (b.count() == 0)
+        // (see tests validation-simple-*-good-language-none)
```
dotasek, line 1226: I see no actual tests with this name. Have these been committed?

dotasek (changes requested):

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/expansion/ValueSetExpander.java
grahamegrieve: that turn out to be tricky. I think it shouldn't be done in this PR

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/validation/ValueSetValidator.java
grahamegrieve: In the terminology tests-  and yes, committed and released

dotasek: Right, I see it after pulling the latest fhir-tx-ecosystem-ig

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/expansion/ValueSetExpander.java
dotasek: OK.

dotasek (approved):

PR #2482, opened 2026-06-13
Fix SpotBugs annotations transitive dependencies

dotasek:
SpotBugs annotations were not included in distributions, but not correctly marked as optional, requiring downstream projects to include them in order to build. 

This fixes the issue and maintains the exclusion of those dependencies.


grahamegrieve (approved):
ok

PR #2483, opened 2026-06-13
2026 06 gg rework caching

grahamegrieve:


dotasek (commented):
## txCache file review (72 files)

*** Performed by Claude sonnet 4.6 ***

All changes are mechanical regeneration driven by two root causes: a **server version bump** (tx-dev 0.9.4 → 0.9.6, dated 2026-05-20 → 2026-06-13) and the **removal of the `cache-id` expansion parameter** from the server's TerminologyCapabilities (which is now handled via the `$cache-control` protocol instead). No test outcomes changed; the cached validation/expansion results are structurally identical.

### `.terminologyCapabilities.*.cache` (5 files — R3×2, R4×2, R5×1)

Updated server snapshots reflecting:
- Version bumped to `0.9.6`, date updated to 2026-06-13
- US Core CodeSystems promoted from `9.0.0-ballot` → `9.0.0`
- `cache-id` expansion parameter entry **removed** from the `expansion.parameter` list (aligns with the new `$cache-control` protocol)

### `servers.ini` (1 file — R5 only)

Added a new local server entry: `local.fhir.org:3001.r5 = http://local.fhir.org:3001/r5`. Matches the `FhirSettingsPOJO.TX_SERVER_LOCAL` port change.

### `system-map.json` (2 files deleted — 3.0.2 and 5.0.0)

Both `system-map.json` files deleted entirely. These cached server→system routing hints; their removal means routing will be re-resolved dynamically. No content was surprising (all entries pointed to `tx-dev.fhir.org`).

### `cs-externals.json` / `vs-externals.json` (6 files — 3.0.2, 4.0.1, 5.0.0)

Filename pointers updated to reflect new UUID-named `cs-*.json` blobs from the re-run. One new entry added to 4.0.1: `https://smt.esante.gouv.fr/fhir/CodeSystem/tre-r374-nature-activite-smsse-regulee` is now resolved to `https://smt.esante.gouv.fr/fhir` (was previously `null` — the server now knows about it). One entry removed from 4.0.1: `http://terminology.hl7.org/CodeSystem/v2-0131` dropped.

### `cs-*.json` blobs (13 renames, 4 new, 3 deleted — 3.0.2, 4.0.1, 5.0.0)

All renames are content-identical (`R100`); only the UUID filename changed because the server issued new cache keys on the re-run. The 4 new blobs are net additions (new CodeSystems the server resolved this run). The 3 deleted blobs correspond to entries removed from `cs-externals.json`.

### `vs-*.json` blobs (1 rename — 4.0.1)

Same pattern: content-identical rename, new UUID filename.

### Named `.cache` files — `all-systems`, `snomed`, `loinc`, `ucum`, `iso3166`, `lang`, `cvx`, `rxnorm`, `11073`, `ndc`, `icd-*`, `v3-NullFlavor`, and various URL-encoded system caches (44 files)

Two kinds of changes only:
1. **Diagnostics UUIDs rotated** — each cache entry's `diagnostics` field contains server-generated `urn:uuid:...` request trace IDs. These changed because the tests were re-run against the refreshed server. Outcomes (ok/error/class/display) are unchanged.
2. **Timing noise** — millisecond timestamps within `diagnostics` strings shifted by ±1ms. No functional change.
   
One named cache deleted: `4.0.1/http___fhir.de_CodeSystem_ask.cache` lost its final entry (a trailing entry for code `06226` was trimmed, likely because that request is now covered by the new caching path).

One named cache deleted: `4.0.1/https___smt.esante.gouv.fr_...` — removed along with its `cs-externals.json` null entry (now resolved to the external server).

---

All 72 files reviewed. Changes are expected cache regeneration; no validation outcomes were altered.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/ITerminologyClient.java
```diff
@@ -78,6 +85,8 @@ public interface ITerminologyClient {
   CanonicalResource read(String type, String id);
   Iterable<HTTPHeader> getClientHeaders();
   ITerminologyClient setClientHeaders(ClientHeaders clientHeaders);
+  /** Add a single persistent client header (e.g. x-cache-id) without replacing the rest. */
```
dotasek: This is a bit disturbing; addClientHeader is now amplifying the amount of work needed to ensure that clientHeaders are correct. 

If addClientHeader is called during the lifecycle of the client because it's needed by cacheControl, it's possible that someone can call setClientHeaders later and remove it.

I think it would be better to have a dedicated cacheControlHeader object that is automatically appended to the headers in setClientHeaders instead of having to carefully coordinate when its added and when setClientHeaders/addClientHeader can be called.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/TerminologyClientContext.java
```diff
@@ -215,15 +229,54 @@ private void initialize() throws IOException {
...
         }
       }
-      if (txcaps != null && TerminologyClientContext.canUseCacheId) {
-        for (TerminologyCapabilitiesExpansionParameterComponent t : txcaps.getExpansion().getParameter()) {
-          if ("cache-id".equals(t.getName())) {
+      // Caching is engaged by the explicit $cache-control protocol: if the server
+      // advertises the operation (and caching is enabled), ask it to start a cache
+      // and use the server-issued id thereafter, carried as an HTTP header. The
+      // server owns the id, so it can authoritatively reject an unknown cache later.
+      this.cacheId = null;
```
dotasek: This might be a good idea to pull out into a `startCache` method to compliment `endCache`. It can still be called in `initialize`, but then the logic could be more clearly compared between the two.

Also, in this state, we have to do all the initialization to start a new cache.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/TerminologyClientManager.java
```diff
@@ -187,7 +180,7 @@ public TerminologyClientContext chooseServer(ValueSet vs, Set<String> systems, b
     if (serverList.isEmpty()) {
       return null;
     }
-    if (systems.contains(UNRESOLVED_VALUESET) || systems.isEmpty()) {
+    if (systems.contains(UNRESOLVED_VALUESET) || systems.isEmpty() || !useEcosystem) {
       return serverList.get(0);
```
dotasek: An explicit return of `getMaster()` would be useful here.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/client/FHIRToolingClient.java
```diff
@@ -360,7 +360,7 @@ public <T extends Resource> Parameters operateType(Class<T> resourceClass, Strin
         result = client.issueGetResourceRequest(url, withVer(getPreferredResourceFormat(), "5.0"), generateHeaders(false), "GET " + resourceClass.getName() + "/$" + name, timeoutLong);
       }
       if (result.isUnsuccessfulRequest()) {
-        throw new EFhirClientException(result.getHttpStatus(), "Server returned error code " + result.getHttpStatus(), (OperationOutcome) result.getPayload());
+        throw new EFhirClientException(result.getHttpStatus(), serverErrorMessage(result.getHttpStatus(), (OperationOutcome) result.getPayload()), (OperationOutcome) result.getPayload());
```
dotasek: NitPick: this complex call could get wrapped and called here and line 389 with the below:

```java
private <T extends Resource> void throwExceptionIfResultIsUnsuccessful(ResourceRequest<T> result) {
    if (result.isUnsuccessfulRequest()) {
      throw new EFhirClientException(result.getHttpStatus(), serverErrorMessage(result.getHttpStatus(), (OperationOutcome) result.getPayload()), (OperationOutcome) result.getPayload());
    }
  }
```

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
```diff
@@ -918,9 +917,8 @@ public ValueSetExpansionOutcome expandVS(ValueSetProcessBase.TerminologyOperatio
     p.addParameter("count", expandCodesLimit);
     p.addParameter("offset", 0);
     txLog("$expand on " + txCache.summary(vs) + " on " + tc.getAddress());
-    if (addDependentResources(opCtxt, tc, p, vs)) {
-      p.addParameter().setName("cache-id").setValue(new IdType(terminologyClientManager.getCacheId()));
-    }
+    // dependent resources are still sent; the cache-id now travels as an HTTP header
+    addDependentResources(opCtxt, tc, p, vs);
```
dotasek, line 921: Return value is no longer used. The method is private, so it would be cleaner to remove the return altogether.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/DiagnosticReportRenderer.java
```diff
@@ -274,17 +274,19 @@ private boolean scanObsForEffective(List<ObservationNode> observations, Resource
...
-        boolean effMatches = eff != null && eff.matches(obs.child("effective[x]"));
-        if (obs != null && obs.has("effective[x]") && (eff == null || !effMatches)) {
-          return true; 
-        } 
-        if (o.contained != null) { 
-          if (scanObsForEffective(o.contained, eff)) { 
-            return true; 
-          } 
-        } 
+        if (obs != null) {
```
dotasek: We can reduce a little nesting here like so:

```java
for (ObservationNode o : observations) {
      if (o.resolution == null || o.resolution.getResource() == null) {
        continue;
      }
      ResourceWrapper obs = o.resolution.getResource();
      //False positive: not using String.matches
      boolean effMatches = eff != null && eff.matches(obs.child("effective[x]"));
      if (obs.has("effective[x]") && (eff == null || !effMatches)) {
        return true;
      }
      if (o.contained != null) {
        if (scanObsForEffective(o.contained, eff)) {
          return true;
        }
      }
    }
    return false;
```

org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/client/FHIRToolingClient.java
dotasek, line 1: See: https://github.com/hapifhir/org.hl7.fhir.core/pull/2483/changes#r3415892235

This could also use the same refactor in many places.

org.hl7.fhir.dstu3.support/src/main/java/org/hl7/fhir/dstu3/support/utils/client/FHIRToolingClient.java
dotasek, line 1: See: https://github.com/hapifhir/org.hl7.fhir.core/pull/2483/changes#r3415892235

This could also use the same refactor in many places.

org.hl7.fhir.utilities/src/main/resources/Messages.properties
```diff
@@ -1562,3 +1562,5 @@ LIQUID_FILTER_NOT_SUPPORTED = The specified liquid filter ''{0}'' is not support
 LIQUID_UNABLE_TO_PARSE_DATETIME = The input to the 'date' Liquid filter must be a valid date - found ''{0}''
 JSON_PROPERTY_INVALID_ARRAY = The JSON array for ''{0}'' is invalid at index {1} - must have at least one valid value
 FHIR_RESOURCE_ID_CANT_BE_EXTENDED = Extensions are not allowed on Resource.id (''{0}'')
+CACHE_ID_UNKNOWN = The cache ''{0}'' is not known to this server. Caches are created with $cache-control?mode=start; this one was never created, or has expired or been released
```
dotasek, line 1565: No references to this in the code?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/ITerminologyClient.java
```diff
@@ -61,6 +61,13 @@ public interface ITerminologyClient {
   Parameters batchValidateVS(Parameters pin) throws FHIRException;
   Parameters subsumes(Parameters pin) throws FHIRException;
   Parameters getValueSetRelationship(ValueSet vsThis, ValueSet vsOther) throws FHIRException, IOException;
+  /**
+   * Invoke the system-level $cache-control operation. mode is "start" (create a
+   * cache and return its server-issued id in the response Parameters), "end"
+   * (release the cache named by the x-cache-id header), or "check". body may carry
+   * resources to front-load on start, or be null.
+   */
+  Parameters cacheControl(String mode, Parameters body) throws FHIRException, IOException;
```
dotasek: This mode would be safer as an enum.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/TxTester.java
```diff
@@ -594,10 +612,24 @@ private ResultInformation runTest(ITxTesterLoader loader, JsonObject suite, Json
       }
       try {
         counter.getAndAdd( 1);
+        // If the server caches, get this thread's cache-id for the suite (starting
+        // the cache + front-loading setup on first use). When cached, the setup is
+        // sent via the cache, not on each request.
+        String cid = cacheForSuite(suite.asString("name"), setup);
```
dotasek: suiteCacheId would be a clearer name.

dotasek (changes requested):

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/ITerminologyClient.java
grahamegrieve: well, the problem is setClientHeaders itself

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/TerminologyClientContext.java
grahamegrieve: I don't understand either of those points.

org.hl7.fhir.utilities/src/main/resources/Messages.properties
grahamegrieve: it's in FHIRsmith

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
grahamegrieve: ok

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/TxTester.java
grahamegrieve: ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/ITerminologyClient.java
grahamegrieve: ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/TerminologyClientContext.java
grahamegrieve: ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/TerminologyClientManager.java
grahamegrieve: ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/ITerminologyClient.java
grahamegrieve: add doco

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/client/FHIRToolingClient.java
grahamegrieve: I don't understand this at all

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/DiagnosticReportRenderer.java
grahamegrieve: ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/client/FHIRToolingClient.java
grahamegrieve: ok

org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/client/FHIRToolingClient.java
grahamegrieve: ok

org.hl7.fhir.dstu3.support/src/main/java/org/hl7/fhir/dstu3/support/utils/client/FHIRToolingClient.java
grahamegrieve: ok

dotasek (approved):
Approved pending @grahamegrieve's confirmation that "compare" and "related" changes are as intended.

dotasek (changes requested):
TerminologyClientR5 is missing the change from "related" to "compare".

There is also a reference to the former operation that probably needs an update: org/hl7/fhir/validation/instance/type/CompliesWithChecker.java:488

dotasek (approved):

PR #2489, opened 2026-06-17
feat: Add `-only-direct-ig-dependencies` flag to suppress transitive package loading

patrick-werner:
## Summary

Adds a new CLI flag `-only-direct-ig-dependencies` to the Java validator that prevents transitive
package dependencies from being loaded when resolving `-ig` package references.

By default the validator walks the full dependency tree of every loaded package. This flag limits
loading to direct dependencies only — for every package loaded (including the hardcoded ones such
as `hl7.terminology` and `hl7.fhir.uv.extensions`).

**Warning:** Enabling this flag may cause validation errors if resources rely on content from
transitively loaded packages.

## Changes

- `ValidationEngineOptions` — new `@CommandLine.Option` field `-only-direct-ig-dependencies`
- `ValidationEngineOptionsConvertor` — maps the CLI flag to `ValidationEngineParameters`
- `ValidationEngineParameters` — new `onlyDirectIgDependencies` field (with `equals`/`hashCode`/`toString`)
- `ValidationContext` — same field added, including copy constructor, `equals`/`hashCode`/`toString`
- `ValidationContextUtilities` — bi-directional mapping between `ValidationContext` and `ValidationEngineParameters`
- `ValidationService#loadIgsAndExtensions` — passes the flag through to `IgLoader` for all packages,
  including `hl7.terminology` and `hl7.fhir.uv.extensions`
- `IgLoader` — new 5-arg `loadIg` overload; when flag is set, direct dependencies are loaded via
  `loadPackageIntoContext` without recursing; extracted `loadPackageIntoContext` helper to remove
  code duplication

## Tests

- `IgLoaderTests` — two new unit tests: verifies default transitive loading behaviour is unchanged,
  and verifies that transitive deps are skipped when the flag is set
- `ValidationEngineOptionsConvertorTest` — new parameterized case for the flag mapping
- `ValidationContextUtilitiesTests` — new round-trip test (`ValidationEngineParameters` →
  `ValidationContext` → `ValidationEngineParameters`)

### DIFF
--- a/org.hl7.fhir.validation.cli/src/main/java/org/hl7/fhir/validation/cli/picocli/options/ValidationEngineOptions.java
+++ b/org.hl7.fhir.validation.cli/src/main/java/org/hl7/fhir/validation/cli/picocli/options/ValidationEngineOptions.java
@@ -47,6 +47,11 @@ public class ValidationEngineOptions {
   @With
   public boolean recursive = false;
 
+  @CommandLine.Option(names = {"-only-direct-ig-dependencies"},
+    description = "Skip transitive package dependencies when loading IGs, including hardcoded ones (e.g. hl7.terminology). Warning: may cause validation errors if resources depend on transitively loaded content.")
+  @With
+  public boolean onlyDirectIgDependencies = false;
+
   @CommandLine.Option(names = {"-clear-tx-cache"},
     description = "Clear the terminology cache before validation")
   @With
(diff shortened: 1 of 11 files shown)


grahamegrieve:
This seems like a very bad idea to me. why on earth would you want to do this?

patrick-werner:
> This seems like a very bad idea to me. why on earth would you want to do this?

To have full control which dependencies are in scope during validation. I am aware that this is an edge case.
The hapi server has already the option to configure if transitive dependencies are loaded, or not.

E.g. hl7.fhir.uv.xver-r5.r4 adds r5 terminology package in the context through its dependencies, this then can lead to valdiation issues on Resource.status elements.

grahamegrieve:
> To have full control which dependencies are in scope during validation. I am aware that this is an edge case.

well, then, why not just say that *no* dependencies are loaded, only what you list?

> E.g. hl7.fhir.uv.xver-r5.r4 adds r5 terminology package in the context through its dependencies, this then can lead to valdiation issues on Resource.status elements.

well, that's an actual problem to solve, not merely to work around then?

PR #2490, opened 2026-06-18
escape html output properly

grahamegrieve:


dotasek (approved):
@grahamegrieve these fixes appear to work. I have added a test for Scanner using a fake malicious IG. If that looks OK, then this can be merged.

Issue #2494, opened 2026-06-23
Saxon Dependency Reintroduced after update to 6.9.10


cz-ehex:
Hey,
after Updating to Version 6.9.10 we experience NoClassDefFoundError for the Saxon Transformer, although we do not use directly Code which is dependent on it. 
[This Commit](https://github.com/hapifhir/org.hl7.fhir.core/commit/01ca2ecdefec9b33204d2495fe78af8c0dc52298) moved SaxonCode from XsltUtilities to XMLUtil, which leads to the error if your Project excludes Saxon and uses the XMLUtil.
Following simple Reproducer throws the Error: `java.lang.NoClassDefFoundError: net/sf/saxon/TransformerFactoryImpl`
```
[...]
    implementation("ca.uhn.hapi.fhir:org.hl7.fhir.utilities:6.9.10") {
        exclude("net.sf.saxon", "Saxon-HE")
    }
[...]
```
```kotlin
   @Test
    void test() {
        XMLUtil.charSetImpliesAscii("US-ASCII");
    }
```

cz-ehex:
This PR https://github.com/hapifhir/org.hl7.fhir.core/pull/2495 should fix it

PR #2495, opened 2026-06-23
fix(#2494): moves Saxon dependent Code only where it is needed

cz-ehex:


dotasek (approved):

dotasek:
@cz-ehex This needs two fixes to org.hl7.fhir.utilities.XsltUtilities to compile:

```java
import javax.xml.XMLConstants;

//Line 103 (104 with new import)
TransformerFactory f = newXXEProtectedSaxonTransformerFactory();
```

dotasek (changes requested):
See comments regarding compilation error. The method move looks good to me otherwise.

dotasek (approved):

PR #2496, opened 2026-06-23
[codex] Bound FHIR web calls across retries

jmandel:
## Summary

This PR adds a whole-call timeout to `ManagedFhirWebAccessor` and tightens retry accounting in `RetryInterceptor` so stalled FHIR web calls are bounded across retries.

## Original bug

An IG Publisher run could hang for hours during terminology validation. The observed thread dump showed the publisher blocked in OkHttp while waiting for response headers from a terminology call:

- `okhttp3.internal.http2.Http2Stream.waitForIo`
- `okhttp3.internal.http2.Http2Stream.takeHeaders`
- `RetryInterceptor.intercept`
- `ManagedFhirWebAccessor.httpCall`
- R4 terminology client validation code

In that state, the existing retry interceptor could not help because the call never returned to the interceptor as either a response or an exception.

## Trigger

The failure mode is triggered when a terminology server connection accepts a request but then stalls before returning response headers. In the reported case this happened while IG Publisher was validating terminology through `tx.fhir.org`; locally, the same shape can be reproduced by pointing publisher `-tx` at a proxy that passes through metadata requests but intentionally holds a `$validate-code` response open.

## Fix

The change configures OkHttp `callTimeout` in addition to the existing connect/write/read timeouts. The call timeout is computed from:

- the per-attempt timeout already configured on the accessor
- the configured retry count
- the retry delay between attempts

That gives each attempt the existing timeout budget while still bounding the entire call, including retries and retry sleeps.

The retry interceptor was also updated to:

- treat `maxRetry` as the number of retries after the initial attempt
- avoid mutable interceptor-level retry state
- close unsuccessful responses before retrying
- return the final unsuccessful response or rethrow the final `IOException`
- preserve interrupt status if retry sleep is interrupted

## Testing

Local validation performed:

- `git diff --check`
- `../../bin/gym mvn -pl org.hl7.fhir.utilities -Dtest=ManagedFhirWebAccessorTests -Danimal.sniffer.skip=true test`
  - `ManagedFhirWebAccessorTests`: 4 tests, 0 failures, 0 errors
- `../../bin/gym build-core`
  - passed
- `../../bin/gym build-publisher --local-core`
  - passed; publisher jar was rebuilt with the patched local core
- Publisher-level smoke test:
  - ran the rebuilt IG Publisher jar against a temporary SMART Web Messaging IG with an added LOINC/UCUM Observation
  - used `-tx http://127.0.0.1:8765/r4`
  - local proxy passed through `GET /r4/metadata` and related lookup GETs to `https://tx.fhir.org`
  - local proxy intentionally stalled the first `POST /r4/CodeSystem/$validate-code` before response headers
  - publisher completed in about 85 seconds under a 300-second hard guard

The publisher smoke test was HTTP/1.1 rather than the exact HTTP/2 production path, but it exercises the same response-header stall class that left the original run blocked inside OkHttp.

### DIFF
--- a/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/ManagedFhirWebAccessor.java
+++ b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/http/ManagedFhirWebAccessor.java
@@ -5,7 +5,6 @@ import org.hl7.fhir.utilities.ToolingClientLogger;
 import org.hl7.fhir.utilities.http.okhttpimpl.LoggingInterceptor;
 import org.hl7.fhir.utilities.http.okhttpimpl.ProxyAuthenticator;
 import org.hl7.fhir.utilities.http.okhttpimpl.RetryInterceptor;
-import org.hl7.fhir.utilities.settings.ServerDetailsPOJO;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -133,7 +132,15 @@ public class ManagedFhirWebAccessor extends ManagedWebAccessorBase<ManagedFhirWe
     builder.proxyAuthenticator(new ProxyAuthenticator());
     return builder.connectTimeout(timeout, timeoutUnit)
       .writeTimeout(timeout, timeoutUnit)
-      .readTimeout(timeout, timeoutUnit).build();
+      .readTimeout(timeout, timeoutUnit)
+      .callTimeout(calculateCallTimeout(timeout, timeoutUnit, retries), TimeUnit.MILLISECONDS).build();
+  }
+
+  static long calculateCallTimeout(long perAttemptTimeout, TimeUnit perAttemptTimeoutUnit, int retries) {
+    long attemptTimeoutMillis = perAttemptTimeoutUnit.toMillis(perAttemptTimeout);
+    int attempts = Math.max(0, retries) + 1;
+    return Math.addExact(Math.multiplyExact(attemptTimeoutMillis, attempts),
+      Math.multiplyExact(RetryInterceptor.RETRY_DELAY_MILLIS, Math.max(0, retries)));
   }
 
 }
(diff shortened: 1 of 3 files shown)


jmandel:
@grahamegrieve for background: this issue causes occasional IG Publisher jobs in the auto-build pipeline to hang for 10h until they're killed by our pipeline. It'd be better if the publisher noticed these hung connections so its internal retry process could kick in.

jmandel:
Follow-up design note for reviewer discussion before going deeper:

This PR is intentionally the minimal bounded-failure fix for the observed IG Publisher hang: add an OkHttp `callTimeout` around FHIR web calls and clean up the existing retry loop. It prevents the response-header wait from lasting forever.

A cleaner follow-up would be to make retry policy explicit at the accessor layer rather than inside an OkHttp application interceptor. The policy shape would be:

1. Give each attempt its own per-attempt timeout, e.g. `60s` for a tx operation.
2. Retry up to a configured count, with fixed/exponential backoff.
3. Optionally apply a broader total deadline as a guardrail.

That would mean each retry creates a fresh OkHttp `Call` with its own `callTimeout(perAttemptTimeout)`. This is more idiomatic than using one OkHttp `RealCall` whose `callTimeout` spans the interceptor and all `chain.proceed()` retry attempts.

The main design question is scope and semantics:

- For terminology validation operations such as `$validate-code`, retrying is generally reasonable because the operation is effectively read-only.
- For arbitrary FHIR `POST`, retrying can be unsafe if the operation has side effects, even though the current interceptor already retries broadly.
- The FHIR REST/terminology clients mostly funnel through `ManagedFhirWebAccessor`, so a follow-up could first improve that path without changing generic package/web downloads or the host-provided `ManagedWebAccess.MANAGED` contract.

So my suggested path is: merge/consider this as the small hang-prevention fix, then discuss a follow-up PR that replaces interceptor-based retry with an explicit per-attempt retry policy in `ManagedFhirWebAccessor`.

dotasek:
@jmandel there has been a refactor of these utilities in https://github.com/hapifhir/org.hl7.fhir.core/pull/2519, so this will need to be reconciled with that. It's still fresh in my mind, so tag me in a comment once it's done, and I can review (Grahame will likely want a few words, too).

Issue #212, opened 2020-06-02
Unable to successfully run mvn install on downloaded or git cloned project


debashish-github:
Hi,
   I am trying to build the project locally both by using git clone / downloading as zip.
When I try to run mvn install  I get the following error :
[INFO] Scanning for projects...
[WARNING] The POM for org.apache.felix:maven-bundle-plugin:jar:3.5.0 is invalid, transitive dependencies (if any) will not be available, enable debug logging for more details
[ERROR] Invalid extension descriptor for org.apache.felix:maven-bundle-plugin:3.5.0: zip END header not found
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] Unknown packaging: bundle @ line 13, column 16
[ERROR] Unknown packaging: bundle @ line 13, column 16
[ERROR] Unknown packaging: bundle @ line 13, column 16
[ERROR] Unknown packaging: bundle @ line 13, column 16
[ERROR] Unknown packaging: bundle @ line 13, column 13
[ERROR] Unknown packaging: bundle @ line 13, column 13
[ERROR] Unknown packaging: bundle @ line 13, column 16
[ERROR] Unknown packaging: bundle @ line 13, column 16
[ERROR] Unknown packaging: bundle @ line 16, column 16
 @
[ERROR] The build could not read 9 projects -> [Help 1]
[ERROR]
[ERROR]   The project ca.uhn.hapi.fhir:org.hl7.fhir.utilities:5.0.6-SNAPSHOT (C:\fhir\org.hl7.fhir.core-master\org.hl7.fhir.utilities\pom.xml) has 1 error
[ERROR]     Unknown packaging: bundle @ line 13, column 16
[ERROR]
[ERROR]   The project ca.uhn.hapi.fhir:org.hl7.fhir.dstu2:5.0.6-SNAPSHOT (C:\fhir\org.hl7.fhir.core-master\org.hl7.fhir.dstu2\pom.xml) has 1 error
[ERROR]     Unknown packaging: bundle @ line 13, column 16
[ERROR]
[ERROR]   The project ca.uhn.hapi.fhir:org.hl7.fhir.dstu2016may:5.0.6-SNAPSHOT (C:\fhir\org.hl7.fhir.core-master\org.hl7.fhir.dstu2016may\pom.xml) has 1 error
[ERROR]     Unknown packaging: bundle @ line 13, column 16
[ERROR]
[...]

PR #2499, opened 2026-06-26
Related to Kindling issue #212 Diff logic isn't handling nested codes (FHIR-54571, FHIR-54489)

brianpos:
Related to Kindling issue #212 Diff logic isn't handling nested codes (FHIR-54571, FHIR-54489)

When comparing xver codes listed for the basic FHIR version diffs html pages, ensure that you walk all the child concepts too.
Test resources to review are condition that before this fix would report that between R4 and R6 recurrence was removed, which it wasnt.


grahamegrieve (approved):
ok

PR #2501, opened 2026-06-30
Fix intermittent ScannerTest failure due to test order

dotasek:
ScannerTest fails when other tests save hl7.fhir.uv.ips#1.1.0 in the cache in R5  format.

This change requests the required IG for the test with more specificity, bypassing the need to reload all other IGs from the cache and bypassing this issue.


grahamegrieve (approved):
ok

PR #2504, opened 2026-07-01
2026 07 gg tx rixes

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/validation/ValueSetValidator.java
```diff
@@ -1185,13 +1191,16 @@ private ValidationResult validateCode(String path, Coding code, CodeSystem cs, C
...
       if ("retired".equals(status)) {
-        String sMsg2 = context.formatMessage(I18nConstants.INACTIVE_CONCEPT_FOUND, status, cc.getCode());
-        info.addIssue(makeIssue(IssueSeverity.WARNING, IssueType.BUSINESSRULE, path, sMsg2, OpIssueCode.CodeComment, null, I18nConstants.INACTIVE_CONCEPT_FOUND));
-        statusMessage = sMsg2+"; "+statusMessage;
+        messageId = I18nConstants.INACTIVE_CONCEPT_FOUND_ADD;
+        statusId1 = status;
+        statusId2 = "inactive";
       }
+      statusMessage = context.formatMessage(messageId, statusId1, cc.getCode(), statusId2);
+      info.addIssue(makeIssue(IssueSeverity.WARNING, IssueType.BUSINESSRULE, path, statusMessage, OpIssueCode.CodeComment, null, I18nConstants.INACTIVE_CONCEPT_FOUND));
```
dotasek, line 1203: This is using INACTIVE_CONCEPT_FOUND as the `msgIg`. Considering the code above, should it not be the `messageId` set via the new logic?

pom.xml
```diff
@@ -25,7 +25,7 @@
         <guava_version>32.0.1-jre</guava_version>
         <gson_version>2.13.1</gson_version>
         <hapi_fhir_version>8.10.0</hapi_fhir_version>
-        <validator_test_case_version>1.7.64</validator_test_case_version>
+        <validator_test_case_version>1.7.65-SNAPSHOT</validator_test_case_version>
```
dotasek, line 28: I see a version bump here, but no actual changes to fhir-test-cases since 1.7.64 was released. Does this require commits to that project?

dotasek (changes requested):
See comments. The tests do not pass for this (maybe because of 'fhir-test-cases'?).

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/validation/ValueSetValidator.java
grahamegrieve: no, it shouldn't be. that's deliberate - the syntactical variation isn't semantically useful. 

Well, that might not be a right decision, but it's deliberate anyway

pom.xml
grahamegrieve: no but my next will, so I committed this anyway

dotasek (approved):
Approved pending resolution of test failures.

PR #2510, opened 2026-07-07
R6 RDF concept IRI fix for examples - FHIR-57496

tmprd:
https://jira.hl7.org/browse/FHIR-57496 - task 4:
> Change the examples generator (in the FHIR build process) to add Concept IRIs to CodeableConcepts using rdf:type (in examples where the appropriate Concept IRI is known to the generator), and no longer add them to Codings.

This PR affects R6 only.

This PR removes `rdf:type` from Codings which wasn't the right interpretation for most codes. It also moves `rdf:type` to example CodeableConcepts in cases where it's more plausible (skipping cases like `ElementDefinition.pattern[CodeableConcept]`) -- like in [DSTU2](https://hl7.org/fhir/DSTU2/rdf.html#1.18.3.4.5). The new guidance stipulates that this should *only* be done when it makes sense for the FHIR element and concept in question. 

This is separate from the Concept IRI Extension https://build.fhir.org/ig/HL7/fhir-extensions/StructureDefinition-rdf-concept-iri.html which is also optional on *Coding*, is more round-trippable in JSON/XML, still RDF-linkable, and helps preserve backwards compatibility. That extension is NOT added to any examples but documented on the RDF page.

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/TurtleConceptIri.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/elementmodel/TurtleConceptIri.java
@@ -1,0 +1,110 @@
+package org.hl7.fhir.r5.elementmodel;
+
+import java.util.List;
+
+import org.hl7.fhir.r5.context.IWorkerContext;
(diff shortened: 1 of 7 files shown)


grahamegrieve (approved):
ok

PR #2512, opened 2026-07-08
fix tx server addresses to https:

grahamegrieve:


dotasek (approved):
~~Looks good to me pending test success.~~

Was good at this point. Changes are below.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/TerminologyClientContext.java
```diff
@@ -256,15 +276,37 @@ public void endCache() {
...
+    // the owning context may still be using it.
+    if (cacheOwned) {
+      try {
+        client.cacheControl(ITerminologyClient.CacheControlMode.END_CACHE, null);
+      } catch (Exception e) {
+        // best-effort release; ignore
+      }
     }
     cacheId = null;
+    cacheOwned = false;
```
dotasek: This portion is a little disturbing. 

`cacheId` and `cacheOwned` are only ever set during instantiation of a TerminologyClientContext. However, they are mutated here by a public `endCache` that can be called multiple times.

I would prefer that TerminologyClientContext never allowed `endCache` to be called more than once (and was named `shutdown` or something), and to have `cacheId` and `cacheOwned` be final after instantiation.

dotasek (changes requested):
See comments.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/client/TerminologyClientContext.java
grahamegrieve: updated

dotasek (approved):
Approved pending passing checks.

PR #2516, opened 2026-07-10
Avoid mutating shared bases during snapshot generation

### LINKED ISSUE #2470: `ProfileUtilities.fixTypeOfResourceId()` mutates shared cached `StructureDefinition` under concurrent validation — thread-safety bug
## Cross-reference

This is the `org.hl7.fhir.core`-side report for a race condition first reported in the HAPI FHIR framework: **hapifhir/hapi-fhir#8034**

That issue contains the full reproduction script, stack traces, and architectural analysis. This report focuses on the root-cause fix that belongs in `ProfileUtilities.java`.

---

## Bug Description

`ProfileUtilities.fixTypeOfResourceId()` mutates `TypeRefComponent.extensions` (a plain unsynchronised `ArrayList`) in-place on a `StructureDefinition` that is shared across concurrent threads. Two threads calling this method simultaneously on the same object race on the list, producing either:

- **`IndexOutOfBoundsException`** — one thread empties the list while the other holds a stale index
- **`"Multiple matching extensions found"`** — interleaved `removeExtension` / `addUrlExtension` calls leave duplicate entries

The exception is non-deterministic, confirmed on `org.hl7.fhir.core` versions **6.6.7 through 6.9.4.1** (the range shipped with HAPI 8.0.0 – 8.10.0).

---

## Root Cause in This Repo

`ProfileUtilities.generateSnapshot()` calls `fixTypeOfResourceId(base)` **unconditionally** at line 769, with no guard and no copy:

```java
// ProfileUtilities.java
if (!base.hasSnapshot()) {
    generateSnapshot(sdb, base, ...);
}
fixTypeOfResourceId(base);  // ← ALWAYS called — mutates shared input in-place
```

The inner overload:

```java
private void fixTypeOfResourceId(List<ElementDefinition> list) {
    for (ElementDefinition ed : list) {
        if (ed.hasBase() && ed.getBase().getPath().equals("Resource.id")) {
            for (TypeRefComponent tr : ed.getType()) {
                tr.setCode("http://hl7.org/fhirpath/System.String");
                tr.removeExtension(ExtensionDefinitions.EXT_FHIR_TYPE);       // ← mutates shared ArrayList
                ExtensionUtilities.addUrlExtension(tr, ExtensionDefinitions.EXT_FHIR_TYPE, "id");  // ← also mutates
            }
        }
    }
}
```

When two threads call this on the **same** `TypeRefComponent` instance (which happens because HAPI caches the R5 canonical in `IBaseResource.userData` and returns the same object to all callers), they race on the `extensions` `ArrayList`.

---

### Option 1 — Defensive copy before mutation *(recommended — no API change)*

```java
// ProfileUtilities.java ~line 769
StructureDefinition baseCopy = base.copy();  // work on a private copy
fixTypeOfResourceId(baseCopy);
// continue using baseCopy for the remainder of generateSnapshot
```

Eliminates the race at source. `copy()` is already used extensively in this class. No API change required.

### Option 2 — Idempotency check + copy

Skip if already normalized (i.e. extension already absent), but still copy first since the check itself is not atomic with respect to concurrent reads.

---

## Verified Versions

`fixTypeOfResourceId` is identical and unpatched in:
- `6.6.7` (HAPI 8.6.0)
- `6.9.4.1` (HAPI 8.10.0 — current latest)
- `6.9.9` (current latest core)

---

## Reproduction

Full reproduction script (single Java file, runs against `hapi-fhir-cli.jar`): see **hapifhir/hapi-fhir#8034**.

mahitha-ada:
## Summary

- generate snapshots from a private copy before normalizing `Resource.id`
- preserve nested model `userData` while creating that copy
- add a regression test that verifies the supplied base remains unchanged

## Root cause

`ProfileUtilities.generateSnapshot()` called `fixTypeOfResourceId(base)` directly on the caller-provided `StructureDefinition`. Cached base definitions can be shared by concurrent validation requests, so changing their type extensions introduced a shared-state race.

The normalization now returns a private working copy for R4+ resource definitions. This keeps the generated snapshot behavior unchanged without modifying the shared base.

Fixes #2470

## Testing

- `mvn -f org.hl7.fhir.r5/pom.xml -Dtest=ProfileUtilitiesTests test`
- `mvn -f org.hl7.fhir.r5/pom.xml -Dtest=SnapShotGenerationTests test` (164 cases)
- `mvn -f org.hl7.fhir.r5/pom.xml -PCHECKSTYLE -DskipTests process-sources`
- `git diff --check`

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java
@@ -766,7 +766,7 @@ public class ProfileUtilities {
       checkNotGenerating(sdb, "an extension base");
       generateSnapshot(sdb, base, base.getUrl(), (sdb.hasWebPath()) ? Utilities.extractBaseUrl(sdb.getWebPath()) : webUrl, base.getName());
     }
-    fixTypeOfResourceId(base);
+    base = fixTypeOfResourceId(base);
     if (base.hasExtension(ExtensionDefinitions.EXT_TYPE_PARAMETER)) {
       checkTypeParameters(base, derived);
     }
@@ -1287,10 +1287,22 @@ public class ProfileUtilities {
     return (i < list.size() - 1) && list.get(i + 1).getPath().startsWith(ed.getPath()+".");
   }
 
-  private void fixTypeOfResourceId(StructureDefinition base) {
+  private StructureDefinition fixTypeOfResourceId(StructureDefinition base) {
     if (base.getKind() == StructureDefinitionKind.RESOURCE && (base.getFhirVersion() == null || VersionUtilities.isR4Plus(base.getFhirVersion().toCode()))) {
+      base = copyBaseWithUserData(base);
       fixTypeOfResourceId(base.getSnapshot().getElement());
       fixTypeOfResourceId(base.getDifferential().getElement());      
+    }
+    return base;
+  }
+
+  private StructureDefinition copyBaseWithUserData(StructureDefinition base) {
+    boolean oldCopyUserData = Base.isCopyUserData();
+    Base.setCopyUserData(true);
+    try {
+      return base.copy();
+    } finally {
+      Base.setCopyUserData(oldCopyUserData);
     }
   }
 
--- a/org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/ProfileUtilitiesTests.java
+++ b/org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/test/ProfileUtilitiesTests.java
@@ -1,0 +1,85 @@
+package org.hl7.fhir.r5.test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.io.IOException;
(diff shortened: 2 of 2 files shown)


grahamegrieve:
This doesn't sit well with me. This means that the fix is never landed, just reapplied with the cost of a full copy every time. 

Why were you having race conditions? Because marking up the user data isn't something that just happens here

mahitha-ada:
Thanks, that’s a fair concern. The reported race originates in [`hapifhir/hapi-fhir#8034`](https://github.com/hapifhir/hapi-fhir/issues/8034). Since HAPI 8.0, [`hapifhir/hapi-fhir#6508`](https://github.com/hapifhir/hapi-fhir/pull/6508) caches the converted R5 canonical in `IBaseResource.userData` and returns the same `StructureDefinition` to concurrent validation callers. Those callers enter `generateSnapshot()`, which invokes `fixTypeOfResourceId(base)` and concurrently mutates the same `TypeRefComponent.extensions` list.

I agree that the current patch turns what should be a one-time normalization into a full per-call copy and never lands the normalized state on the cached canonical. I’ll revisit the ownership boundary rather than defend that approach. The better direction may be to normalize the canonical once before it is published to HAPI’s `userData` cache, or otherwise prevent the shared mutable canonical from being returned.

Would you prefer this to be fixed in `hapi-fhir` at the cache boundary rather than in core?

grahamegrieve:
I think that in general, these kind of issues should be seen as load time problems - everything should be prepared before made available to the operationall system 

but I think that the core library is not well prepared to do this. I will discuss with @dotasek

PR #2517, opened 2026-07-10
Npm package generator fixes

GinoCanessa:
## Summary

`NPMPackageGenerator` builds the `package.json` manifest for generated FHIR packages.
This PR fixes three related issues in that manifest generation:

1. **R5/R6 IGs got no core dependency.** The private `packageForVersion(...)` helper
   used a hard-coded `startsWith` ladder that only mapped R2–R4B. Any `5.0.x` (R5) or
   `6.0.x` (R6) FHIR version fell through to `null`, so **no `hl7.fhir.rX.core`
   dependency was emitted** for R5/R6 implementation guides.
2. **A core dependency could be added twice.** When an IG's `dependsOn` already declared
   the core package, the auto-add could write the same key onto the dependencies object
   again.
3. **`dependsOn` entries without a version were only a soft warning.** A dependency
   missing its required `version` produced an incomplete manifest instead of failing.

### `org.hl7.fhir.r5` - `NPMPackageGenerator`
- `packageForVersion(...)` now delegates to the shared
  `VersionUtilities.packageForVersion(...)`, which maps **R2 through R6** and normalizes
  raw base-path codes (e.g. `5.0`, `4.5.0`, `5.0.0-ballot` → `hl7.fhir.r5.core`).
  Non-semver codes (e.g. `current`, `0.01`) throw `FHIRException`, which is caught and
  mapped to `null` (no core dep) - preserving the old helper's lenient behavior.
- Core-dependency auto-add is now guarded:
  `if (vp != null && !dep.has(vp) && !dependsOnDeclaresPackage(ig, vp))`. This prevents
  a duplicate key and lets an author-declared `dependsOn` version win.
- New `dependsOnDeclaresPackage(...)` returns true only for a **non-aliased** `dependsOn`
  that has a version and a matching `packageId`. Aliased deps serialize as
  `<id>@npm:<packageId>` and cannot collide, so they don't suppress the auto-add.
- New `validateDependencies(ig)` runs in all three constructors and throws
  `FHIRException` naming the offending `packageId`/`uri` when a `dependsOn` entry has no
  `version`. This replaces the previous soft warning that merely appended
  `dependsOn.version(<uri>)` to the license-notice buffer.

### `org.hl7.fhir.r4b` - `NPMPackageGenerator`
- Applies the same `packageForVersion(...)` → `VersionUtilities.packageForVersion(...)`
  refactor so the R4B generator copy maps R5/R6 core packages consistently. (Dependency
  validation and dedupe are scoped to the R5 generator.)

### Tests - new `org.hl7.fhir.r5.utils.NPMPackageGeneratorTest` (JUnit 5)
- Emits R5 / R6 core dependency for `5.0.0` / `6.0.0`.
- Maps raw base-path version codes (`5.0`, `5.0.0-ballot`, `4.5.0` → `hl7.fhir.r5.core`).
- Non-semver codes (`current`, `0.01`, `0.06`) add no core dep and don't throw.
- Preserves existing R2–R4B mapping (`1.0.2`, `1.4.0`, `3.0.2`, `4.0.1`, `4.3.0`).
- Author-declared `dependsOn` core version wins; no double-add.
- Rejects an explicit `dependsOn` without a version (`FHIRException` naming pkg + uri).
- `CORE` package kind emits no `dependencies` block.
- An aliased core `dependsOn` does not suppress the auto-added core dep.

## Behavior changes for reviewers
- **R5/R6 IGs now gain** a `hl7.fhir.r5.core` / `hl7.fhir.r6.core` dependency they
  previously lacked.
- **IGs with a `dependsOn` missing `version` now fail fast** with a `FHIRException`
  instead of silently producing an incomplete manifest. This is intentional but is a
  stricter behavior worth noting.

## Testing
```
mvn test -pl org.hl7.fhir.r5 -am -Dtest=NPMPackageGeneratorTest
```


grahamegrieve (approved):
ok

grahamegrieve (approved):
ok

GinoCanessa:
Note: I am working on the rebase and adding an additional fix related to the ConceptMap generator relating to `<csId>-<code>` vs. `#<code>`.  Will ping once everything is ready for review again.

GinoCanessa:
Rebase/sync update, plus a scope change I'd like re-confirmed before you merge.

**Synced.** This branch is now on `eae36d901d` and merges cleanly into current
`master` (`4f52ba67c0`).

**Scope has widened into `org.hl7.fhir.services`.** The r5 renderer layer
was forked into that module before this branch's ConceptMap work, so the fork carries
the same anchor/href bugs verbatim and with `org.hl7.fhir.standalone` now
depending on it, they're no longer latent. I've ported the fix mechanically into
the two affected files, in its own commit, so you can drop it with a single
revert if you'd rather handle the fork separately. No tests added there, since
the module has no `src/test`.

**This is the part I'd like re-confirmed given the wider scope** the prior
approval predates both the services module and the `NO_SAVE` feature.

**One thing you need to merge first.** Two `NarrativeGenerationTests` cases
(`ex-cm`, `cm-complex`) currently fail here, and they're not a defect: they
expect the old, broken anchor form. The corrected fixtures are already pushed to
`FHIR/fhir-test-cases` on `map-generation-test-fixes` (`06ce73ff`) but aren't
merged, so the published `1.7.68-SNAPSHOT` still has the old ones. Point
`FHIR-TEST-CASES` at that branch and the suite is 52/52 green. **Merging that
and republishing will turn this PR's r5 job green.**

Two smaller notes: this branch also happens to fix
`r4b NarrativeGeneratorTests.testDateTimeRendering4`, which fails on `master`
under JDK 20+ because of the CLDR 42 U+202F change. And I tried to demonstrate
the fix end-to-end by building `fhir-ig-publisher` against this branch, but the
publisher no longer compiles against current core (44 errors from the package
reorg, starting with `org.hl7.fhir.r5.utils.UserDataNames`).

GinoCanessa:
Added the missing R6/services package-generation fixes: R5/R6 core dependency inference, author overrides, aliases, no-save suppression, and actionable missing-version warnings, while preserving JSON-null compatibility and existing ConceptMap behavior.

The longstanding `IdType` lazy-cache behavior is intentionally unchanged and remains a separate follow-up.

PR #2518, opened 2026-07-12
Code Cleanup - Extract URLUtil, remove dead code, and tighten test visibility

dotasek:
Extracts the duplicated nocache-URL logic into URLUtil, drops unused imports/dead code in SHLParser and SHCParserTests, removes redundant modifiers, and tightens test method/class visibility to package-private per project convention. 

No behavior changes.

PR #2519, opened 2026-07-12
Switch SimpleHTTPClient to okhttp3

dotasek:
Requires #2518 

This switches SimpleHTTPClient to okhttp3 and routes all GET/PUT/POST requests through the same execute method (gives PUT/POST the same redirect behaviour as GET).


grahamegrieve (approved):
ok

PR #2520, opened 2026-07-13
Enhance CDA Narrative export

JasonRoberts-smile:
The current `CDANarrativeFormat` has a key limitation in the XHTML -> CDA Narrative direction: it only consumes those XHTML elements that the class produces in the CDA Narrative -> XHTML direction. For our project, we require a best-effort transformation of a broader set of inputs. We have enhanced the transformation as follows:

- add support for `h1`, `h3`, `h4`, `h5`, and `h6` elements, equivalent to the existing support for `h2` elements
- add support for `div` elements, equivalent to the existing support for `span` elements
- if the input happens to contain CDA Narrative elements (e.g., `content`, `footnote`), these elements will pass through to the output unchanged
- any unsupported elements will be silently dropped, instead of failing the entire transformation with an error
- make the whole thing case insensitive

We also fixed a bug where the `href` attribute of an `a` element was not getting converted.

### DIFF
--- a/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/CDANarrativeFormat.java
+++ b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/CDANarrativeFormat.java
@@ -355,65 +355,41 @@ public class CDANarrativeFormat {
   
   private void processChildNode(IXMLWriter xml, XhtmlNode n) throws IOException, FHIRException {
     switch (n.getNodeType()) {
-    case DocType: 
-    case Document: 
-    case Instruction: 
+      case DocType, Document, Instruction, CData:
       return;
-    case Comment: 
+    case Comment:
       xml.comment(n.getContent(), true);
       return;
     case Text: 
       xml.text(n.getContent());
       return;
     case Element:
-      if (n.getName().equals("br"))
-        processBreak(xml, n);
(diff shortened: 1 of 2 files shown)


grahamegrieve (commented):
what's the deal with 'paragraph'? why is an illegal element supported? why isn't it processed as a paragraph? 

more generally, why support illegal elements?

JasonRoberts-smile:
> what's the deal with 'paragraph'? why is an illegal element supported? why isn't it processed as a paragraph?
> 
> more generally, why support illegal elements?

We frequently encounter invalid XHTML in the field. We need a lenient transformation so that "garbage in" still produces a valid output that retains as much of the original information as possible.

If this isn't compatible with your project philosophy, I can withdraw this MR and submit another that leaves the logic alone and simply opens up some of the private methods to protected so that I can override the behaviour as needed in a local extension.

grahamegrieve:
I think it should be a setting, at least - so that support for non-valid xhtml is an option.

JasonRoberts-smile:
That's a reasonable compromise. Thanks for the feedback. I will update the MR with an optional `isLenient` flag so that the status quo behaviour for unsupported tags is retained, but the lenient processing can be available to those who want it.

JasonRoberts-smile:
Sorry for going silent on this for a while. I was caught up in helping prepare for the new Smile release.

I think I've captured all your suggestions. Lenient processing is now optional, with strict mode being the default for backward compatibility. Please let me know what you think.

grahamegrieve (approved):
ok

PR #2521, opened 2026-07-13
2026 07 gg various fixes

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/validation/ValueSetValidator.java
```diff
@@ -1817,6 +1817,12 @@ private Object notNull(String s) {
   }
 
   private boolean versionIsMoreDetailed(VersionAlgorithm va, String criteria, String candidate) {
+    if (candidate == null) {
+      return false;
+    }
+    if (criteria == null) {
```
dotasek, line 1823: Is this correct? If there's no criteria at all, we just skip all the logic below? It feels counterintuitive.

org.hl7.fhir.validation/src/test/resources/txCache/org.hl7.fhir.validation/4.0.1/.terminologyCapabilities.tx-dev.fhir.org.r4.cache
```diff
@@ -1,12 +1,12 @@
 {
   "resourceType" : "TerminologyCapabilities",
   "id" : "FhirServer",
-  "url" : "https://tx-dev.fhir.org/r4/TerminologyCapabilities/tx",
+  "url" : "http://tx-dev.fhir.org/r4/TerminologyCapabilities/tx",
```
dotasek, line 4: This is weird; is there a local fhir-settings.json on your machine with this hard-coded? I ask because I'm now very https sensitive, and I can't find any other place this might have come from.

dotasek (changes requested):
Requesting feedback on the two comments I left. I will have to review the backward compatibility change in another review.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/validation/ValueSetValidator.java
grahamegrieve: it was wrong. But yes, no logic below if the criteria is null

org.hl7.fhir.validation/src/test/resources/txCache/org.hl7.fhir.validation/4.0.1/.terminologyCapabilities.tx-dev.fhir.org.r4.cache
grahamegrieve: this a canonical URL. we always use http not https for these

dotasek: But it was pulled from https before this PR. That's a little odd.

grahamegrieve: nah, I fixed them to https in search and replace - and shouldn't have

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/validation/ValueSetValidator.java
dotasek: I don't follow. Which part is wrong? And is the current logic in this PR correct, since I see no changes since your comment reply?

dotasek (approved):
Looks good at this point.

PR #2522, opened 2026-07-19
Delegate FHIRPath custom-function hooks to ITransformerServices in StructureMapUtilities

jgsuess:
## What's broken today

`StructureMapUtilities` wires an inner `FHIRPathHostServices` class (implementing
`IHostApplicationServices`) into the `FHIRPathEngine` it uses to evaluate FHIRPath
expressions inside `StructureMap`/FML transforms. That class hardcodes three of the
FHIRPath engine's extensibility hooks:

```java
@Override
public FunctionDetails resolveFunction(FHIRPathEngine engine, String functionName) {
  return null; // throw new Error("Not Implemented Yet");
}

@Override
public TypeDetails checkFunction(...) throws PathEngineException {
  throw new Error("Not Implemented Yet");
}

@Override
public List<Base> executeFunction(...) {
  throw new Error("Not Implemented Yet");
}
```

This means custom FHIRPath functions can never be used inside a `StructureMap`/FML
transform — every custom function invocation dies at parse (`resolveFunction`), analysis
(`checkFunction`), or runtime (`executeFunction`) — *even though* the caller's own
`ITransformerServices` is right there and is already used for the equivalent
`resolveReference` hook (`FHIRPathHostServices.resolveReference` delegates to
`services.resolveReference(...)`). There's no reason for `resolveReference` to be
delegatable but `resolveFunction`/`checkFunction`/`executeFunction` to be dead ends.

## The fix

Add three `default` methods to `ITransformerServices` — `resolveFunction`,
`checkFunction`, `executeFunction` — whose default bodies reproduce the exact previous
hardcoded behavior (return `null`, throw `PathEngineException("Unknown function '...'")`,
throw `Error("Not Implemented Yet")`, respectively). Then have `FHIRPathHostServices`
delegate each call to `services.<method>(...)`, mirroring the existing `resolveReference`
delegation pattern exactly:

```java
@Override
public FunctionDetails resolveFunction(FHIRPathEngine engine, String functionName) {
  return services == null ? null : services.resolveFunction(engine, functionName);
}
```

Applied identically across `org.hl7.fhir.r4`, `org.hl7.fhir.r4b`, and `org.hl7.fhir.r5`.

## Why it's backward-compatible

The three new methods are `default` methods on an existing public interface, and their
default bodies are byte-for-byte equivalent to what `FHIRPathHostServices` used to do
unconditionally. Any existing `ITransformerServices` implementor that does not override
them observes exactly the same behavior as before — same return values, same exceptions,
same messages. This is verified by a regression test
(`StructureMapUtilitiesFunctionDelegationTest.testUnknownCustomFunctionStillFailsTheSameWayWithoutOverride`)
using a plain `ITransformerServices` that only implements the pre-existing methods.

A second test in the same file
(`testCustomFunctionIsDelegatedToTransformerServices`) proves the positive case: a
caller-supplied `ITransformerServices` implementing a custom `negate` function is
invoked end-to-end through `parse`/`analyse`/`transform`, turning `source.active=true`
into `target.active=false` via `tgt.active = (v.negate())` in the FML map — with no
changes to FHIR core needed beyond the delegation itself.

`org.hl7.fhir.r4` includes the full regression test above. `org.hl7.fhir.r4b` and
`org.hl7.fhir.r5` received the identical fix (same delegation pattern, same default
bodies) and are verified by successful compilation of each module; no test was
duplicated there since the r4/r4b/r5 `StructureMapUtilities`/`FHIRPathHostServices`
implementations are otherwise structurally the same.

## Motivating use case

I'm implementing an HL7 v2-to-FHIR bridge using `StructureMap`/FML as the transform
engine. FHIRPath's own spec defines an HL7 v2 message object model appendix
(https://build.fhir.org/ig/HL7/FHIRPath/en/appendices.html#use-of-fhirpath-on-hl7-version-2-messages)
with `group()`, `elements()`, `simple()`, and `components()` functions for navigating v2
message structure directly from FHIRPath expressions. Implementing that appendix means
supplying custom FHIRPath functions from application code — which today is impossible
from within a `StructureMap`/FML transform because of the hardcoded stubs above. This
change lets an application supply those functions (or any custom function) via the
`ITransformerServices` it already passes to `StructureMapUtilities`, with no forking of
this library required.

### DIFF
--- a/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/StructureMapUtilities.java
+++ b/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/StructureMapUtilities.java
@@ -156,6 +156,22 @@ public class StructureMapUtilities {
     public Base resolveReference(Object appContext, String url) throws FHIRException;
 
     public List<Base> performSearch(Object appContext, String url) throws FHIRException;
+
+    // FHIRPath custom-function hooks (see IHostApplicationServices). Default implementations
+    // preserve the previous hardcoded FHIRPathHostServices behavior for callers that don't
+    // override them, so adding these methods is backward-compatible.
+    default FunctionDetails resolveFunction(FHIRPathEngine engine, String functionName) {
+      return null;
+    }
+
+    default TypeDetails checkFunction(FHIRPathEngine engine, Object appContext, String functionName, TypeDetails focus, List<TypeDetails> parameters)
+        throws PathEngineException {
+      throw new PathEngineException("Unknown function '" + functionName + "'");
+    }
+
+    default List<Base> executeFunction(FHIRPathEngine engine, Object appContext, List<Base> focus, String functionName, List<List<Base>> parameters) {
+      throw new Error("Not Implemented Yet");
+    }
   }
 
   private class FHIRPathHostServices implements IHostApplicationServices {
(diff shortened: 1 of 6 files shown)


grahamegrieve (approved):
ok

jgsuess:
Rebased on current master to pick up the `fixme-check.yml` workflow and associated scripts (`.github/workflows/fixme-check/`). The previous CI failure was because these files didn't exist on the branch (it was forked before they were added to master).

The workflows are now showing as `action_required` waiting for maintainer approval to run (standard for first-time fork PRs). Once approved, they should pass — no new FIXME comments are introduced by this PR.

Please approve the workflow runs and re-evaluate for merge. Thank you!

jgsuess:
@grahamegrieve the last round was blocked, because the CI had shifted in the background. Can you please re-check.

PR #2523, opened 2026-07-20
FIXME check for PRs

dotasek:
Adds a trivial check for FIXME text in the diff for a PR. Applies only to *.java files.


grahamegrieve (approved):
ok

PR #2537, opened 2026-08-05
Batch CodeSystem supplement concept validation

vadi2:
`CodeSystemValidator` validates supplement concepts one `$validate-code` call per concept, which costs one round trip each - on a real IG with 124 supplements and 3514 concepts that is 2403 requests and 22.9 minutes against tx.fhir.org on a cold cache. This batches them through `CodeSystem/$batch-validate-code`, wiring up the previously-unused `ITerminologyClient.batchValidateCS` and mirroring the way `validateOnServer2` already selects `CodeSystem/$validate-code` when no value set is in play, taking the same workload to 26 requests and 21.3 seconds with byte-identical validation output.

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
@@ -1211,6 +1211,26 @@ public abstract class BaseWorkerContext extends I18nBase implements IWorkerConte
 
   @Override
   public void validateCodeBatch(ValidationOptions options, List<? extends CodingValidationRequest> codes, ValueSet vs, boolean passVS) {
+    validateCodeBatch(options, codes, vs, passVS, false);
+  }
+
+  /**
+   * Validate a batch of codes against the code systems they name, without reference to any value set.
+   * <p>
+   * This is the batch equivalent of {@link #validateCode(ValidationOptions, String, String, String, String)}:
+   * it asks "is this code valid in this code system", not "is this code in this value set". It therefore
+   * uses CodeSystem/$batch-validate-code, mirroring the way {@link #validateOnServer2} chooses
+   * CodeSystem/$validate-code when no value set is in play.
+   *
+   * @param options controls the validation process
+   * @param codes   the codes to validate; each request carries its own result, so callers can report per-code diagnostics
+   */
+  @Override
+  public void validateCodeBatchCS(ValidationOptions options, List<? extends CodingValidationRequest> codes) {
+    validateCodeBatch(options, codes, null, false, true);
+  }
+
+  private void validateCodeBatch(ValidationOptions options, List<? extends CodingValidationRequest> codes, ValueSet vs, boolean passVS, boolean csBatch) {
     if (options == null) {
       options = ValidationOptions.defaults();
     }
@@ -1282,10 +1302,19 @@ public abstract class BaseWorkerContext extends I18nBase implements IWorkerConte
 
     if (items.size() > 0) {
       TerminologyClientContext tc = terminologyClientManager.chooseServer(vs, systems, false, findValidationLanguage(options));
-      Parameters resp = processBatch(tc, batch, systems, items.size());
+      Parameters resp = processBatch(tc, batch, systems, items.size(), csBatch);
       List<ParametersParameterComponent> validations = resp.getParameters("validation");
       for (int i = 0; i < items.size(); i++) {
         CodingValidationRequest t = items.get(i);
+        if (i >= validations.size()) {
+          // the server returned fewer results than we asked about. We can't tell which ones are missing,
+          // so rather than silently mis-attributing results to the wrong codes, every code we have no
+          // answer for gets its own error
+          t.setResult(new ValidationResult(IssueSeverity.ERROR,
+              formatMessage(I18nConstants.TX_SERVER_BATCH_RESPONSE_INCOMPLETE, validations.size(), items.size()), TerminologyServiceErrorClass.SERVER_ERROR, null)
+              .setTxLink(txLog == null ? null : txLog.getLastId()));
+          continue;
+        }
         ParametersParameterComponent r = validations.get(i);
 
         if (r.getResource() instanceof Parameters) {
@@ -1300,7 +1329,7 @@ public abstract class BaseWorkerContext extends I18nBase implements IWorkerConte
     }
   }
 
-  private Parameters processBatch(TerminologyClientContext tc, Parameters batch, Set<String> systems, int size) {
+  private Parameters processBatch(TerminologyClientContext tc, Parameters batch, Set<String> systems, int size, boolean csBatch) {
     txLog("$batch validate for " + size + " codes on systems " + systems.toString());
     if (terminologyClientManager == null) {
       throw new FHIRException(formatMessage(I18nConstants.ATTEMPT_TO_USE_TERMINOLOGY_SERVER_WHEN_NO_TERMINOLOGY_SERVER_IS_AVAILABLE));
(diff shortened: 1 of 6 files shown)


grahamegrieve:
I asked Claude to review it from 4 perspectives.

it has a different pattern to the value set batch equivalent. is that justified?
Is there any test cases ?
is it only using batches on servers that support batches?
Does it do client side caching proper.ly?

Here's it comments:

# PR 2537 review comments

## Summary comment (post as the review body — "Request changes")

Thanks for this — batching supplement concepts is the right idea and it's a real win; a 5000-concept supplement going from 5000 round trips to 17 is worth having. The caching is sound (reusing `validateCodeBatch` unchanged means the cache tokens are identical to the old per-code path, so batched results still get reused by later single `validateCode` calls), and the tests are more thorough than most.

Three things I want changed before this goes in, and one question.

**1. The truncated-response handling doesn't do what its comment says, and it changes behaviour for three other callers.**

The comment in `validateCodeBatch` says "we can't tell which ones are missing, so rather than silently mis-attributing results to the wrong codes...". But the loop still assigns `validations.get(i)` to `items.get(i)` for every `i < validations.size()`. If the server dropped result #2 rather than truncating the tail, results 3..n are all mis-attributed and only the tail gets flagged — which is exactly the failure mode the comment says it's avoiding. The test bakes the assumption in (it asserts the first 4 answers are trusted).

Worse, the guard sits in the shared `validateCodeBatch`, so it also changes `ValueSetValidator`, `ConceptMapValidator` and `ValueSetRenderer`. Today a malformed or empty batch response throws `IndexOutOfBoundsException`, which their `try/catch` turns into a single EXCEPTION message. After this PR it becomes N per-code ERROR results — and in the CodeSystem case those surface as `CODESYSTEM_CS_SUPP_INVALID_CODE`, i.e. an IG author sees "the code 'X' is not declared in the base CodeSystem" several hundred times when what actually happened is that tx.fhir.org hiccuped. The new `TX_SERVER_BATCH_RESPONSE_INCOMPLETE` text never reaches them, because the supplement rule doesn't print `res.getMessage()`.

Please make a size mismatch fail the whole slice — throw, or leave the results null — so `validateSupplementConcepts` falls back to individual validation. That's the behaviour you've already built for the exception case, and it's the one that gives the right answer.

**2. Don't swallow the batch exception silently.**

`catch (Exception e) { validateSupplementConceptsIndividually(slice, options); }` discards the exception with no log and no validation message. On a genuinely broken server this turns one fast failure into 300 sequential failing round trips per slice — slower than the code this PR replaces. At minimum log it; consider giving up on batching for the rest of the resource once a batch has failed.

**3. The test has unexplained nondeterminism.**

The class comment says "a small, varying number of leading concepts get resolved by the worker context without ever reaching the server", and the assertions were loosened to `> 10` / `> 5` / `> 600` to accommodate it. But `assertEquals(1, server.batchValidateCSCalls)` and `assertEquals(expectedBatches, server.batchValidateCSCalls)` were *not* loosened, so if that varying number ever swallows a whole slice those tests flake. I'd like to know what the variation actually is before this lands rather than have it papered over — a unit test against a fake client shouldn't be nondeterministic at all.

**Question: is `TX_BATCH_VERSION` the right gate?**

`TX_BATCH_VERSION` (1.7.8) is the tx test-suite version for `ValueSet/$batch-validate-code`. This PR uses it to gate `CodeSystem/$batch-validate-code`, a different operation. If a server implements one and not the other we'll send a batch that 404s on every slice. The fallback covers it, but silently and at the cost of a wasted round trip per 300 codes. If the test version does cover both operations, say so in a comment next to the check.

**Design question, not a blocker:** `validateCodeBatch` with `vs == null` already means "no value set", and `validateOnServer2` handles that case by routing to `CodeSystem/$validate-code`. The batch path doesn't — it calls `batchValidateVS` with no `url` parameter — and all three existing callers pass `vs == null`. So rather than a new interface method, `csBatch = vs == null` inside `validateCodeBatch` would fix all four callers, need no change to `IWorkerContext`, and make the batch path agree with the single-code path. Was there a reason to add the parallel entry point instead?

---

## Inline comments

### `BaseWorkerContext.java` — the `i >= validations.size()` block

The comment and the code disagree. This still attributes `validations.get(i)` to `items.get(i)` for the whole prefix, so a server that drops a result from the *middle* mis-attributes everything after it and only flags the tail — the exact failure the comment says it's preventing.

A length mismatch means we can't trust any of the mapping, not just the tail. Suggest failing the whole call and letting the caller fall back:

```java
if (validations.size() != items.size()) {
  throw new FHIRException(formatMessage(I18nConstants.TX_SERVER_BATCH_RESPONSE_INCOMPLETE, validations.size(), items.size()));
}
```

Note this method is also called by `ValueSetValidator`, `ConceptMapValidator` and `ValueSetRenderer`, so whatever we do here changes their behaviour too — today they get one EXCEPTION message from the `IndexOutOfBoundsException`, and as written they'd get N per-code errors instead. Throwing keeps them where they are.

### `BaseWorkerContext.java` — `validateCodeBatchCS`

See the design question in the summary: `vs == null` already carries this information, and `validateOnServer2` already branches on it. Deriving `csBatch` from `vs == null` fixes the three existing callers as well and avoids adding a method to `IWorkerContext` (which is public API — a non-default method there is a breaking change for anyone implementing the interface outside this repo).

### `CodeSystemValidator.java` — `catch (Exception e)` in `validateSupplementConcepts`

Exception discarded with no log. Please at least `log.debug`/`log.warn` it — a batch failing is exactly the thing we'll want in the tx log when someone reports slow validation.

Also, on a server that's down this does 300 individual failing calls per slice, which is worse than the code being replaced. Consider setting a flag on first batch failure so subsequent slices don't retry the batch (or, since each individual call will also fail, don't fall back at all when the failure looks like a transport error rather than a rejected batch).

### `CodeSystemValidator.java` — the supplement rule

```java
ok = rule(errors, NO_RULE_DATE, IssueType.BUSINESSRULE, cv.getStack().getLiteralPath(), res != null && res.isOk(),
    I18nConstants.CODESYSTEM_CS_SUPP_INVALID_CODE, supp, cv.getCoding().getCode()) && ok;
```

`res.getMessage()` is dropped, so anything that isn't literally "the code isn't in the base code system" — server error, unsupported code system, truncated batch — gets reported to the IG author as if the code were invalid. `ValueSetValidator.executeValidationBatch` passes `cv.getResult().getMessage()` through and calls `processConceptIssues`; worth doing the same here, or at least distinguishing `res.getErrorClass() == SERVER_ERROR` from a genuine invalid code.

### `CodeSystemValidator.java` — `canUseBatchValidation`

Two notes:

- `getTxSupportInfo` returns `LATEST_VERSION` for any code system known locally with content `complete` or `fragment`, so this returns true regardless of what the server can do. Harmless for `complete` (the local pass in `validateCodeBatch` resolves the codes before any server call), but for `fragment` we'll batch to a possibly-old server and lean on the fallback.
- `isSupported()` is already checked by the caller at the top of the supplement block, so no need to re-check — but do note the `catch (Exception e) { return false; }` will also swallow a `TerminologyServiceException` from a server that's genuinely unreachable, which is worth a log line.

### `CodeSystemValidator.java` — `VALIDATION_BATCH_SIZE`

The comment says "Matches ValueSetValidator", but `ValueSetValidator` checks `batch.size() > VALIDATION_BATCH_SIZE`, so its slices are actually 301. Not important in itself, but if we're claiming they match, either fix the VS off-by-one or drop the claim. Better still, hoist the constant somewhere shared — we now have three copies of this and three near-identical `*CodingValidationRequest extends CodingValidationRequest { NodeStack stack; }` classes (`VSCodingValidationRequest`, `CMCodingValidationRequest`, `CSCodingValidationRequest`). A single one with a `NodeStack` would do.

### `CodeSystemValidatorSupplementBatchTest.java` — class comment

> a small, varying number of leading concepts get resolved by the worker context without ever reaching the server

What is the variation, and what drives it? A test with a fake client and a fresh cache shouldn't be nondeterministic, and the loosened assertions (`> 10`, `> 5`, `> 600`, `> answered`) are hiding rather than handling it. `assertEquals(1, server.batchValidateCSCalls)` and the `expectedBatches` assertion aren't loosened, so if this ever swallows a whole slice they'll flake in CI.

### `CodeSystemValidatorSupplementBatchTest.java` — coverage

This only exercises the fake. Could you add a supplement case to fhir-test-cases so the real path against tx.fhir.org is covered too? That's where a `CodeSystem/$batch-validate-code` mismatch would actually show up.

grahamegrieve:
I'm thinking I might implement this

vadi2:
go for it!

grahamegrieve:
redone in #2546

PR #2539, opened 2026-08-05
Fix loading of ManagedWebAccess from fhir-settings

### LINKED ISSUE #2538: fhir-settings.json fields are not used in CLI execution
When run from the CLI, the validator does not correctly load fields from fhir-settings.json

The following should set `ssrfProtectionEnabled` as well as the `allowHttp` and `allowPrivateNetwork` for the server, but does not when a resource is validated.

```json
{
    "ssrfProtectionEnabled": false,
    "servers": [
        {
          "url": "http://myserver",
          "authenticationType": "none",
          "allowHttp": true,
          "allowPrivateNetwork": true
        }
      ]
}
```

dotasek:
Fixes https://github.com/hapifhir/org.hl7.fhir.core/issues/2538


grahamegrieve (approved):
ok

PR #2542, opened 2026-08-10
FML: fix copy() aliasing bug in StructureMap transforms

### LINKED ISSUE #2541: FML: COPY shares object references with source - mutating a copy corrupts the original
## Summary

`copy()` in a StructureMap target transform doesn't clone — it shares child element references with the source. If a later rule overwrites an *existing* field on the copied resource (e.g. `tgtRes.active = true`), that mutation is silently visible on the original source resource, and therefore on *every other* `copy()` taken from it elsewhere in the map (including copies already built and finished earlier, in unrelated groups).

## Steps to reproduce

Map: two entries built from the *same* source Patient (`active: false`) — one plain `copy()`, one `copy()` followed by `tgtRes.active = true`.

```
map "http://example.org/StructureMap/CopyAliasingBug" = "CopyAliasingBug"

uses "http://hl7.org/fhir/StructureDefinition/Bundle" alias Bundle as source
uses "http://hl7.org/fhir/StructureDefinition/Bundle" alias BundleTarget as target

group main(source src : Bundle, target tgt : BundleTarget) {
    // Entry 1: a plain, untouched copy() of the Patient - built and "done" first
    src.entry as e where $this.resource.ofType(FHIR.Patient) -> tgt.entry as te1 then copyPlain(e, te1) "plain-copy";

    // Entry 2: copy() the SAME Patient again, then flip an existing field
    // (active) on THIS copy only
    src.entry as e where $this.resource.ofType(FHIR.Patient) -> tgt.entry as te2 then copyAndMutate(e, te2) "mutate-copy";
}

group copyPlain(source srcEntry, target tgtEntry) {
    srcEntry.resource as srcRes -> tgtEntry.resource = copy(srcRes) "copy";
}

group copyAndMutate(source srcEntry, target tgtEntry) {
    srcEntry.resource as srcRes -> tgtEntry.resource = copy(srcRes) as tgtRes then {
        srcRes -> tgtRes.active = true "set-active";
    } "copy-then-mutate";
}
```

Input Patient: `"active": false`.

**Expected:** `plain-copy` entry keeps `active: false`; `mutate-copy` entry shows `active: true`.

**Actual**:

```json
{
  "resourceType" : "Bundle",
  "entry" : [{
    "resource" : {
      "resourceType" : "Patient",
      "id" : "abc123",
      "active" : true
    }
  },
  {
    "resource" : {
      "resourceType" : "Patient",
      "id" : "abc123",
      "active" : true
    }
  }]
}
```

Both entries show `active: true` — `plain-copy` runs to completion *before* `mutate-copy` even starts and never references it, yet its output is retroactively changed.

Full example: https://github.com/mrunibe/fhir-mapping-tutorial/tree/copy-resource/copy-resource

## Likely cause

`StructureMapUtilities#runTransform`'s `COPY` case returns the source `Element` itself rather than a clone.

mrunibe:
`COPY` in `StructureMapUtilities.runTransform` returned the source `Element` itself instead of a clone, so mutating a copied target silently mutated the source and every other `copy()` taken from it — even ones already finished in unrelated groups. Fixed by returning `val.copy()`.

Fixes #2541

### DIFF
--- a/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/StructureMapUtilities.java
+++ b/org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/StructureMapUtilities.java
@@ -1965,7 +1965,8 @@ public class StructureMapUtilities {
           res.setUserData("profile", tgt.getUserData("profile"));
         return res;
       case COPY:
-        return getParam(vars, tgt.getParameter().get(0));
+        Base val = getParam(vars, tgt.getParameter().get(0));
+        return val != null ? val.copy() : val;
       case EVALUATE:
         ExpressionNode expr = (ExpressionNode) tgt.getUserData(MAP_EXPRESSION);
         if (expr == null) {
--- a/org.hl7.fhir.r4b/src/main/java/org/hl7/fhir/r4b/utils/structuremap/StructureMapUtilities.java
+++ b/org.hl7.fhir.r4b/src/main/java/org/hl7/fhir/r4b/utils/structuremap/StructureMapUtilities.java
@@ -1834,7 +1834,8 @@ public class StructureMapUtilities {
           res.setUserData("profile", tgt.getUserData("profile"));
         return res;
       case COPY:
-        return getParam(vars, tgt.getParameter().get(0));
+        Base val = getParam(vars, tgt.getParameter().get(0));
+        return val != null ? val.copy() : val;
       case EVALUATE:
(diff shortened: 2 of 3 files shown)


grahamegrieve (approved):
ok

grahamegrieve:
why is this a draft?

mrunibe:
> why is this a draft?

It's no longer in draft. Ready to merge @grahamegrieve

PR #2544, opened 2026-08-11
Update fetchTheOldWay urls to https

dotasek:
fetchTheOldWay fetched older IGs via direct URLs, including now blocked http protocol URLs. 

This makes these URLs secure (https) before fetching.


grahamegrieve (approved):
ok

PR #2545, opened 2026-08-11
Add expansionParams as a fileInfo field in ValidationRequest

dotasek:
This adds expansionParams as a FileInfo field in ValidationRequest.

It provides separate handling of the original CLI usage (expansion parameters as a loadable file) as well as handling a ValidationRequest, where it is explicitly included as a FileInfo object.


grahamegrieve (approved):
ok

PR #2546, opened 2026-08-12
Codesystem validation size limit

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/BaseWorkerContext.java
```diff
@@ -1289,6 +1286,7 @@ public void validateCodeBatch(ValidationOptions options, List<? extends CodingVa
         ParametersParameterComponent r = validations.get(i);
 
         if (r.getResource() instanceof Parameters) {
+          checkBatchResultMatches(tc, t, (Parameters) r.getResource(), i);
```
dotasek: issue: Though this looks fine logically, I had to rename a lot of variables for review. This is what I came up with:

```java
 if (items.size() > 0) {
      TerminologyClientContext terminologyClientContext = terminologyClientManager.chooseServer(vs, systems, false, findValidationLanguage(options));
      Parameters resp = processBatch(terminologyClientContext, batch, systems, items.size());
      List<ParametersParameterComponent> validations = resp.getParameters("validation");
      for (int i = 0; i < items.size(); i++) {
        CodingValidationRequest requestAtIndex = items.get(i);
        ParametersParameterComponent responseAtIndex = validations.get(i);

        if (responseAtIndex.getResource() instanceof Parameters) {
          checkBatchResultMatches(terminologyClientContext, requestAtIndex, (Parameters) responseAtIndex.getResource(), i);
          requestAtIndex.setResult(processValidationResult((Parameters) responseAtIndex.getResource(), null, terminologyClientContext.getAddress()));
          if (txCache != null) {
            txCache.cacheValidation(requestAtIndex.getCacheToken(), requestAtIndex.getResult(), TerminologyCache.PERMANENT);
          }
        } else {
          requestAtIndex.setResult(new ValidationResult(IssueSeverity.ERROR, getResponseText(responseAtIndex.getResource()), null).setTxLink(txLog == null ? null : txLog.getLastId()));
        }
      }
    }

...

private void checkBatchResultMatches(TerminologyClientContext terminologyClientContext, CodingValidationRequest requestAtIndex, Parameters responseAtIndex, int index) {
    String system = null;
    String code = null;
    for (ParametersParameterComponent parameter : responseAtIndex.getParameter()) {
      if (parameter.hasValue()) {
        if ("system".equals(parameter.getName())) {
          system = parameter.getValue().primitiveValue();
        } else if ("code".equals(parameter.getName())) {
          code = parameter.getValue().primitiveValue();
        }
      }
    }
    Coding requestCoding = requestAtIndex.getCoding();
    // the system is only checked when we asked with one - with inferSystem, the server picks it
    boolean mismatch = (code != null && !code.equals(requestCoding.getCode()))
        || (system != null && requestCoding.hasSystem() && !system.equals(requestCoding.getSystem()));
    if (mismatch) {
      throw new FHIRException(formatMessage(I18nConstants.TX_SERVER_BATCH_RESPONSE_MISMATCH, terminologyClientContext.getAddress(), index,
          (system == null ? "" : system + "#") + code, requestCoding.getSystem() + "#" + requestCoding.getCode()));
    }
  }
```

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/IWorkerContext.java
```diff
@@ -780,14 +780,48 @@ public interface IWorkerContext {
...
+   * @param codes the codes to validate. Each request carries its own result, so every code keeps its own
+   *          diagnostics. Every request is given a result unless this method throws.
+   * @param vs the value set to validate the codes against - "is this code in this value set". May be
+   *          null to ask only "is this code valid in its own code system", which is the batch equivalent
+   *          of {@link #validateCode(ValidationOptions, Coding, ValueSet)} with a null value set.
+   *          <p>
+   *          Note that where the single code path uses CodeSystem/$validate-code when there is no value
+   *          set, this method always uses ValueSet/$batch-validate-code, and simply omits the url
+   *          parameter when vs is null.
+   * @param notUsed not used - pass null. This was once a flag doing something, but no caller ever set it, and it did not make sense.
```
dotasek, line 818: thought: Appreciate the documentation. A default method and a deprecation would be a good path forward on this.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/testfactory/dataprovider/BaseDataTableProvider.java
```diff
@@ -51,26 +51,47 @@ public String getPrimitiveValue(String elementId, String typeName) throws SQLExc
...
+   * keeps randomly drawn values (unit, version, value etc) coherent with a system that the profile
+   * has already fixed. If no row matches the system, returns null rather than an incoherent row.
+   */
+  public Map<String, String> getComplexValue(String elementId, String typeName, String system) throws SQLException {
     String ek = elements.get(elementId);
     String tk = types.get(typeName);
     if (tk == null) {
       return null;
     }
+    String sf = systemFilter(system);
```
dotasek: issue: `sf` would be clearer as `optionalSystemPredicate`

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/testfactory/dataprovider/BaseDataTableProvider.java
```diff
@@ -51,26 +51,47 @@ public String getPrimitiveValue(String elementId, String typeName) throws SQLExc
...
     if (ek != null) {
-      String sql = "select ValueData from TestValues where ElementKey = "+ek+" and TypeKey = "+tk+(testing ? " ORDER BY ValueKey" : " ORDER BY RANDOM() LIMIT 1" );
+      String sql = "select ValueData from TestValues where ElementKey = "+ek+" and TypeKey = "+tk+sf+(testing ? " ORDER BY ValueKey" : " ORDER BY RANDOM() LIMIT 1" );
```
dotasek: note: The sf string is an optional predicate here which introduces its own `AND`. It works, but initially looks like a concatenation for `tk` or `typeKey`.

org.hl7.fhir.utilities/src/test/java/org/hl7/fhir/utilities/regex/RegexTimeoutTests.java
```diff
@@ -36,8 +36,8 @@ static void beforeAll()
 
       @AfterAll
       static void afterAll() throws InterruptedException {
```
dotasek: issue: Since RegexTimeout no longer uses threads to do its work, the `@BeforeAll`,`@AfterAll`, and related fields can be removed altogether.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/ValueSetValidator.java
```diff
@@ -1,1049 +1,1065 @@
...
+   * request as though it had none - which sends a Czech display check to the default server rather than to
+   * the one that is authoritative for Czech.
+   * <p>
+   * withLanguage() copies (getValidationOptionsLanguage never returns null), which matters because the
+   * withExampleOK/setDisplayWarningMode/withExternalSource that follow all mutate in place - and settings
+   * is shared with every other validator in the run.
+   */
+  private ValidationOptions txOptions(NodeStack stack, ValueSet externalSource) {
+    String lang = ((InstanceValidator) parent).getValidationOptionsLanguage(stack);
+    return settings.withLanguage(lang).withExampleOK().setDisplayWarningMode(true).withExternalSource(externalSource);
```
dotasek, line 526: note: this is out of scope for this PR, but ValidatonOptions varies how `with` methods are applied. Some generate a copy of validationOptions in which the new value is set (`withLanguage`), while others (`withLanguageOK`, `withExternalSource`) mutate the original object. 

We probably get lucky here in that the first one is a copy, but order matters here, which may cause weird side-effects elsewhere.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/CodeSystemValidator.java
```diff
@@ -847,16 +907,38 @@ private boolean hasHierarchy(Element cs) {
...
     }
+    return res;
+  }
 
+  private boolean executeSupplementBatch(List<ValidationMessage> errors, List<CSCodingValidationRequest> batch, String supp, ValidationOptions options) {
+    boolean ok = true;
+    if (!batch.isEmpty()) {
+      context.validateCodeBatch(options, batch, null, false);
+      for (CSCodingValidationRequest cv : batch) {
+        ok = rule(errors, NO_RULE_DATE, IssueType.BUSINESSRULE, cv.getStack().getLiteralPath(), cv.getResult().isOk(),
```
dotasek, line 937: issue: If this is correct, only the last validation request value is returned. Shouldn't this be ok only if all pass or one passes?

dotasek (changes requested):
See comments.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/CodeSystemValidator.java
grahamegrieve: I think you're wrong on it - && ok at the end

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/ValueSetValidator.java
grahamegrieve: we should definitely fix that

org.hl7.fhir.utilities/src/test/java/org/hl7/fhir/utilities/regex/RegexTimeoutTests.java
grahamegrieve: done

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/testfactory/dataprovider/BaseDataTableProvider.java
grahamegrieve: I don't understand

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/CodeSystemValidator.java
dotasek: Ah, it's at the end. Was expecting it at the beginning.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/context/IWorkerContext.java
grahamegrieve: ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/testfactory/dataprovider/BaseDataTableProvider.java
dotasek: When reading, it looks like TypeKey = "+tk+sf+... as if tk+sf is the TypeKey, whereas sf is it's own separate predicate that adds an AND if it's used.

dotasek (approved):

PR #2548, opened 2026-08-12
npm protection fixes

grahamegrieve:


ahurbean:
The code is going in the right direction, but the issue still persists for callers of NpmPackage.fromPackage due to defaultMaxLoadSize defaulting to 0 (no limit). The resulting behavior if callers don't also update is the same as before.

We checked, and none of the current callers in hapi-fhir or hapi-fhir-jpaserver-starter use the new overload or set the default, including the $install operation we PoC'd against. So as merged, this has no effect on our target.

We would recommend a safe/sane default that caps the maximum size of loaded files to something sensible that would not be reached during normal usage (for example even 300 MB, although even this is arguably quite large).

grahamegrieve:
that isn't necessarily sane at all. If I do that, then I have to chase down every invocation and give the users a configuration to override it. When it's perfectly appropriate for them to get an out of memory 

HAPI should take advantage of it

ahurbean:
Fair point on not wanting to force a hard cap that breaks legitimate large-file workflows for CLI-style, operator-driven usage. That's not really what we're asking for though.

The distinction we'd draw is between an operator pointing the tool at their own file, where an OOM is their own problem, and a server accepting npmContent from an anonymous caller, where it isn't since the impact lands on other users of the service, not the caller.

An even more generous default (1-2GB, say) wouldn't affect the first case at all, no real IG package comes close to that, while still closing off the second. setDefaultMaxLoadSize(0) already exists as an opt-out for anyone who explicitly wants the old unbounded behavior, so this doesn't require chasing down every invocation.

On HAPI taking advantage of it: agreed, and since neither side currently does, would you be open to us raising this directly with the hapi-fhir maintainers, referencing this new API?

grahamegrieve:
@dotasek do you have an opinion? It's hard to figure out an upper limit - the biggest published package is ~100MB unzipped but there are people working with much bigger resources than that - the biggest I have heard of is 2-3GB in size. (I think that's unworkable, but people have themselves a hammer)

grahamegrieve:
but you should certainly ask HAPI to make use of the limit, yes. They'll want a different limit that anything that makes sense in a CLI anyway

ahurbean:
imo, if we can work with HAPI to utilize the new API for loading NPM packages in a bounded fashion (with a sensible upper bound) this fix would be fine, as the responsibility for safe loading would be shifted to callers of the function.

PR #2550, opened 2026-08-13
Fix issues in ValidationOptions with... methods + cleanup

### LINKED ISSUE #2547: ValidationOptions has inconsistent implementation of `with...` pattern
In ValidationOptions the `with...` methods should be constructing a copy of the original object before changing a value in the copy. However, some methods do not:

https://github.com/hapifhir/org.hl7.fhir.core/blob/d7216f7e2a27738d20d0ee984bfd5593f1bf7860/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/validation/ValidationOptions.java#L338

https://github.com/hapifhir/org.hl7.fhir.core/blob/d7216f7e2a27738d20d0ee984bfd5593f1bf7860/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/validation/ValidationOptions.java#L403

This should be reviewed, and all methods mutating the original object should be fixed.

dotasek:
Fixes #2547

### DIFF
--- a/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/validation/ValidationOptions.java
+++ b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/validation/ValidationOptions.java
@@ -34,15 +34,15 @@ public class ValidationOptions {
 
   private AcceptLanguageHeader langs = null;
   /**
-   *  true (which is default) means that the validator will use the server if it needs to for this request
-   *  (There are a few corner cases where you want to turn this off?)
+   *  true (default) means that the validator will try to resolve the terminology using a server if it needs to for this
+   *  request. (There are a few use cases, both in the code and externally where this is needed)
    */
   @Getter
   private boolean useServer = true;
 
   /**
-   *  True means that the validator will try to resolve the terminology request locally with the resources at hand
-   *  There have been a few corner case code systems where the local code system from THO is wrong
+   *  true (default) means that the validator will try to resolve the terminology request locally with the resources at
+   *  hand (There are a few use cases, both in the code and externally where this is needed)
    */
   @Getter
   private boolean useClient = true;
@@ -57,14 +57,15 @@ public class ValidationOptions {
   private boolean displayWarningMode = false;
   private boolean vsAsUrl;
   /**
-   *  Don't know exactly what this does -Grieve
+   *  If the version requested has to match the version actually found, or not. It's used for e.g. display, where you
+   *  pass a version in, because you have one, but you don't mind getting a display from a different version instead
    */
   @Getter
   private boolean versionFlexible = true;
   @Getter
   private boolean useValueSetDisplays;
   /**
-   *  if the language is other than english, should the validator accept english as well? -Grieve
+   *  If true and the language is other than english, the validator will accept english as well
    */
   @Getter
   private boolean englishOk = true;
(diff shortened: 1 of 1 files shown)


org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/validation/ValidationOptions.java
```diff
@@ -35,20 +35,51 @@ public static R5BundleRelativeReferencePolicy fromCode(String code) {
   }
 
   private AcceptLanguageHeader langs = null;
+  /**
+   *  true (which is default) means that the validator will use the server if it needs to for this request
+   *  (There are a few corner cases where you want to turn this off?)
```
grahamegrieve: Yes there are internal uses of the API where the code wants to manage the flow, and some external uses

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/validation/ValidationOptions.java
```diff
@@ -35,20 +35,51 @@ public static R5BundleRelativeReferencePolicy fromCode(String code) {
...
+   */
+  @Getter
   private boolean useServer = true;
+
+  /**
+   *  True means that the validator will try to resolve the terminology request locally with the resources at hand
+   *  There have been a few corner case code systems where the local code system from THO is wrong
```
grahamegrieve: That doesn't make sense- you'd want to not use the wrong resource. This is the inverse of the one above for the same reason.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/validation/ValidationOptions.java
```diff
@@ -35,20 +35,51 @@ public static R5BundleRelativeReferencePolicy fromCode(String code) {
...
+   */
+  @Getter
   private boolean guessSystem = false;
+  @Getter
   private boolean membershipOnly = false;
+  @Getter
   private boolean displayWarningMode = false;
   private boolean vsAsUrl;
+  /**
+   *  Don't know exactly what this does -Grieve
```
grahamegrieve: If the version requested has to match the version actually found, or not. It's used for e.g. display, where you pass a version in, because you have one, but you don't mind getting a display from a different version instead

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/validation/ValidationOptions.java
```diff
@@ -35,20 +35,51 @@ public static R5BundleRelativeReferencePolicy fromCode(String code) {
...
+   */
+  @Getter
   private boolean versionFlexible = true;
+  @Getter
   private boolean useValueSetDisplays;
+  /**
+   *  if the language is other than english, should the validator accept english as well? -Grieve
```
grahamegrieve: yes that

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/validation/ValidationOptions.java
```diff
@@ -91,82 +119,12 @@ public boolean hasLanguages() {
...
-  }
-  public boolean isNoAbstract() {
-    return noAbstract;
-  }
-
-  /**
-   * Don't know what this does
-   * 
-   * @return
+   * Don't know what this does -Grieve
```
grahamegrieve: tells the engine to pass the ValueSet to the server as a URL rather than the whole VS. But I don't remember why this exists now

grahamegrieve:
Already handled elsewhere. thanks

PR #2551, opened 2026-08-13
2026 08 gg r5 clean up

grahamegrieve:


grahamegrieve:
@dotasek ignore the R6 part of this - the work is for R6 but those files are incomplete - that'll be my next PR

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackage.java
dotasek, line 572: issue: This could just call `fromPackage(tgz, desc, progress, defaultMaxLoadSize)` so we ensure both methods are using the exact same logic otherwise.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackage.java
```diff
@@ -572,7 +574,18 @@ public static NpmPackage fromPackage(InputStream tgz, String desc, boolean progr
     res.readStream(tgz, desc, progress);
     return res;
   }
-  
+
+  /**
+   * Load a package, enforcing the given limit on the total decompressed size of the package
+   * (in bytes, 0 = no limit) instead of the default limit (see {@link #setDefaultMaxLoadSize(long)})
+   */
+  public static NpmPackage fromPackage(InputStream tgz, String desc, boolean progress, long maxLoadSize) throws IOException {
```
dotasek, line 582: question: This is only called from tests as far as I can see. Is this intended to be called by downstream projects?

dotasek (changes requested):
Per request, this only reviews the changes in 318bf68cc18c03b858f60a5b94f2191ef6836f9e#r3786972206 and treats the rest of the PR as non-functional refactoring.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackage.java
grahamegrieve: that wouldn't be the same though - because the first one doesn't override the default maxLoadSize, and the second does.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackage.java
grahamegrieve: yep. they can choose - set the default, or call this directly

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/NpmPackage.java
dotasek: OK. It's three lines of repeated code. I would prefer it be refactored, but it is non-blocking.

dotasek (approved):

PR #2552, opened 2026-08-14
R6 initial commit

grahamegrieve:


dotasek (approved):
Approved for merge.

Note, this is NOT a full review. 

The files and configurations committed in this PR are not accessed from outside the new  modules.

org.hl7.fhir.model is only accessible within:

- org.hl7.fhir.convertors
- org.hl7.fhir.services
- org.hl7.fhir.core.generator

org.hl7.fhir.core.generator and org.hl7.fhir.services use org.hl7.fhir.model, but are self-contained (no imports of either in other modules).

org.hl7.fhir.convertors contains conversion code using the new models, but the functional code in the following packages is not accessible from other modules:

- `org.hl7.fhir.convertors.conv*_N`
- `org.hl7.fhir.convertors.context.ConversionContext*_N`
- `org.hl7.fhir.convertors.advisors.impl.BaseAdvisor*_N`

PR #2555, opened 2026-08-19
Dependency updates

dotasek:
jackson 2.22.1 -> 2.22.2
kotlin-stdlib 2.4.0 -> 2.4.10


pom.xml
madduci, line 35: couldn't also these dependencies be upgraded?

madduci: couldn't also these dependencies be upgraded? 1.6.3 solves many CVEs

dotasek: I will address these in another PR. Updating the build plugins is another maintenance task.

dotasek: Addressed logback in latest commit, thank you.

PR #2560, opened 2026-08-22
PropertyFilter: test every value of a repeating concept property, not just the first

### LINKED ISSUE #2559: ValueSet expansion: property filter only tests a concept's FIRST matching concept.property, silently under-matching repeating properties
### Summary

`PropertyFilter` matches a compose filter against only the **first** `concept.property` with the
filtered code, so a concept whose matching value appears in any later position is silently excluded
from the expansion.

`CodeSystem.concept.property` is `0..*`, and a filter of the form *"property `X` = `Y`"* should be
satisfied when the concept **has** a property `X` with value `Y`. Today the result depends on the
order the properties happen to be serialised in.

### Affected versions

Reproduced on **6.9.12** and **6.10.2** (latest at time of writing). The method is unchanged between
them.

### Cause

`org.hl7.fhir.r5.terminologies.expansion.PropertyFilter`:

```java
private ConceptPropertyComponent getPropertyForConcept(ConceptDefinitionComponent def) {
  for (ConceptPropertyComponent pc : def.getProperty()) {
    if (pc.hasCode() && pc.getCode().equals(property.getCode())) {
      return pc;              // <-- first match only; later values are unreachable
    }
  }
  return null;
}
```

`includeConcept` then tests that single component, so every value after the first is never consulted.
The same applies to the `Coding` branch via `CodingUtilities.filterEquals`.

### Minimal reproduction

Two concepts carrying identical data — `kind=target` and `kind=other` — differing only in the order
of the two values. Only the one where `target` comes first is returned.

```java
SimpleWorkerContext ctx = new SimpleWorkerContext.SimpleWorkerContextBuilder()
    .withAllowLoadingDuplicates(true).build();
ctx.setExpansionParameters(new Parameters());

CodeSystem cs = new CodeSystem();
cs.setUrl("http://example.org/cs").setVersion("1").setName("Demo");
cs.setStatus(PublicationStatus.ACTIVE);
cs.setContent(CodeSystemContentMode.COMPLETE);
cs.addProperty().setCode("kind").setType(CodeSystem.PropertyType.CODE);

CodeSystem.ConceptDefinitionComponent c1 = cs.addConcept().setCode("first");
c1.addProperty().setCode("kind").setValue(new CodeType("target"));
c1.addProperty().setCode("kind").setValue(new CodeType("other"));

CodeSystem.ConceptDefinitionComponent c2 = cs.addConcept().setCode("second");
c2.addProperty().setCode("kind").setValue(new CodeType("other"));
c2.addProperty().setCode("kind").setValue(new CodeType("target"));   // same data, later position
ctx.cacheResource(cs);

ValueSet vs = new ValueSet();
vs.setUrl("http://example.org/vs").setVersion("1").setName("Demo");
vs.setStatus(PublicationStatus.ACTIVE);
vs.getCompose().addInclude().setSystem("http://example.org/cs")
  .addFilter().setProperty("kind").setOp(FilterOperator.EQUAL).setValue("target");
ctx.cacheResource(vs);

ValueSetExpansionOutcome out = ctx.expandVS(new ExpansionOptions(), vs);
// prints the expanded codes
```

**Actual:** `first`
**Expected:** `first second`

### Why it is easy to miss

The expansion succeeds, reports no error or warning, and `expansion.total` agrees with the truncated
`contains` — so the result looks internally consistent. On real content the size of the under-match
depends entirely on serialisation order.

Measured on LOINC 2.82 held locally: the filter
`CLASS = http://loinc.org/part#LP7787-7` returned **7,762** concepts where the code system contains
**28,614** carrying that value — the 7,762 being those where it happened to sort first. LOINC concepts
routinely carry several `CLASS` values (the class ancestry).

### Suggested fix

Test every property with the filtered code rather than the first, i.e. collect the matches and apply
the existing per-value logic across them: satisfied if **any** value matches for the positive
operators, and for the negating operator (`NOTIN`) only if **every** value fails to match. Behaviour
for concepts with a single value for the property is unchanged.

I'm happy to open a PR along those lines if that would help.

### Note on workarounds

There is no extension point for this: `PropertyFilter` is constructed inline in
`ValueSetExpander.doInternalIncludeCodes`, which is private, on an expander whose state is
package-private — so it can be neither subclassed nor substituted. A downstream consumer's only
option is to shadow the class on the classpath, which is what we currently do, pinned to a specific
core version.

igorboss:
Fixes #2559.

`CodeSystem.concept.property` is `0..*`, and a compose filter on a property is satisfied when the concept **has** a matching value. `getPropertyForConcept` returned the *first* property carrying the filtered code and `includeConcept` tested only that one, so every later value was unreachable — making the outcome depend on the order the properties happen to be serialised in.

It fails in **both** directions, and the second is the one I'd flag:

- **`EQUAL` / `IN` / `REGEX` under-match** — a concept whose matching value isn't first is dropped. On LOINC 2.82 held locally, `CLASS = http://loinc.org/part#LP7787-7` returned **7,762 of the 28,614** concepts that carry that value, with no error and a self-consistent `total`.
- **`NOTIN` over-matches** — a concept that *does* carry an excluded value is **included**, because only its first value was examined. A value set can contain a code it explicitly excluded.

## The change

Every value with the filtered code is tested. The positive operators are satisfied by any one value; `NOTIN` requires every value to pass, preserving its meaning of "the concept has no value in this list". A concept carrying one value, or none, behaves exactly as before.

The diff is larger than the logic because removing the vestigial `else` block de-indents the existing `switch` body — the actual change is the loop in `includeConcept` plus `getPropertyForConcept` → `getPropertiesForConcept`. Reviewing with whitespace ignored shows it clearly.

## Verification

Compiled against 6.10.2 and run with only this class replaced, on a concept holding `kind=target` and `kind=other`:

| filter | before | after |
|---|---|---|
| `kind = target` | `first` | `first second` |
| `kind not-in target` | `second` | *(none)* |

## Regression test

Proposed separately as HL7/fhir-tx-ecosystem-ig#65 — it adds a repeating property to the `simple` code system and filters on the value that is not first. `TerminologyServiceTests` picks that up through `hl7.fhir.uv.tx-ecosystem`, so this fix gets covered there rather than by a new unit test here. That test fails on stock 6.9.12 (`total = 0`) and passes with this change (`total = 1`).

## One related thing, not fixed here

`expansion.contains.property` also collapses a repeating property to a single value, and keeps the **last** one — with `property=kind` requested, the concept above comes back as `first[kind=other]`. That's a different code path from the filter (which keeps the first), so I've left it alone rather than widen this PR. Happy to take it on separately if you'd like it fixed.

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/expansion/PropertyFilter.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/expansion/PropertyFilter.java
@@ -1,5 +1,6 @@
 package org.hl7.fhir.r5.terminologies.expansion;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.TimeoutException;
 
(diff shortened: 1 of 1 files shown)


grahamegrieve:
this was resolved in the other PR

PR #2562, opened 2026-08-24
Improved Measure and MeasureReport validation

brynrhodes:
Added R6 group-level scoring validation
Added support for linkId/id fallback for R4 backwards-compatibility

These changes are covered/tested in the following fhir-tests-cases PR:

https://github.com/FHIR/fhir-test-cases/pull/277


org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
```diff
@@ -2355,7 +2355,7 @@ private boolean checkMaxValueSet(List<ValidationMessage> errors, String path, El
           if (vr.getErrorClass() != null && vr.getErrorClass().isInfrastructure())
             txWarning(errors, NO_RULE_DATE, vr.getTxLink(), vr.getDiagnostics(), IssueType.CODEINVALID, element.line(), element.col(), path, false, I18nConstants.TERMINOLOGY_TX_NOVALID_9, describeReference(maxVSUrl, valueset, BindingContext.BASE, null), vr.getMessage(), value);
           else {
-            ok = txRule(errors, NO_RULE_DATE, vr.getTxLink(), vr.getDiagnostics(), IssueType.CODEINVALID, element.line(), element.col(), path, false, I18nConstants.TERMINOLOGY_TX_NOVALID_11, describeReference(maxVSUrl, valueset, BindingContext.BASE, null), vr.getMessage()) && ok;
+            ok = txRule(errors, NO_RULE_DATE, vr.getTxLink(), vr.getDiagnostics(), IssueType.CODEINVALID, element.line(), element.col(), path, false, I18nConstants.TERMINOLOGY_TX_NOVALID_11, describeReference(maxVSUrl, valueset, BindingContext.BASE, null), null, value, vr.getMessage()) && ok;
```
grahamegrieve, line 2358: why this change?

brynrhodes: When I started adding the changes, I was on an older version of the validator and that code path was getting hit and the parameters are wrong for the message, so I fixed the parameters. Then when I merged in main, that code path was no longer hit. But I'm still confident that's the right fix for that message, if it is error does get hit again.

grahamegrieve (approved):
ok

PR #2563, opened 2026-08-24
Remove name from Swedish editor entry

annarossander:
Removed the name associated with the Swedish (SV) editor entry. Joakim Berg no longer works with FHIR.


grahamegrieve (approved):
ok

PR #2569, opened 2026-08-25
CompareUtilities: implement the $only$ version-gate marker

### LINKED ISSUE #2566: txTests: CompareUtilities does not implement the $only$ marker used by tx-ecosystem's translate-1
`tests/translate/translate-1-response-parameters.json` in HL7/fhir-tx-ecosystem-ig (current master) now uses an `$only$` marker to version-gate individual match parts:

```json
{ "$only$": "version:4", "name": "equivalence", "valueCode": "equivalent" },
{ "name": "originMap", "valueCanonical": "http://hl7.org/fhir/test/ConceptMap/full|0.1.0" },
{ "$only$": "version:5", "name": "relationship", "valueCode": "equivalent" },
{ "$only$": "version:4", "name": "source", "valueCanonical": "..." },
```

`CompareUtilities` (`org.hl7.fhir.r5.test.utils`, 6.10.2 and current master) implements `$optional$`, `$optional-properties$`, `$count-arrays$` etc., but has no handling for `$only$`. The `$only$`-marked entries are therefore treated as required array items, and the comparison fails on the raw count before any per-item matching:

```
-- translate-1: Fail
   array item count differs at .parameter[0].part
Expected:"6" for translate-1
Actual  :"4"
```

Since the expected file mixes `version:4` and `version:5` entries, no server of either version can produce all six parts — the test is currently unsatisfiable for every implementation. Either `CompareUtilities` should learn `$only$` (filter entries whose gate doesn't match the server version before comparing, the way `$optional$` entries are tolerated), or the expectation should fall back to `$optional$` markers until it does.

Observed with validator 6.10.2 running txTests in general mode against an independent terminology server implementation; verified by experimentally emitting the new `sourceConcept` part as well — an R5 server's part list then reads `concept, originMap, relationship, sourceConcept` and the count still stops at 4.

igorboss:
## Problem

Fixes #2566.

`tests/translate/translate-1-response-parameters.json` in the tx-ecosystem suite (current master) version-gates individual match parts with an `$only$` marker (`"version:4"` / `"version:5"`), but `CompareUtilities` only implements `$optional$`/`$optional-properties$`/`$count-arrays$`. The `$only$` items are treated as required for every server, so the comparison fails on the raw array count (`Expected:"6" Actual:"4"`) regardless of server version — the expected file mixes both versions' parts, so no server of either version can satisfy it.

## Change

Before comparing an expected array, filter out items whose `$only$` gate does not match the current context — the same token grammar `$optional$` already uses (`version:N` against the `version` variable the tx tester supplies from the server's reported FHIR version, mode names, `!` negation). Unlike `$optional$`, an item whose gate *does* match stays **required**. A property whose expected array is gated away entirely may be absent from the actual response, and the marker property itself is ignored during object comparison.

Unit tests added in `CompareUtilitiesTests` covering: gate matching per version, a matching-gate item being required, an other-version item being rejected as unexpected, and a fully-gated-away property.

## Evidence

Against an independent R5 terminology server (tests: tx-ecosystem current master content):

* before: `translate-1` fails `array item count differs at .parameter[0].part — Expected:"6" Actual:"4"` no matter what the server returns;
* after: the comparison correctly reduces the expectation to the four R5 parts (`concept`, `originMap`, `relationship`, `sourceConcept`), and with the server producing them, `translate-1` **passes**; a server missing one now gets a genuine diff instead of an unsatisfiable count mismatch.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,6 @@
 ## Validator Changes
 
-* no changes
+* txTests: support the `$only$` version-gate marker in expected responses (used by the tx-ecosystem test cases)
 
 ## Other code changes
 
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/test/utils/CompareUtilities.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/test/utils/CompareUtilities.java
@@ -378,7 +378,7 @@ public class CompareUtilities extends BaseTestingUtilities {
   }
 
   private boolean isOptional(String n, List<String> optionals) {
-    return n.equals("$optional$") || optionals.contains("*")  || optionals.contains(n);
+    return n.equals("$optional$") || n.equals("$only$") || optionals.contains("*")  || optionals.contains(n);
   }
 
   private boolean allOptional(JsonElement value) {
@@ -387,7 +387,8 @@ public class CompareUtilities extends BaseTestingUtilities {
       for (JsonElement e : a) {
         if (e.isJsonObject()) {
           JsonObject o = e.asJsonObject();
-          if (!o.has("$optional$")) {
+          boolean filteredOnly = o.isJsonString("$only$") && !passesOptionalFilter(o.asString("$only$"));
+          if (!o.has("$optional$") && !filteredOnly) {
             return false;
           }
         } else {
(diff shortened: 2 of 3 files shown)


grahamegrieve:
patience. This is all work in progress - the core PR is coming in the next 24 hours

PR #2570, opened 2026-08-25
txTests: evaluate test-level version gates against the server's FHIR version

### LINKED ISSUE #2567: txTests: test-level version gates never engage - -test-version is the tests-package version, and the '!4.0' / partial criteria are not handled
`tests/test-cases.json` (HL7/fhir-tx-ecosystem-ig, current master) now version-gates whole test entries, e.g. the split reverse-translation pair:

```json
{ "name" : "translate-reverse-r5+", "version": "!4.0", ... },
{ "name" : "translate-reverse-r4", "version": "4.0", ... }
```

`TxTester.passesVersion` is supposed to skip the variant that doesn't apply, but the gate never engages:

1. The `version` field it compares against is wired from the CLI's `-test-version` option - the version of the *tests package* to load (default null → `hl7.fhir.uv.tx-ecosystem#current`), not a FHIR version. With the default, `version == null` and `passesVersion` returns true for everything, so every version-gated variant runs against every server. Observed with 6.10.2 and current master: `translate-reverse-r4` (gated `"version": "4.0"`) is executed against a server whose /metadata reports fhirVersion 5.0.0, and necessarily fails.
2. Even when a FHIR-shaped version is supplied (as the in-repo `ExternalTerminologyServiceTests` does - it passes "5.0.0" to the constructor), the comparison is inverted: `VersionUtilities.versionMatches(version, item.asString("version"))` passes the *gate* as the semver candidate, so `"!4.0"` throws ("Invalid candidate") and a plain `"4.0"` gate can never partial-match a concrete "4.0.1".
3. Nothing implements the `!` negation the new entries use.

Presumably the gate should be evaluated against the server's actual FHIR version - which TxTester already determines during connect ("Server version 5.0.0 from /metadata") - with the entry's value as the criteria and the `!` prefix honoured. Until then the new split tests are unsatisfiable as a set: whatever the server's version, the variant for the other version runs too and fails.

Related: #2564 (the split translate-reverse fixture files missing from the IG), #2565, #2566 - found running the same suite against an independent terminology server implementation.

igorboss:
## Problem

Fixes #2567.

test-cases.json now version-gates whole test entries (`"version": "4.0"` / `"!4.0"` on the split `translate-reverse-r4` / `-r5+` pair; `4.0`/`5.0` gates already existed in the UCUM and bugs suites). The gate never engages:

1. `passesVersion` compares against the `version` constructor argument, which the CLI wires from `-test-version` — the version of the *tests package* to load (null by default) — so with defaults every gated variant runs against every server: `translate-reverse-r4` executes against a server whose /metadata reports 5.0.0, and necessarily fails.
2. Even with a FHIR-shaped version supplied (as `ExternalTerminologyServiceTests` does), the arguments are inverted — the gate is parsed as the semver *candidate*, so `"!4.0"` throws and `"4.0"` can never partial-match a concrete `4.0.1`.
3. Nothing implements the `!` negation.

## Change

* `TxTester` records the FHIR version the server reports during connect (`$versions`, falling back to /metadata) and `passesVersion` evaluates the entry's gate as the criteria against it (explicit constructor version still wins, preserving the `ExternalTerminologyServiceTests` usage), honouring the `!` prefix and partial criteria (`4.0` matches `4.0.x`).
* `TxTestsCommand` stops passing the tests-package version into the gate.

## Evidence

Against an independent server reporting 5.0.0 (tests: tx-ecosystem current master content): before, `translate-reverse-r4` runs and fails (its R4-form request means nothing to an R5 server); after, it is skipped and the run's test count drops accordingly, while `translate-reverse-r5+` still runs.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,6 @@
 ## Validator Changes
 
-* no changes
+* txTests: evaluate test-level `version` gates against the server's reported FHIR version (previously the tests-package version was compared, so gated tests ran against every server), and support the `!` negation prefix
 
 ## Other code changes
 
--- a/org.hl7.fhir.validation.cli/src/main/java/org/hl7/fhir/validation/cli/picocli/commands/TxTestsCommand.java
+++ b/org.hl7.fhir.validation.cli/src/main/java/org/hl7/fhir/validation/cli/picocli/commands/TxTestsCommand.java
@@ -100,13 +100,15 @@ public class TxTestsCommand extends ValidationServiceCommand implements Callable
       // Load externals if provided
       JsonObject externalsJson = loadExternals(externals);
 
-      // Create TxTester instance
+      // Create TxTester instance. The last argument is the FHIR version used to
+      // evaluate test-level "version" gates - not the tests-package version, so
+      // pass null and let TxTester use the version the server reports.
       TxTester txTester = new TxTester(
         new TxTester.InternalTxLoader(version),
         tx,
         false,
         externalsJson,
-        testVersion
+        null
       );
 
       // Add input loaders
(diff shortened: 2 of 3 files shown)


grahamegrieve:
this functionality is in https://github.com/hapifhir/org.hl7.fhir.core/pull/2571

PR #2577, opened 2026-08-28
fix: terminate XhtmlParser.readUntil at end-of-input

### LINKED ISSUE #2576: XhtmlParser.readUntil never terminates at EOF - unbounded allocation on truncated narrative
### Summary

`XhtmlParser.readUntil` cannot terminate at end-of-input. Both overloads guard on
`peekChar() != 0`, while every other loop in the class guards on
`peekChar() != END_OF_CHARS`. Since `END_OF_CHARS` is `(char) -1` (i.e. `￿`) and never
`0`, the loop condition stays true forever once the reader is exhausted — and because
`peekChar()`/`readChar()` return `END_OF_CHARS` *without consuming anything*, each iteration
appends another `￿` to the `StringBuilder` and makes no progress. The builder grows
until the heap is gone.

Observed on `org.hl7.fhir.utilities` **6.9.12**; the same two guards are present in 6.9.9,
so this is not a recent regression.

### The code

`XhtmlParser.java:63`
```java
private static final char END_OF_CHARS = (char) -1;
```

`XhtmlParser.java:1004-1021` — the two guards in question:
```java
private String readUntil(char ch) throws IOException {
  StringBuilder s = new StringBuilder();
  while (peekChar() != 0 && peekChar() != ch)     // <-- should be END_OF_CHARS
    s.append(readChar());
  readChar();
  return s.toString();
}

private String readUntil(String sc) throws IOException {
  StringBuilder s = new StringBuilder();
  while (peekChar() != 0 && sc.indexOf(peekChar()) == -1)   // <-- should be END_OF_CHARS
    s.append(readChar());
  readChar();
  return s.toString();
}
```

`peekChar()` (`:824-841`) returns `END_OF_CHARS` when `!rdr.ready()` or `rdr.read() == -1`,
leaving `cache` empty — so it is stable at EOF and the loop makes no progress.

For contrast, the surrounding code gets this right, e.g. `:890`:
```java
} else if (c != END_OF_CHARS)
```

### How it is reached

`readUntil(String)` has a single caller, `parseLiteral` (`:1026`):
```java
String c = readUntil(";&'\"><");
```
and `parseLiteral` is called from ordinary narrative parsing — element text content
(`:671`) and attribute values (`:769`).

So any XHTML narrative containing an entity reference that reaches end-of-input before any
of `;`, `&`, `'`, `"`, `>` or `<` enters the non-terminating loop. In practice that means a
**truncated `<div>`** — a few bytes are enough.

`mustBeWellFormed` does not help: the loop never exits to reach a well-formedness check.

### Impact

The failure surfaces as `OutOfMemoryError`, which is an `Error` rather than an `Exception`,
so the `catch (Exception e)` guards that calling code typically wraps parsing in do not
contain it. In a service that parses narrative from submitted resources — a validation
endpoint, for instance — a very small malformed input can exhaust the heap. We measured OOM
in roughly 300 ms from a minimal case.

Flagging that availability angle for prioritisation rather than to overstate it; the fix
itself looks like a two-token change.

### Suggested fix

Guard both overloads on `END_OF_CHARS`, matching the rest of the class:

```java
while (peekChar() != END_OF_CHARS && peekChar() != ch)
while (peekChar() != END_OF_CHARS && sc.indexOf(peekChar()) == -1)
```

Worth fixing both even though `readUntil(char)` currently has no callers — it carries the
identical defect and would be a trap for the next caller.

Happy to raise a PR if that is useful.

jimsteel:
Fixes #2576.

### The defect

Both `readUntil` overloads guarded on `peekChar() != 0`, while every other loop in `XhtmlParser` guards on `peekChar() != END_OF_CHARS`. `END_OF_CHARS` is `(char) -1` and never `0`, so at end-of-input the condition stayed true — and since `peekChar()`/`readChar()` return `END_OF_CHARS` *without consuming*, each iteration appended another `(char) -1` and made no progress. The `StringBuilder` grew until the heap was exhausted.

It is reached from `parseLiteral`, which handles entity references in both element text and attribute values, so an XHTML narrative whose entity reference runs into end-of-input triggers it. A handful of bytes is enough. The failure arrives as `OutOfMemoryError` — an `Error`, not an `Exception` — so it escapes the `catch (Exception e)` guards callers typically wrap parsing in.

### The fix

Guard both overloads on `END_OF_CHARS`, matching the rest of the class. `readUntil(char)` has no callers today but carries the identical defect, so both are changed.

### Tests

Added to `XhtmlParserTests`, covering both post-fix outcomes rather than assuming one:

| input | behaviour after the fix |
|---|---|
| `<div…>&` | `FHIRFormatError` |
| `<div… title="&` | `FHIRFormatError` |
| `<div…>text &amp` | parses leniently, terminates |
| `<div…><p>&#3` | parses leniently, terminates |
| `<div…>a &amp; b</div>` | unchanged |

The last case guards against an over-tight EOF check breaking ordinary entities.

Verified these are genuine regression tests by reverting the fix and re-running: the suite then **kills the forked JVM** (`Tests run: 0`, `MojoExecutionException`) rather than reporting failures, which is the OOM taking the VM down.

`mvn -pl org.hl7.fhir.utilities test -Dtest='Xhtml*'` → 48 tests, 0 failures.

### DIFF
--- a/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/XhtmlParser.java
+++ b/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/xhtml/XhtmlParser.java
@@ -1004,7 +1004,7 @@ public class XhtmlParser {
   private String readUntil(char ch) throws IOException
   {
     StringBuilder s = new StringBuilder();
-    while (peekChar() != 0 && peekChar() != ch)
+    while (peekChar() != END_OF_CHARS && peekChar() != ch)
       s.append(readChar());
     readChar();
     return s.toString();
@@ -1014,7 +1014,7 @@ public class XhtmlParser {
   private String readUntil(String sc) throws IOException
   {
     StringBuilder s = new StringBuilder();
-    while (peekChar() != 0 && sc.indexOf(peekChar()) == -1)
+    while (peekChar() != END_OF_CHARS && sc.indexOf(peekChar()) == -1)
       s.append(readChar());
     readChar();
     return s.toString();
--- a/org.hl7.fhir.utilities/src/test/java/org/hl7/fhir/utilities/xhtml/XhtmlParserTests.java
+++ b/org.hl7.fhir.utilities/src/test/java/org/hl7/fhir/utilities/xhtml/XhtmlParserTests.java
@@ -32,4 +32,36 @@ class XhtmlParserTests {
   void testModeratelyNestedDivStillParses() throws FHIRFormatError, IOException {
     Assertions.assertNotNull(new XhtmlParser().parse(nestedDiv(100), "div"));
   }
+
+  private static final String DIV = "<div xmlns=\"http://www.w3.org/1999/xhtml\">";
+
+  // An entity reference that runs into end-of-input must fail with a FHIRFormatError. Both
+  // readUntil overloads used to test peekChar() != 0, but peekChar() returns END_OF_CHARS
+  // ((char) -1) at EOF and does not consume, so the loop never terminated and appended
+  // (char) -1 to the StringBuilder until the heap was exhausted. A few bytes were enough.
+  @ParameterizedTest
+  @ValueSource(strings = {
+      DIV + "&",
+      "<div xmlns=\"http://www.w3.org/1999/xhtml\" title=\"&",
+  })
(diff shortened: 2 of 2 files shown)


grahamegrieve (approved):
ok

PR #2583, opened 2026-09-01
Terminology: null-guard supplement merge in ValueSetValidator.resolveCodeSystem

### LINKED ISSUE #2540: NPE in ValueSetValidator.resolveCodeSystem when a ValueSet declares a supplement and includes an unresolvable code system
`ValueSetValidator.resolveCodeSystem` dereferences `cs` without a null check. When a ValueSet carries a `valueset-supplement` extension, `analyseValueSet` eagerly resolves every `compose.include`, so any include naming a code system that cannot be resolved locally throws.

The supplement itself is incidental - it only has to be present. In our case it supplements `http://terminology.hl7.org/CodeSystem/v3-RoleClass`, which resolves perfectly well; the include that fails is a different one.

The exception is caught by the broad `catch (Exception e)` in `BaseWorkerContext.validateCode` and turned into `BLOCKED_BY_OPTIONS`, which `ConceptMapValidator` then reports as `CONCEPTMAP_GROUP_TARGET_CODE_CANT_CHECK`. In our IG that produced 330 warnings of the form:

```
ConceptMap.group[0].element[0].target[0].code: The target code '2266.7' in the code system
https://terminology.dhp.uz/fhir/core/CodeSystem/position-and-profession-cs|0.7.0 cannot be
checked due to special rules around the code system
```

Neither the code nor the code system named in the message is involved. The crash happens during setup, before any code is examined, so every target code in the ConceptMap fails identically regardless of which code system its group targets.

## The code

`org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/validation/ValueSetValidator.java`, lines 604-615:

```java
if (!requiredSupplements.isEmpty()) {
  List<CodeSystem> additionalSupplements = new ArrayList<>();
  for (String s : requiredSupplements) {
    CodeSystem scs = context.fetchResource(CodeSystem.class, s, IWorkerContext.VersionResolutionRules.defaultRule());
    if (scs != null && cs.getUrl().equals(scs.getSupplements())) {   // <- cs may be null
```

`cs` can be null: lines 584-592 attempt `fetchSupplementedCodeSystem`, `findSpecialCodeSystem` and `fetchResource` in turn, and all three can return null.

The same loop appears in `ValueSetExpander.java` (lines 1441-1456), where it sits inside an `else` that is only reachable when `cs != null`:

```java
} else if (ValueSetUtilities.isServerSide(inc.getSystem()) || (cs == null || (cs.getContent() != COMPLETE && cs.getContent() != FRAGMENT))) {
  doServerIncludeCodes(...);
} else {
  ...
  if (scs != null && cs.getUrl().equals(scs.getSupplements())) {
```

So expansion is unaffected; only validation crashes. It looks like the validator's copy just lost the guard the expander's copy has.

## Stack trace

From an instrumented local build; line numbers adjusted back to the 6.10.0 tag.

```
java.lang.NullPointerException: Cannot invoke "org.hl7.fhir.r5.model.CodeSystem.getUrl()" because "cs" is null
  at ValueSetValidator.resolveCodeSystem(ValueSetValidator.java:608)
  at ValueSetValidator.checkCodeSystemResolves(ValueSetValidator.java:208)
  at ValueSetValidator.analyseValueSet(ValueSetValidator.java:175)
  at ValueSetValidator.<init>(ValueSetValidator.java:160)
  at BaseWorkerContext.constructValueSetCheckerSimple(BaseWorkerContext.java:1660)
  at BaseWorkerContext.validateCode(BaseWorkerContext.java:1464)
```

## To reproduce

A ValueSet declaring a supplement to one code system, and including another that the validator cannot resolve locally:

```json
{
  "resourceType": "ValueSet",
  "url": "http://example.org/ValueSet/demo",
  "extension": [{
    "url": "http://hl7.org/fhir/StructureDefinition/valueset-supplement",
    "valueCanonical": "http://example.org/CodeSystem/roleclass-supplement|1.0.0"
  }],
  "compose": { "include": [
    { "system": "http://terminology.hl7.org/CodeSystem/v3-RoleClass" },
    { "system": "http://snomed.info/sct", "concept": [{ "code": "158965000" }] }
  ]}
}
```

where `roleclass-supplement` has `content = supplement` and `supplements = http://terminology.hl7.org/CodeSystem/v3-RoleClass`. Point a ConceptMap's `targetScopeCanonical` at the ValueSet and validate.

SNOMED CT is just a convenient example of an include the validator will not resolve locally - LOINC, or any code system simply absent from the package, reaches the same line. Enumerating the concepts rather than including the system wholesale makes no difference either, since `checkCodeSystemResolves` looks only at `include.system`.

dylanpulver:
Fixes #2540.

`resolveCodeSystem` dereferences `cs` at `ValueSetValidator.java:607` while merging required supplements, but `cs` is null whenever none of the three lookups at `:583-591` resolve the system. With a `valueset-supplement` extension present, `analyseValueSet` (`:171-181`) eagerly calls `checkCodeSystemResolves` on every `compose.include`/`exclude`, so the constructor throws on any include naming a system the context cannot resolve locally.

`ValueSetExpander` has the same supplement loop at `:1457`, but there it sits in an `else` that `:1446` only reaches when `cs != null`. The validator's copy lost that guard; this restores it. The guard goes in the shared method rather than at the reporting call site, so all eight in-class callers of `resolveCodeSystem` are covered rather than the one path the issue names.

Behavior change is confined to one case: `cs == null` *and* a required supplement resolves. Every other case is unchanged — when no supplement resolves, `scs != null &&` already short-circuited before the dereference. Worth flagging that the changed case is a real verdict change downstream: today the NPE is swallowed by `BaseWorkerContext.validateCode`'s broad `catch (Exception e)` and returned as `BLOCKED_BY_OPTIONS`, which surfaces as `CONCEPTMAP_GROUP_TARGET_CODE_CANT_CHECK` warnings naming codes that were never actually examined. After this, those fall through to normal unknown-code-system handling, so some IGs will see "cannot be checked" warnings replaced by proper unknown-system messages.

Adds `ValueSetValidatorSupplementTest` — 3 tests, no network, built on `SimpleWorkerContextBuilder().fromNothing()`. There was no test package under `terminologies.validation` before this, and `BaseWorkerContextTests` supplies `ValueSetValidator` as a Mockito mock, so the real constructor never ran; nothing in the repo referenced `valueset-supplement` or `useSupplement`.

`mvn -pl org.hl7.fhir.r5 test`: 5197 tests on master, 5200 on the branch, 0 failures and 0 errors on both, checkstyle enabled. All three new tests fail on clean master with the reporter's exact trace. A mutant that catches the NPE at the caller instead still fails the test that calls `resolveCodeSystem` directly, which is what pins the fix to the shared method.

Not tested: the `org.hl7.fhir.validation` module suite, and no real IG reproduction — the tests use a synthetic minimal context. Built locally on JDK 26; CI covers 17/21/25.

---

Disclosure: the implementation, tests and this description were drafted with Claude Opus 5. The measurements above are reproducible from the branch.


grahamegrieve (approved):
ok

Issue #2511, opened 2026-07-08
Pattern discriminator on `Coding.userSelected` cannot be evaluated (`SLICING_CANNOT_BE_EVALUATED`)


jschoen-gefyra:
## Summary

When a slice uses a `pattern` discriminator whose pattern is a `Coding` that only
constrains `userSelected`, the validator fails to build a slicing expression and reports:

```
Slicing cannot be evaluated: Unsupported: no properties with values found on type
Coding for pattern for discriminator ($this) for slice <slice-id>
```

`userSelected` is a valid element of `Coding` and is addressable in FHIRPath, so a
pattern that only constrains `userSelected` should be a perfectly valid discriminator.
This is a bug.

## Environment

- Module: `org.hl7.fhir.validation`
- FHIR version of the example: R4 (`4.0.1`)

## Root cause

`InstanceValidator.buildCodingExpression(...)` builds a FHIRPath `where(...)` clause from a
pattern `Coding`, but only handles `system`, `version`, `code` and `display`. It never
inspects `userSelected`:

```java
// org.hl7.fhir.validation / InstanceValidator.buildCodingExpression(...)
boolean first = true;
expression.append(discriminator + ".where(");
if (c.hasSystem())  { first = false; expression.append("system = '"  + c.getSystem()  + "'"); }
if (c.hasVersion()) { ...            expression.append("version = '" + c.getVersion() + "'"); }
if (c.hasCode())    { ...            expression.append("code = '"    + c.getCode()    + "'"); }
if (c.hasDisplay()) { ...            expression.append("display = '" + c.getDisplay() + "'"); }
// <-- c.hasUserSelected() is never checked
if (first) {
  // pattern only constrained userSelected -> first is still true -> throws
  throw new DefinitionException(context.formatMessage(
      I18nConstants.UNSUPPORTED_IDENTIFIER_PATTERN_NO_PROPERTY_NOT_SUPPORTED_FOR_DISCRIMINATOR_FOR_SLICE,
      discriminator, ed.getId(), c.fhirType()));
}
expression.append(").exists()");
```

Because a `userSelected`-only pattern leaves `first == true`, the method throws
`DefinitionException` (message id
`UNSUPPORTED_IDENTIFIER_PATTERN_NO_PROPERTY_NOT_SUPPORTED_FOR_DISCRIMINATOR_FOR_SLICE`),
which surfaces to the user as `SLICING_CANNOT_BE_EVALUATED`.

[...]

PR #2599, opened 2026-09-02
Resolve extension(url) discriminators against inline sub-extensions

bvdh:
A profile cannot re-slice a complex extension when that extension's sub-extensions are defined inline. `FHIRPathEngine.evaluateDefinition()` throws `FHIRPATH_DISCRIMINATOR_CANT_FIND_EXTENSION` while resolving the discriminator, and snapshot generation and validation both fail.

This is the symptom reported in #2138 and #2511, and the fix #2279 attempted.

## Two distinct defects

**1. An absent extension is treated as fatal.** When the named extension is not present on a slice, the `extension(url)` branch throws. But a composite discriminator may legitimately designate an element that only some slices carry — R5 §5.1.0.13 says it then contributes an empty value for the others. The sibling `ofType()` branch a few lines up already models exactly this with `okToNotResolve`; this change makes the `extension()` branch consistent with it.

**2. Inline sub-extensions are unaddressable.** The branch matched sub-extensions only through `type.profile`. A complex extension that declares its children inline has no profile to match on, so `extension('function')` could never resolve. Inline children are now matched on the fixed (or pattern) value of their `url` element, which is the thing that identifies them.

Defect 2 is the more consequential one in practice: **every** extension generated into `hl7.fhir.uv.xver-r5.r4` declares its sub-extensions inline, so no R5-to-R4 backported complex extension could be re-sliced at all. That is how we hit this — re-slicing the backported `ImagingSelection.performer` in the HL7 Europe Imaging IG.

## Scope

20 lines of production change in one file, plus one new sub-extension-url helper. No behaviour changes for discriminators that already resolved.

## Test case

The validator test case asked for in #2279 now exists: **FHIR/fhir-test-cases#280**. It exercises both defects — a sibling slice with no `function` child (defect 1) and inline sub-extensions with no `type.profile` (defect 2). It fails on current `master` and passes with this change.

This PR also adds 6 unit tests driving `evaluateDefinition()` through its public API over purpose-built StructureDefinitions. Four fail without the fix; two cover behaviour that already worked, to guard it.

## Verification

- `org.hl7.fhir.r5`: 5203 tests, 0 failures (includes SnapShotGenerationTests, SqlOnFhirRunnerTests)
- `org.hl7.fhir.validation` `ValidationTests`: 988 tests, 0 failures
- End to end on the IG that surfaced this: 171 `Unable to resolve discriminator` errors before, 0 after

## Two things worth your steer

**r4/r4b parity.** The identical throw exists at `org.hl7.fhir.r4/.../FHIRPathEngine.java:6820` and `org.hl7.fhir.r4b/.../FHIRPathEngine.java:6507`. I've deliberately left them alone rather than guess at your policy for porting fixes across the versioned modules. Happy to mirror it if you'd like.

**Relationship to #2279.** That PR fixed a neighbouring problem — an xver fallback in `ProfileUtilities.findProfile`. The two are complementary rather than overlapping; this one does not subsume it.

Refs #2138, #2511. Also reported downstream as ahdis/matchbox#424.

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/fhirpath/FHIRPathEngine.java
@@ -7116,6 +7116,20 @@ private TimeType timeAdd(TimeType d, Quantity q, boolean negate, ExpressionNode 
   }
 
 
+  private String fixedExtensionUrl(StructureDefinition sd, ElementDefinition ed) throws DefinitionException {
+    for (ElementDefinition child : profileUtilities.getChildMap(sd, ed, false).getList()) {
+      if (child.getPath().endsWith(".url")) {
+        if (child.hasFixed() && child.getFixed().isPrimitive()) {
+          return child.getFixed().primitiveValue();
+        }
+        if (child.hasPattern() && child.getPattern().isPrimitive()) {
+          return child.getPattern().primitiveValue();
+        }
+      }
+    }
+    return null;
+  }
+
   /** given an element definition in a profile, what element contains the differentiating fixed 
    * for the element, given the differentiating expression. The expression is only allowed to 
    * use a subset of FHIRPath
@@ -7197,10 +7211,15 @@ private TimeType timeAdd(TimeType d, Quantity q, boolean negate, ExpressionNode 
               focus = new TypedElementDefinition(t);
               break;
             }
+            if (exsd == null && targetUrl.equals(fixedExtensionUrl(sd, t))) { // sub-extension defined inline, identified by its fixed url
+              focus = new TypedElementDefinition(t);
+              break;
+            }
           }
         }
         if (focus == null) { 
-          throw makeException(expr, I18nConstants.FHIRPATH_DISCRIMINATOR_CANT_FIND_EXTENSION, expr.toString(), targetUrl, element.getElement().getId(), sd.getUrl());
+          // the extension is absent on this slice, so the discriminator contributes an empty value (R5 5.1.0.13)
+          okToNotResolve = true;
         }
       } else if ("ofType".equals(expr.getName())) {
         if (!element.getElement().hasType()) {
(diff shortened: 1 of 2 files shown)


grahamegrieve:
Thanks — I've reproduced this end to end (patch-compiled your FHIRPathEngine over the
released validator and run the fixture from FHIR/fhir-test-cases#280): master gives 12
errors, this gives 0.

The second defect is real and well found. I checked `hl7.fhir.uv.xver-r5.r4`: 232 of the
complex extensions declare their sub-extensions inline with a fixed url and no
`type.profile`, so matching only through `type.profile` made every one of them
impossible to re-slice. That's worth fixing, and it means #2138 shouldn't have stayed
wontfix.

I also checked that the fix isn't merely silencing the error: with two performers both
coded AUT, `performer/author: max allowed = 1, but found 2` is correctly raised, and a
`performer/author` missing its mandatory `actor` sub-extension is still caught. So the
discrimination genuinely works afterwards.

Before I take it, though:

## The blanket `okToNotResolve = true` is wider than the case it cites

That throw is the only signal that a discriminator names an extension which exists
nowhere at all. With a typo'd url (`extension('funktion')`), the patched validator no
longer says "the extension 'funktion' wasn't found in the profile". What comes out
instead, against the *instance*:

    Element matches more than one slice - performer/author, performer/device
    The pattern [... code AUT ...] not found. Issues: [... Value is 'DEV' but is fixed
    to 'AUT' ...]

So the profile bug isn't silently swallowed, but it's now reported as though the data
were at fault, somewhere the profile author won't look.

R5 5.1.0.13 is about an element that some slices carry and others don't — which is
narrower than "couldn't resolve for any reason". Could you scope it to that: tolerate
non-resolution only when the named extension resolves for at least one sibling slice in
the same slicing group, and keep the error otherwise? That preserves the case you need
and keeps the diagnostic for a genuinely bad discriminator.

## R6

- `org.hl7.fhir.services/src/main/java/org/hl7/fhir/services/fhirpath/FHIRPathEngine.java`
  is the R6 copy of this file and carries the identical code (line ~7192). Changes here
  need to go into both (we are transitioning to R6)

## Reference

#2511 is a different bug — `buildCodingExpression` doesn't handle `Coding.userSelected`
— and nothing here touches it. #2138 is the right reference; I'll reopen it.

Nit: `fixedExtensionUrl` matches any child whose path ends `.url`. Comparing against
`t.getPath()+".url"` costs nothing and can't drift if `getChildMap` ever returns more
than immediate children.

Issue #2262, opened 2025-12-14
Performance degradation when validating multiple CodeableConcept.coding entries against terminology services


benny779:
We are observing significant performance degradation when validating a `CodeableConcept` that contains multiple coding entries, even when only one of them requires terminology validation (e.g., SNOMED CT).

Validation time increases linearly (or worse) as additional codings are added, including local code systems that should not require expensive terminology resolution.

This behavior is reproducible and appears to be related to repeated terminology validation per `coding` entry.

#### Resource example
```JSON
{
  "resourceType": "Condition",
  "id": "1234",
  "code": {
    "coding": [
      {
        "code": "100372200900",
        "system": "http://snomed.info/sct",
        "userSelected": false
      },
      {
        "code": "719.46",
        "display": "Knee pain",
        "system": "http://fhir.org.com/cs/diag-code"
      },
      {
        "code": "719.4601",
        "display": "Knee pain",
        "system": "http://fhir.org.com/cs/diag-extended-code"
      }
    ],
    "text": "Knee pain"
  },
  "subject": {
    "reference": "Patient/1234"
  }
}
```

#### Observed Validation Times
```JSON
[
  {
    "operation": "warm up with invalid sct code and 2 more local codes",
    "time": "1809 ms"
  },
  {
    "operation": "no change",
    "time": "1389 ms"
  },
  {
    "operation": "no change",
    "time": "1117 ms"
  },
  {
    "operation": "remove last coding (2 remain)",
    "time": "1460 ms"
  },
  {
    "operation": "no change",
    "time": "933 ms"
  },
  {
    "operation": "remove last coding (only invalid sct remain)",
    "time": "1282 ms"
  },
  {
    "operation": "no change",
    "time": "0 ms"
  },
  {
    "operation": "change to valid sct code (1003722009)",
    "time": "1381 ms"
  },
  {
    "operation": "no change",
    "time": "0 ms"
  },
  {
    "operation": "adding a local coding back",
    "time": "1419 ms"
  },
  {
    "operation": "no change",
    "time": "942 ms"
  }
]
```

### Key Observations
- Validation time jumps to ~1.5s when multiple codings are present.
- Removing all non-SNOMED codings drops validation time to ~0 ms.
- Adding a single local coding back immediately reintroduces the delay.
- Re-validating unchanged content still incurs the full cost.
- Results may not be cached or reused across codings.
- Local code systems may be processed with the same cost as SNOMED.
[...]

PR #2600, opened 2026-09-03
Fixes #2540. Fixes #2524. Partially addresses #2262

### LINKED ISSUE #2524: `TerminologyCache` persists transient tx failures (`SERVER_ERROR`, `NOSERVICE`) — a single timed-out call poisons the on-disk cache indefinitely
## Summary

`TerminologyCache#store` (`org.hl7.fhir.r5.terminologies.utilities`) persists any tx server
response that carries a `TerminologyServiceErrorClass` other than `CODESYSTEM_UNSUPPORTED`,
including the transient infrastructure classes `SERVER_ERROR` (any HTTP failure, e.g.
`java.net.SocketTimeoutException`) and `NOSERVICE` (unable to reach the tx server). Once
such a failure is written to the on-disk `-txCache` under the request's cache token, it is
served back for every subsequent identical request forever — across process restarts,
`validator_cli.jar` upgrades, even after the tx server has fully recovered. The only
recovery today is to delete the `-txCache` directory by hand.

## Environment

- Module: `org.hl7.fhir.r5.terminologies.utilities.TerminologyCache` (the same shape is
  present in the r4 and r4b copies)
- Reproduced on `validator_cli.jar` **6.9.11** and **6.9.12** (the switch to okhttp3 in
  #2519 did not change this behavior)

## Root cause

In `TerminologyCache.java`:

```java
public void store(CacheToken cacheToken, boolean persistent, NamedCache nc, CacheEntry e) {
    if (noCaching) {
      return;
    }

    if ( !cacheErrors &&
        ( e.v!= null
        && e.v.getErrorClass() == TerminologyServiceErrorClass.CODESYSTEM_UNSUPPORTED
        && !cacheToken.hasVersion)) {
      return;
    }

    CacheEntry previous = nc.map.put(cacheToken.key, e);
    if (persistent) {
      ...
      nc.dirty = true;
      ... save(nc, now);
    }
}
```

Only `CODESYSTEM_UNSUPPORTED` (and only conditionally) is excluded from the store. Every
other class — including `SERVER_ERROR` and `NOSERVICE` — falls through to `nc.map.put(...)`
and, when `persistent` is `true`, `save(...)` writes the entry to
`<txCache>/<system>.cache` on disk.

`TerminologyServiceErrorClass` already exposes exactly the semantic we want to exclude:

```java
// TerminologyServiceErrorClass.java
public boolean isInfrastructure() {
    return this == NOSERVICE || this == SERVER_ERROR || this == VALUESET_UNSUPPORTED;
}
```

but no caller in `TerminologyCache` consults it before storing.

## Reproduction (deterministic unit test)

The JUnit test below fits `org.hl7.fhir.r5/src/test/java/.../TerminologyCacheTests.java`
and reproduces the bug with no external dependencies. On the current `master` it fails
because `loinc.cache` is written and contains `"class" : "SERVER_ERROR"`. With the fix
proposed below, it passes.

```java
@Test
public void infrastructureErrorsMustNotBePersisted() throws IOException {
    Object lock = new Object();
    Path tempCacheDirectory = createTempCacheDirectory();

    TerminologyCache terminologyCache = new TerminologyCache(lock, tempCacheDirectory.toString());

    // Simulate what happens when a validate-code call fails with a transport-level
    // error: the tx client wraps the failure into a ValidationResult with
    // errorClass = SERVER_ERROR (or NOSERVICE) and asks the cache to persist it.
    ValidationResult tx = new ValidationResult(
        ValidationMessage.IssueSeverity.ERROR,
        "java.net.SocketTimeoutException: Read timed out",
        TerminologyServiceErrorClass.SERVER_ERROR,
        null);

    Coding coding = new Coding()
        .setSystem("http://loinc.org")
        .setCode("34823-5");
    ValueSet valueSet = new ValueSet();
    valueSet.setUrl("http://hl7.org/fhir/ValueSet/doc-typecodes");

    TerminologyCache.CacheToken token =
        terminologyCache.generateValidationToken(
            CacheTestUtils.validationOptions, coding, valueSet, new Parameters());

    terminologyCache.cacheValidation(token, tx, /* persistent */ true);
    terminologyCache.save();

    Path loincCache = tempCacheDirectory.resolve("loinc.cache");
    if (Files.exists(loincCache)) {
        String contents = Files.readString(loincCache);
        assertThat(contents)
            .as("infrastructure errors (SERVER_ERROR/NOSERVICE) must not be persisted "
              + "to the on-disk tx cache — they would poison it indefinitely")
            .doesNotContain("\"class\" : \"SERVER_ERROR\"")
            .doesNotContain("\"class\" : \"NOSERVICE\"");
    }

    deleteTempCacheDirectory(tempCacheDirectory);
}
```

## Reproduction at scale

Poisoned cache entries persist across `validator_cli.jar` restarts and even across
6.9.11 → 6.9.12: on a synthetic 178k-resource JP Core / JP-CLINS dataset (produced by
`clinosim`, not real patient data), six specific `Composition` resources returned
`java.net.SocketTimeoutException` on every run regardless of chunking, concurrency, or
JVM socket-timeout tuning. Wiping `<txCache>` and rerunning the same validations against
the same tx server produced zero timeouts. Full before/after logs and a self-contained
step-by-step reproduction are collected in a separate writeup for anyone who wants to
replay them:
<https://github.com/TomoOkuyama/fhir-jp-validator/blob/main/docs/hapi-txcache-poisoning.md>

## Proposed fix

Skip the store when the entry carries an infrastructure error class:

```java
public void store(CacheToken cacheToken, boolean persistent, NamedCache nc, CacheEntry e) {
    if (noCaching) {
      return;
    }

    // Never persist transient infrastructure errors (network / server-side).
    // Caching them poisons the cache indefinitely: once a request times out or the
    // server errors once, every subsequent validation of the same tuple replays
    // the cached failure even after the underlying condition has cleared.
    if (e.v != null && e.v.getErrorClass() != null && e.v.getErrorClass().isInfrastructure()) {
      return;
    }

    if ( !cacheErrors &&
        ( e.v!= null
        && e.v.getErrorClass() == TerminologyServiceErrorClass.CODESYSTEM_UNSUPPORTED
        && !cacheToken.hasVersion)) {
      return;
    }
    ...
}
```

Built locally against `master` (`mvn -pl org.hl7.fhir.validation.cli -am install
-DskipTests`) and verified:

- All 59 existing `TerminologyCacheTests` still pass.
- The new `infrastructureErrorsMustNotBePersisted` test fails without the guard and passes
  with it.
- Normal operation is unchanged — success entries and non-infrastructure error entries
  still cache and are re-served.

## Related concerns (out of scope, mentioned for context)

- The r4 and r4b copies of `TerminologyCache` have a similar `store` shape and would
  benefit from the same guard.
- Recovery of caches already poisoned by earlier versions still requires the user to
  clear `<txCache>` (or some background sweep). Happy to open a follow-up if useful.

Happy to submit this as a PR (fix + the test above) if the approach looks right.

### LINKED ISSUE #2540: NPE in ValueSetValidator.resolveCodeSystem when a ValueSet declares a supplement and includes an unresolvable code system

grahamegrieve:


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/expansion/ValueSetExpander.java
```diff
@@ -1027,6 +1038,14 @@ public ValueSetExpansionOutcome doExpand(ValueSet source, Parameters expParams)
     for (ParametersParameterComponent p : expParams.getParameter()) {
       processParameter(p.getName(), p.getValue());
     }
+    // ValueSet.compose.property names the properties to return "if the client doesn't ask for any
```
dotasek: issue: repeats the same logic as line 866? Can we wrap this in a dedicated method?

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/expansion/ValueSetExpander.java
```diff
@@ -1027,6 +1038,14 @@ public ValueSetExpansionOutcome doExpand(ValueSet source, Parameters expParams)
...
+    // particular properties", so it is only consulted when the request named none
+    if (source.hasCompose() && !source.getCompose().getProperty().isEmpty() && !expParams.hasParameter("property")) {
+      expParams = expParams.copy();
+      for (StringType t : source.getCompose().getProperty()) {
```
dotasek: nitpick: t could be renamed composeProperty.

org.hl7.fhir.standalone/src/main/java/org/hl7/fhir/standalone/terminology/expansion/ValueSetExpander.java
```diff
@@ -1020,6 +1031,14 @@ public ValueSetExpansionOutcome doExpand(ValueSet source, Parameters expParams)
     for (ParametersParameterComponent p : expParams.getParameterList()) {
       processParameter(p.getName(), p.getValue());
     }
+    // ValueSet.compose.property names the properties to return "if the client doesn't ask for any
```
dotasek: issue: repeated code. I think this is a copy-paste from the other ValueSetExpander as well.

dotasek (changes requested):

PR #2606, opened 2026-09-03
Add ProfileBasedManipulator for applying set/add/remove operations to a resource

costateixeira:
Adds a way to mutate an existing FHIR resource by applying a sequence of `set` / `add` / `remove` operations to the element model, with optional post-mutation validation.

This is the counterpart to `ProfileBasedFactory`, which synthesises a fresh resource from a profile. This one modifies a resource that already exists, sharing the same path/parts vocabulary. The use case is test-data authoring: generate a conformant baseline, then perturb one element to produce a negative case.

**Purely additive.** Three files, +717 lines, no deletions and no changes to existing behaviour — the only edit to an existing file is a new method, a new nested class, and one import in `ValidationEngine`.

## What's in it

| File | |
| --- | --- |
| `ProfileBasedManipulator.java` (r5, `testfactory`) | new — applies the operations to an `Element` tree |
| `ValidationEngine.java` | +66 — `manipulateResource()` + `ManipulationResult`, placed next to `convertFormat()` |
| `ProfileBasedManipulatorTests.java` | new — 15 tests |

## The contract

Operations are a JSON array; each entry is one operation:

```json
{ "op": "set",    "path": "AllergyIntolerance.recordedDate", "value": "2026-04-01" }
{ "op": "set",    "path": "AllergyIntolerance.code.coding",
                  "parts": [ {"name":"system","value":"..."}, {"name":"code","value":"..."} ] }
{ "op": "add",    "path": "AllergyIntolerance.note", "parts": [ {"name":"text","value":"..."} ] }
{ "op": "remove", "path": "AllergyIntolerance.clinicalStatus" }
```

**Paths** are dotted element paths with an optional `[i]` index. The leading resourceType segment is optional. Polymorphic elements must be spelled out concretely (`Observation.valueQuantity`, not `value[x]`).

**Semantics**, which are the main thing to review:

- `set` — navigate to, or create, the target and write it. On a list without an explicit index, the **first** entry is targeted. With an explicit `[i]`, that entry must already exist.
- `add` — append a new entry. Lists only; an explicit `[i]` on an `add` target is rejected.
- `remove` — with an explicit index, only that entry goes; without one, **all** entries with that name are removed. A missing path is a no-op.

`manipulateResource(bytes, inputFormat, profileUrl, operations, enforce, outputFormat)` parses, applies, reserialises, and — when `enforce` is true — validates the result against `profileUrl` (or base FHIR) and returns the `OperationOutcome` alongside the bytes. **Validation never throws**; callers inspect the outcome. With `enforce` false, no validation runs and the outcome is null.

The manipulator itself does no profile-aware enforcement — slicing, value sets and cardinality are the caller's business, via `enforce`.

## Suggested reading order

1. The class javadoc on `ProfileBasedManipulator` — the full contract, and the source of the summary above.
2. `applySet` / `applyAdd` / `applyRemove`, then `walkOrCreate` and `parsePath` beneath them.
3. `ValidationEngine.manipulateResource` — thin: parse, apply, compose, optionally validate.
4. The tests, if you want the semantics as executable examples.

## Testing

15 tests, passing in ~10s, covering each operation kind, the `enforce` on/off split, and rejection of bad input (unknown `op`, missing `path`, `value` on a non-primitive, `add` to a non-list, explicit index on `add`).

Fixtures are written inline rather than generated, so the tests exercise the manipulator alone and need no test-data download.

---

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/testfactory/ProfileBasedManipulator.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/testfactory/ProfileBasedManipulator.java
@@ -1,0 +1,271 @@
+package org.hl7.fhir.r5.testfactory;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.hl7.fhir.exceptions.FHIRException;
+import org.hl7.fhir.r5.elementmodel.Element;
+import org.hl7.fhir.utilities.json.model.JsonArray;
+import org.hl7.fhir.utilities.json.model.JsonElement;
+import org.hl7.fhir.utilities.json.model.JsonObject;
+
+/**
+ * Applies a sequence of {@code set} / {@code add} / {@code remove} operations to an
+ * existing Element-model FHIR resource. Companion to {@link ProfileBasedFactory},
+ * which synthesises a fresh resource from a profile; this class mutates one that
+ * already exists. Both share the same path / parts vocabulary.
+ *
+ * <h3>Operations payload</h3>
+ * Each entry is a JSON object:
+ * <pre>
+ *   { "op": "set",    "path": "AllergyIntolerance.recordedDate", "value": "2026-04-01" }
+ *   { "op": "set",    "path": "AllergyIntolerance.code.coding",
+ *                     "parts": [ {"name":"system","value":"..."}, {"name":"code","value":"..."} ] }
+ *   { "op": "add",    "path": "AllergyIntolerance.note",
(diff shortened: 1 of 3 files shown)


grahamegrieve:
why isn't this using one of the patch options defined in the FHIR specification?

costateixeira:
Closing in favour of #2647, which does the same job with the standard FHIRPath Patch `Parameters` document instead of a format of our own: the paths are FHIRPath, the five operations are the spec's, and a patch written for a test step also works against a real server's `PATCH`. Nothing from this branch is needed there.

PR #2607, opened 2026-09-03
Allow a literal value in test data mappings, and handle base64Binary as FHIR allows

costateixeira:
Reworked after review.

## What it does

A mapping entry, or a part inside one, can carry a literal `value` instead of an `expression`. And a base64Binary value supplied by a mapping or a data column is handled the way FHIR allows: whitespace is stripped, what remains is validated, and a value that is not base64 is rejected with a line in the generation log saying why, then replaced by the generated value. The generated fallback uses `Base64.getEncoder()` rather than the MIME encoder, so it no longer line-wraps.

## What changed since the review

- The same change is made in both copies of `ProfileBasedFactory`, r5 and services (R6); the two diffs are identical.
- Base64 uses `Utilities.decodeBase64`, which strips whitespace, and writes the stripped value. Rejections are logged. The database-sourced values are not checked.
- The validation-module test that started a `ValidationEngine` and called tx-dev is deleted. The cases are now mapping data in the instance-generation harness: FHIR/fhir-test-cases#286 adds them, and this PR extends `TestInstanceGenerationTester`'s file lists. This PR goes green once that one is released.
- Documentation: FHIR/ig-guidance#69.

## Decisions taken, with the alternative

| Case | Chosen | Why | Alternative |
| --- | --- | --- | --- |
| An entry has both `value` and `expression` | **Error** naming the entry | It is almost always a mistake; letting one silently win hides it. | Log a warning and let `value` win. |
| `value` is not a string | **Stringify**: a number or boolean gives its text, an object or array its JSON text | Authors are not expected to be technical; quoting rules should not matter. | Reject anything that is not a string. |
| `"value": ""` | **Suppresses the element**, same as an entry with no expression | It is the natural way to blank one element out and matches existing behaviour. | Treat as an error. |
| `base64Binary` from a mapping or column | Strip whitespace, validate, write the stripped value; on rejection log why and use a generated value | FHIR allows whitespace and line-wrapped content was common before R5; a silent drop is what the review objected to. | Reject on any whitespace (the previous behaviour). |

The database-sourced values are not checked: the test-data database is ours and its 63 base64Binary values are known valid.

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/testfactory/ProfileBasedFactory.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/testfactory/ProfileBasedFactory.java
@@ -234,6 +234,13 @@ public class ProfileBasedFactory {
           if (val == null && data != null) { 
             val = getPrimitiveValue(ls, b.fhirType(), path, pe.path(), pe.definition().getId(), pe.definition().getPath());
           }
+          if (val != null && "base64Binary".equals(b.fhirType())) {
+            try {
+              java.util.Base64.getDecoder().decode(val);
+            } catch (IllegalArgumentException e) {
+              val = null; // not valid Base64; fall through and generate a conformant value instead
+            }
+          }
           if (val == null && pe.valueSet() != null) {
             ValueSetExpansionContainsComponent cc = doExpansion(ls, pe.valueSet());
             if (cc != null) {
@@ -558,6 +565,11 @@ public class ProfileBasedFactory {
   private String getPrimitiveValue(LogSet ls, String fhirType, String... ids) {
     JsonObject entry = findMatchingEntry(ls, ids);
     if (entry != null) {
+      if (entry.has("value")) {
+        String val = entry.asString("value");
+        ls.others.add("literal value = '"+val+"'");
+        return val;
+      }
       JsonElement expression = entry.get("expression");
       if (expression == null || !expression.isJsonPrimitive() || Utilities.noString(expression.asString())) {
         ls.others.add("Found an entry for "+entry.asString("path")+" but it had no expression");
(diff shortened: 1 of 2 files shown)


grahamegrieve:
I don't see any matching PR against the documentation in https://github.com/FHIR/ig-guidance? And it also needs the same changes made against the new R6 code as well

grahamegrieve:
Thanks, this is useful and there are no security concerns. It isn't mergeable yet, though. 

While you're documenting it, please decide a few edge cases and write them down. At the moment value silently wins when expression is also present, which is more likely a mistake in the mapping and should at least produce a warning. A non-string value (an object or array) is silently ignored. And "value": "" removes the element, the same as an entry with no expression.

The Base64 check needs rework. Base64.getDecoder() rejects any whitespace, but FHIR says readers should ignore whitespace in base64Binary, and line-wrapped content was common before R5. So a valid line-wrapped value from a mapping or data column gets thrown away without any log entry and replaced with "Some Binary Value". Please strip whitespace, validate what's left, and write out the stripped value rather than discarding it. Core is getting a Utilities.decodeBase64() that does exactly that, so use it once it lands. Log a line when a value is rejected, so the user can see why their value disappeared. Also, the fallback should use getEncoder(), not getMimeEncoder(): the MIME encoder breaks lines every 76 characters, and R5+ content shouldn't contain whitespace at all. It only works today because the sample string is short. On the database: the test-data database comes from [www.fhir.org](http://www.fhir.org/), so it is ours. I checked it, and all 63 base64Binary values in it are valid with no whitespace, so it doesn't need the check.

Finally, the tests depend on the network. On first run they download the roughly 30MB test-data database, and they call tx-dev when expanding value sets. That will make them slow and flaky in CI. There's already a harness that reads its cases from the test-cases repo (TestInstanceGenerationTester, using rX/instance-generation/factories). Please add these three cases there as mapping data instead of a new Java class that starts up a ValidationEngine.

costateixeira:
Reworked along the lines of your review, and set back to draft because it depends on FHIR/fhir-test-cases#286 being released before it can go green here:

- same change in the services (R6) copy of `ProfileBasedFactory`;
- Base64 through `Utilities.decodeBase64`, stripped value written, rejections logged, `getEncoder()` fallback, database values left unchecked;
- the validation-module test is gone; the three cases are mapping data in the instance-generation harness (FHIR/fhir-test-cases#286), and this PR extends `TestInstanceGenerationTester`'s lists;
- documentation in FHIR/ig-guidance#69.

The edge cases are decided and written down in the PR body with the reason and the alternative for each; I went for an error when both `value` and `expression` are given, stringifying a non-string value, and keeping `""` as suppression. Happy to switch any of them.

PR #2608, opened 2026-09-03
Add a requiredOnly mode to test data generation

costateixeira:
Stacked on #2607 (this branch carries its commits until it merges). Reworked after review: documentation and the R6 copy, and the tests moved out of the validation module.

## What it does

`requiredOnly` restricts profile-based generation to the elements the profile makes mandatory at the resource's top level, plus any element that has a mapping entry or a fixed value. Optional elements deeper in the resource are handled as before, since that path already drops the ones that end up empty.

It is available in two places, spelled the same:

- as a parameter of the validator's `POST /testdata` (documented in `validator-server.md` and the OpenAPI files);
- as a factory option in `factories.json`, next to `bundle` and `mark-profile`, so an IG author running factories from the IG gets it too (documented in FHIR/ig-guidance#69).

## What changed since the review

- The same change in both copies of `ProfileBasedFactory` and `TestDataFactory`, r5 and services (R6).
- The control-file option is new: it is what makes the feature documentable in ig-guidance and testable in the instance-generation harness.
- Tests: one for each entry point. The factory option is a case in FHIR/fhir-test-cases#286 (`RequiredOnlyGenerator`, on the test patient profile; the expected output has only the mandatory elements, where the same profile without the option also gets `telecom` and `managingOrganization`), with the file added to `TestInstanceGenerationTester`'s lists. The endpoint parameter is a case in the existing HTTP suite, generating from `bp` and checking that the mandatory elements are there and the optional ones are not.

## Decisions taken, with the alternative

| Point | Chosen | Why | Alternative |
| --- | --- | --- | --- |
| Where pruning applies | **Top level of the resource only** | Deeper optional elements are already dropped when they end up with no content, so pruning there changes little and costs a second rule to explain. | Prune at every level. |
| Name of the control-file option | **`requiredOnly`**, same as the HTTP parameter | The two entry points should agree; the file already mixes styles (`bundle`, `mark-profile`). | `required-only`, matching `mark-profile`. |
| Elements that are mapped or fixed | **Kept even when optional** | A mapping entry is an explicit request for the element; a fixed value is the profile's. | Prune everything optional regardless. |

Goes green once FHIR/fhir-test-cases#286 is released and #2607 has merged.

### DIFF
--- a/documentation/validator-server.md
+++ b/documentation/validator-server.md
@@ -209,6 +209,8 @@ used.
 ```
 
 `bundle` is compared as the *string* `"true"`, so a JSON boolean `true` does not switch it on.
+`requiredOnly` works the same way, and restricts generation to the elements the profile makes
+mandatory at the resource's top level.
 
 ### POST /loadIG
 
--- a/documentation/validator-server.openapi.yaml
+++ b/documentation/validator-server.openapi.yaml
@@ -445,6 +445,14 @@ paths:
                     JSON boolean `true` does not switch it on.
                   enum: ['true', 'false']
                   default: 'false'
+                requiredOnly:
+                  type: string
+                  description: >-
+                    The string `true` restricts generation to the elements the profile makes
+                    mandatory at the resource's top level. Compared as a string, like `bundle`,
+                    so a JSON boolean `true` does not switch it on.
+                  enum: ['true', 'false']
+                  default: 'false'
                 format:
                   type: string
                   description: Output format for the generated resources.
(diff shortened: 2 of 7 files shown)


grahamegrieve:
Also this needs (a) documentation and (b) R6 synchronisation

costateixeira:
Reworked along the lines of your review, still draft because it is stacked on #2607 and shares its dependency on FHIR/fhir-test-cases#286:

- same change in the services (R6) copies of `ProfileBasedFactory` and `TestDataFactory`;
- `requiredOnly` is now also a factory option in `factories.json`, spelled like the HTTP parameter, which is what makes it documentable in ig-guidance (FHIR/ig-guidance#69) and testable in the instance-generation harness;
- tests for both entry points: a harness case in FHIR/fhir-test-cases#286 (`RequiredOnlyGenerator`), and a case in the existing HTTP suite for the endpoint parameter.

The body lists the decisions with the reason and the alternative for each.

PR #2609, opened 2026-09-03
Fix inverted binding strength in QuestionnaireBuilder answer constraints

costateixeira:
`constraintTypeForBinding` has its condition the wrong way round, so the answer constraint `QuestionnaireBuilder` derives from a binding is inverted for every bound element.

```java
- else if (binding.getStrength() != BindingStrength.REQUIRED)
+ else if (binding.getStrength() == BindingStrength.REQUIRED)
      return QuestionnaireAnswerConstraint.OPTIONSONLY;
  else
      return QuestionnaireAnswerConstraint.OPTIONSORTYPE;
```

`OPTIONSONLY` means answers must come from the listed options; `OPTIONSORTYPE` allows the options *or* any value of the element's type. So today:

- a **required** binding gets `optionsOrType` — permitting any value of the type, even though a required binding says the answer has to come from the value set
- an **extensible / preferred / example** binding gets `optionsOnly` — restricting answers to the listed options, which those strengths do not call for

Building a Questionnaire from `Observation` on current master shows it directly:

| Element | Binding strength | Before | After |
| --- | --- | --- | --- |
| `Observation.status` | required | `optionsOrType` | `optionsOnly` |
| `Observation.code` | example | `optionsOnly` | `optionsOrType` |
| `AllergyIntolerance.clinicalStatus` | required | `optionsOrType` | `optionsOnly` |
| `AllergyIntolerance.type` | preferred | `optionsOnly` | `optionsOrType` |
| `AllergyIntolerance.code` | example | `optionsOnly` | `optionsOrType` |

The R4 and R4B builders have this the right way round; the inversion came in with the R5 move to `answerConstraint`. The R6 copy in `org.hl7.fhir.services` has the same condition and gets the same fix.

## Tests

`QuestionnaireBuilder` never expands a value set, so the tests need only the shared worker context: no validation engine, no terminology server.

- `org.hl7.fhir.r5` — `QuestionnaireBuilderTests` on `TestingUtilities.getSharedWorkerContext()`: required (`Observation.status`, `AllergyIntolerance.clinicalStatus`), preferred (`AllergyIntolerance.type`) and example (`Observation.code`, `AllergyIntolerance.code`). For a CodeableConcept the constraint sits on the generated `coding` item, which is what the test reads.
- `org.hl7.fhir.standalone` — the same three cases against the R6 copy. Standalone skips its tests by default; run with `-DskipTests=false`.

All three fail against the previous condition, with the constraints swapped. The earlier test in the validation module, which started a full engine on R4 core against tx-dev, is removed.

## Note on impact

This changes the output of every Questionnaire generated from a profile with bindings, so it is a behaviour change rather than a pure defect fix — worth a look from anyone relying on the current constraints.

---

### DIFF
--- a/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/QuestionnaireBuilder.java
+++ b/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/QuestionnaireBuilder.java
@@ -879,10 +879,13 @@ public class QuestionnaireBuilder {
   private QuestionnaireAnswerConstraint constraintTypeForBinding(ElementDefinitionBindingComponent binding) {
     if (binding == null) 
       return null;
-    else if (binding.getStrength() != BindingStrength.REQUIRED) 
+    // A required binding means the answer must come from the value set, so only the
+    // listed options are allowed. Any weaker strength (extensible, preferred, example)
+    // also permits a value of the element's own type.
+    else if (binding.getStrength() == BindingStrength.REQUIRED)
       return QuestionnaireAnswerConstraint.OPTIONSONLY;
     else
-      return QuestionnaireAnswerConstraint.OPTIONSORTYPE; 
+      return QuestionnaireAnswerConstraint.OPTIONSORTYPE;
   }
 
   private void addCodingQuestions(QuestionnaireItemComponent group, ElementDefinition element, String path, List<QuestionnaireResponse.QuestionnaireResponseItemComponent> answerGroups) throws FHIRException {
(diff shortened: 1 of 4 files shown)


grahamegrieve:
Thanks, the fix is right. The R4 and R4B builders get this the right way round (a non-required binding gives open-choice, a required one gives choice), so the inversion looks like it came in when the R5 port moved to answerConstraint. It needs making in the R6 copy too: org.hl7.fhir.services/.../utilities/QuestionnaireBuilder.java has the same != BindingStrength.REQUIRED. On the null guard in isExempt: if getWorkingCode() can really return null, then the other getWorkingCode().equals(...) calls in the same class (isInlineDataType, the "*" checks, the Reference/Quantity branches) will fail the same way. Either guard them all, or tell us which input produced the null so we can see whether it's the StructureDefinition that's wrong.

Thanks for adding tests. This class has had no real ones: the QuestionnaireBuilderTester classes are dead main() harnesses with a hard-coded Windows path, and I'll remove them. But these tests are heavier than they need to be. QuestionnaireBuilder is in org.hl7.fhir.r5 and never expands a value set, yet the tests are in the validation module, start a full ValidationEngine on R4 core, and connect to tx-dev. That makes them slow, and flaky whenever the terminology server is. Please move them into the r5 module on TestingUtilities.getSharedWorkerContext() with no terminology server, and add the same test in standalone against the services copy so R6 is covered too.

costateixeira:
@grahamegrieve is this ok?

PR #2610, opened 2026-09-03
Add GET /questionnaire to the validator HTTP server

costateixeira:
Adds an endpoint that builds a FHIR Questionnaire from a StructureDefinition profile and serves it over the validator's HTTP API.

```
GET /questionnaire?profile=<canonical-url>[&select=...][&format=json|xml]
```

The Questionnaire is built from the profile's **snapshot**, generated on the fly when the profile does not carry one, and the ValueSets of coded elements are expanded into answer options. The profile must already be loaded, via `-ig` at start up or `/loadIG`.

**Purely additive** — a new handler, a new engine method, one route registration. No existing behaviour changes.

## Selecting a subset

`select` takes a JSON array of FHIRPath expressions, evaluated against every `ElementDefinition` of the snapshot. An element is selected when *any* expression is true. An item is kept when its path is selected, is an **ancestor** of a selection, or is a **descendant** of one — so the item tree stays connected rather than losing its parents.

```
select=["mustSupport = true"]        only mustSupport elements
select=["min > 0"]                   only required elements
select=["path = 'Patient.name'"]     one named element
```

MustSupport filtering is deliberately not a special case — it is just one expression among others.

```sh
curl -G 'http://localhost:8080/questionnaire' \
  --data-urlencode 'profile=http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient' \
  --data-urlencode 'select=["mustSupport = true"]'
```

`format` (`json` or `xml`, default `json`) picks the output format. `Accept` is not consulted, matching the other query-parameter endpoints rather than the body-posting ones.

## What's in it

| File | |
| --- | --- |
| `ValidationEngine.java` | +165 — `generateQuestionnaire` in two forms (whole profile, or pruned) plus the selection and pruning helpers |
| `QuestionnaireHTTPHandler.java` | new — parameter handling and error responses |
| `validator-http-openapi.json` | the served OpenAPI document |
| `validator-server.openapi.yaml` | its YAML rendering, kept in step |
| `validator-server.md` | endpoint table, conventions, and a section |
| `FhirValidatorHttpServiceTests.java` | +4 tests |

## Tests

Four, added to the existing `TestsThatCallAServer` class so they run against a live server:

| Test | Covers |
| --- | --- |
| `testQuestionnaireGenerateLegacy` | generation from a profile |
| `testQuestionnaireSelectFilter` | `select` prunes to matching elements |
| `testQuestionnaireMissingProfile` | missing `profile` → 400 with an OperationOutcome |
| `testQuestionnairePostNotAllowed` | POST → 405 |

---

### DIFF
--- a/documentation/validator-server.md
+++ b/documentation/validator-server.md
@@ -73,8 +73,9 @@ else - including a missing header - means JSON.
 
 **Output format** comes from `Accept`: a media type containing `xml` means XML, anything else
 means JSON. `/convert` and `/transform` use this to pick their output format. `/snapshot`,
-`/narrative` and `/version` return the same format they were given, and `/testdata` uses the
-`format` field in its body.
+`/narrative` and `/version` return the same format they were given, `/testdata` uses the
+`format` field in its body, and `/questionnaire` ignores `Accept` in favour of a `format`
+query parameter.
 
 **Errors** are FHIR `OperationOutcome` resources, served as `application/fhir+json` or
 `application/fhir+xml`. The one exception is a wrong HTTP method, which returns `405` with a
@@ -100,6 +101,7 @@ the HTTP status.
 | POST | `/convert` | Convert a resource between JSON and XML |
 | POST | `/version` | Convert a resource between FHIR versions |
 | POST | `/snapshot` | Generate the snapshot for a StructureDefinition |
+| GET | `/questionnaire` | Generate a Questionnaire from a profile |
 | POST | `/narrative` | Generate the narrative for a resource |
 | POST | `/transform` | Run a StructureMap over a resource |
 | GET | `/compile` | Fetch a StructureMap by canonical URL |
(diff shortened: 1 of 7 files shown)


costateixeira:
Rebased on master. Master has moved the validator onto the services line since this was opened, so the engine code is ported rather than merged - worth a look:

- `generateQuestionnaire` and the pruning helpers now use `org.hl7.fhir.services.utilities.QuestionnaireBuilder` and the R6 model (`org.hl7.fhir.model.core.Questionnaire`, `getItemList()`, `getElementList()`), and serialise with `XmlParser(context.getModelContext())` / `JsonParser(...)` as the rest of the engine does.
- `QuestionnaireHTTPHandler` takes `(IModelContext, service)` like the other handlers now do.

Behaviour is unchanged; the `/questionnaire` tests pass along with the rest of the HTTP suite (45/45).

PR #2611, opened 2026-09-03
Add GET /package, the CRMI $package operation

costateixeira:
Adds the CRMI `$package` operation to the validator's HTTP API: given a root canonical artifact, return a `collection` Bundle holding that artifact plus everything it transitively references.

```
GET /package?url=<canonical-url>[&expand=true][&includeRules=true][&format=json|xml]
```

Profiles, extensions, value sets, code systems, `Library`, `ActivityDefinition`, `PlanDefinition`, `ConceptMap`, `NamingSystem` and so on are pulled in. Core FHIR resources are not included, and the root artifact must already be loaded — via `-ig` at start up, or `/loadIG`.

| Parameter | Required | Default | |
| --- | --- | --- | --- |
| `url` | yes | — | canonical URL of the root artifact |
| `expand` | no | `false` | replace each ValueSet entry by its expansion (best effort — one that cannot be expanded is included unexpanded) |
| `includeRules` | no | `false` | also follow canonical URLs inside StructureMap rules |
| `format` | no | `json` | output format; `Accept` is not consulted |

**Not implemented**, from the full CRMI parameter surface: paging (`count`/`offset`), `contentEndpoint`/`terminologyEndpoint`, `packageOnly`, `manifest`, and capability-based filtering.

## The one change outside this feature — please look here

`ResourceDependencyWalker` learns to walk a `StructureMap`:

- the `StructureDefinition`s named by `structure[].url`
- other maps pulled in by `import`
- and, **only when opted in** via `setIncludeRuleReferences(true)`, canonical URLs appearing as parameters of rule-level transforms — notably the ConceptMap in `translate(<concept-map>, code, ...)`

Worth being explicit about the risk here, since the walker has an existing caller in `ResourceDependencyPackageBuilder`: **this cannot regress anything that currently works.** Before this change a `StructureMap` fell through the walker's final `else` and threw `Error("Resource StructureMap not Processed yet")`. Any dependency graph containing one was fatal. So the change converts a failure into working behaviour rather than altering a working path.

Rule-walking is off by default because it pulls in a long tail of extra artifacts that not every caller wants — `/package` exposes it as `includeRules`.

## Cost of a large graph

Worth flagging for anyone deploying this: a `/package` call walks the whole transitive dependency graph, and with `expand=true` each ValueSet entry also costs a terminology server round trip. `includeRules=true` compounds it, since rule-walking reaches ConceptMaps that the plain walk does not. A request against a large root artifact is therefore expensive in both time and outbound calls, and there is no paging (`count`/`offset` are among the unimplemented CRMI parameters).

This matches the server's existing posture rather than changing it — it binds to loopback unless `-allowNetworkAccess` is given, has no authentication, and already exposes `/loadIG` and `/txTest`, both of which make outbound calls on request. But it is the kind of endpoint worth keeping behind that default.

## Broken-link filtering

Not every unresolvable reference is a missing dependency. `urn:oid:`, `urn:uuid:`, `urn:ietf:`, `mailto:`, `tel:`, and SNOMED concept/edition IRIs are filtered out of the reported broken links rather than being surfaced as absent artifacts.

## What's in it

| File | |
| --- | --- |
| `ResourceDependencyWalker.java` | +51 — StructureMap support, opt-in rule references |
| `ValidationEngine.java` | +149 — `packageResource` in two forms, bundle serialisation for the engine's FHIR version, broken-link filtering |
| `PackageHTTPHandler.java` | new |
| `validator-http-openapi.json` / `.openapi.yaml` / `validator-server.md` | the served spec, its YAML rendering and the prose docs, kept in step |
| `FhirValidatorHttpServiceTests.java` | +3 tests, run against a live server |

---

### DIFF
--- a/documentation/validator-server.md
+++ b/documentation/validator-server.md
@@ -102,6 +102,7 @@ the HTTP status.
 | POST | `/snapshot` | Generate the snapshot for a StructureDefinition |
 | POST | `/narrative` | Generate the narrative for a resource |
 | POST | `/transform` | Run a StructureMap over a resource |
+| GET | `/package` | Package an artifact with its dependencies |
 | GET | `/compile` | Fetch a StructureMap by canonical URL |
 | GET | `/txTest` | Run one terminology ecosystem test against a server |
 | POST | `/stop` | Shut the server down |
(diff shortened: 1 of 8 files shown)


costateixeira:
Before I rebase this one: master has moved the validator onto the services line, and `packageResource` is built on `ResourceDependencyWalker`, which only exists in r5 (this PR also extends it with StructureMap walking). So bringing this up to date means a services-side copy of the walker, about 470 lines, before the endpoint itself can be ported.

Is the walker on your migration list? If you're about to move it, I'll wait and rebase on top of that rather than collide. If not, I can port it myself - either inside this PR or as a separate small PR first, whichever you'd rather review.

PR #2612, opened 2026-09-03
Add POST /fml, parsing FHIR Mapping Language into a StructureMap

costateixeira:
Parses posted FHIR Mapping Language source into a StructureMap and returns it — the counterpart of `/transform`, which runs a map rather than parsing one.

```
POST /fml?name=<name>&format=json|xml       body: the FML source text
```

Nothing is registered; the parsed map is returned to the caller. `name` is used only in parse error messages. `format` (`json` or `xml`, default `json`) picks the output format — `Accept` is not consulted, matching the other query-parameter endpoints rather than the body-posting ones.

**Purely additive** — a new handler, a new engine method, one route registration. No existing behaviour changes.

## The map is serialised for the engine's FHIR version, not R5

This is the part worth a look. A StructureMap is emitted in the version the engine is actually running, rather than always in R5.

It matters because of a specific shape difference: R5 carries a **typed `parameter` array** on `group.rule.dependent`, where R4 expects a **flat `variable` string list**. An R4-mode validator that emitted R5 JSON would silently drop the variable names — the resource parses without complaint, and the loss only surfaces later, at transform time, as the misleading

```
Rule 'X' has N but the invocation has 0 variables
```

from the R4 executor. Serialising for the runtime version avoids producing a map that is quietly wrong for the engine that will run it.

## What's in it

| File | |
| --- | --- |
| `ValidationEngine.java` | +52 — `parseStructureMap` and the version-aware serialisation behind it |
| `FmlHTTPHandler.java` | new |
| `validator-http-openapi.json` / `.openapi.yaml` / `validator-server.md` | the served spec, its YAML rendering and the prose docs, kept in step |
| `FhirValidatorHttpServiceTests.java` | +4 tests, run against a live server |

## Tests

| Test | Covers |
| --- | --- |
| `testFmlParseLegacy` | FML source parses to a StructureMap |
| `testFmlParseEmitsRuntimeVersion` | the map comes back in the engine's version, not R5 |
| `testFmlParseMissingBody` | empty body → 400 with an OperationOutcome |
| `testFmlGetNotAllowed` | GET → 405 |

---

### DIFF
--- a/documentation/validator-server.md
+++ b/documentation/validator-server.md
@@ -102,6 +102,7 @@ the HTTP status.
 | POST | `/snapshot` | Generate the snapshot for a StructureDefinition |
 | POST | `/narrative` | Generate the narrative for a resource |
 | POST | `/transform` | Run a StructureMap over a resource |
+| POST | `/fml` | Parse FHIR Mapping Language into a StructureMap |
 | GET | `/compile` | Fetch a StructureMap by canonical URL |
 | GET | `/txTest` | Run one terminology ecosystem test against a server |
 | POST | `/stop` | Shut the server down |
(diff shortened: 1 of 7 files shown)


costateixeira:
Rebased on master. Master has moved the validator onto the services line since this was opened, so the engine method is ported rather than merged - worth a look:

- `parseStructureMap` now parses with `org.hl7.fhir.services.fml.StructureMapTools` and serialises through the engine's existing `handleOutputToStream`, which already converts to the runtime version (R4, R4B, R3) via the `_N` converter factories. The hand-written `serialiseStructureMapForVersion` is gone.
- `FmlHTTPHandler` takes `(IModelContext, service)` like the other handlers now do, and uses the model-utilities `FhirFormat` / `OperationOutcomeUtilities`.

Behaviour is unchanged; the four `/fml` tests pass along with the rest of the HTTP suite (45/45).

PR #2613, opened 2026-09-03
Add POST /loadResource, registering a resource in the running engine (depends on #2612)

costateixeira:
> **Depends on #2612 — please merge that one first.**
>
> This branch is stacked on it, so the diff below currently shows both changes. To review only what this PR adds, look at the single commit:
> [`20296f7`](https://github.com/costateixeira/org.hl7.fhir.core/commit/20296f7774)
>
> Three of its tests drive `/fml` to produce their input, which is the workflow the two endpoints are for — hence the stack rather than two independent PRs. Left as a draft until #2612 merges.

---

Registers the posted resource — or every entry of a posted `collection`, `batch` or `transaction` Bundle — in the shared validation engine, so that it resolves by canonical URL exactly as if it had been loaded from a package.

```
POST /loadResource?format=json|xml&replace=true|false
```

This is how an artifact built at run time, typically a StructureMap parsed by `/fml`, is made available to `/transform`, `/compile` and validation without rebuilding an NPM package. Like `/loadIG`, the change is **permanent for the lifetime of the process** — there is no unload.

The response is plain JSON rather than an `OperationOutcome`:

```json
{ "loaded": 1, "resources": ["StructureMap/Example"] }
```

## `replace`, and why it isn't the default

Without `replace=true`, a resource whose canonical URL is already in the context is reported as **skipped** rather than overwritten, and the context is left alone.

The flag exists for the authoring loop: someone iterating on the same artifact resubmits it under the same canonical URL repeatedly, without bumping `version` on every edit. Silently replacing by default would make it too easy to overwrite a package-loaded artifact by accident, so the safe behaviour is the default and replacement is opt-in.

## The `(warning: ...)` descriptor

A returned descriptor may carry a trailing `(warning: ...)` note. The case that prompted it: loading an **R5-format StructureMap into a non-R5 engine**. The R4/R4B/R3 JSON parsers drop the R5-only `parameter` array on `group.rule.dependent` — R4 uses a flat `variable` list instead. The resource loads without any error, and the loss only surfaces much later, at transform time, as a misleading

```
Rule 'X' has N but the invocation has 0 variables
```

Rather than fail the load, the mismatch is detected and reported alongside the resource that was registered.

## Body format detection

`format` overrides it explicitly. Without it, the format comes from `Content-Type`; failing that, from the first non-whitespace byte (`<` means XML, anything else JSON).

## What's in it

| File | |
| --- | --- |
| `ValidationEngine.java` | +150 — `loadResourceFromBytes` in two forms, registration with the replace/skip semantics, and the version-loss detection |
| `LoadResourceHTTPHandler.java` | new |
| `validator-http-openapi.json` / `.openapi.yaml` / `validator-server.md` | the served spec, its YAML rendering and the prose docs, kept in step — including the note that `/loadIG` **and** `/loadResource` mutate the shared engine |
| `FhirValidatorHttpServiceTests.java` | +7 tests, run against a live server |

## Tests

| Test | Covers |
| --- | --- |
| `testLoadResourceEndToEnd` | register a resource, then resolve it |
| `testLoadResourceBundle` | every entry of a `collection` Bundle is registered |
| `testLoadResourceReplaceUpdatesExisting` | `replace=true` overwrites; the default skips |
| `testLoadResourceWarnsOnVersionMismatch` | the R5-into-R4 warning |
| `testLoadResourceRoundTripsParsedFml` | `/fml` output loads and is usable |
| `testLoadResourceMissingBody` | empty body → 400 |
| `testLoadResourceGetNotAllowed` | GET → 405 |

---

### DIFF
--- a/documentation/validator-server.md
+++ b/documentation/validator-server.md
@@ -49,8 +49,8 @@ The server runs until it is stopped with Ctrl-C or by a `POST /stop`.
 * A single `ValidationEngine` is created at start up and shared by every request. Per-request
   query parameters build a fresh `InstanceValidatorParameters` for that request; they do not
   change the engine.
-* `POST /loadIG` *does* change the engine, for every subsequent request, permanently - there is
-  no unload.
+* `POST /loadIG` and `POST /loadResource` *do* change the engine, for every subsequent request,
+  permanently - there is no unload.
 * The server is a `com.sun.net.httpserver.HttpServer` created with the default executor, so
   requests are handled sequentially on the dispatcher thread. It is designed for one developer,
   one build pipeline, or one test suite at a time - not as a shared multi-tenant service.
@@ -97,6 +97,7 @@ the HTTP status.
 | POST | `/matchetype` | Compare a resource against a matchetype pattern |
 | POST | `/testdata` | Generate test data for a profile |
 | POST | `/loadIG` | Load an IG into the running engine |
+| POST | `/loadResource` | Register a resource in the running engine |
 | POST | `/convert` | Convert a resource between JSON and XML |
 | POST | `/version` | Convert a resource between FHIR versions |
 | POST | `/snapshot` | Generate the snapshot for a StructureDefinition |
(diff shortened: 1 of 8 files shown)


costateixeira:
Rebased onto the updated #2612 (this branch still carries that commit until #2612 merges). As with #2612, master has moved the validator onto the services line, so the engine code is ported rather than merged - worth a look:

- `loadResourceFromBytes` and its helpers use the R6 model (`getEntryList()`), `context.getFHIRVersion()`, and the engine's existing `igLoader.loadResourceByVersion` / `seeResource` / `context.dropResource`, all of which already exist on master in services form.
- `LoadResourceHTTPHandler` takes `(IModelContext, service)` like the other handlers now do.

Behaviour is unchanged; the `/loadResource` tests pass along with the rest of the HTTP suite (52/52).

PR #2614, opened 2026-09-03
Test non-id dependent parameters and the R6 StructureMap converters

costateixeira:
Reworked after review: the separate test class duplicated `testBidirectionalStructureMapConversion` and `testR4ToR5StructureMapConversion`, so it is gone. In its place, the two gaps named in review, added to `Convertor_Factory_40_50Test` on a map parsed from FML:

## Non-id dependent parameters

A dependent call passing a string and an integer literal (`then item(item, patient, 'patient.sex', 1)`). R4 only has string variables, so each non-id parameter goes to R4 as an **empty** variable carrying the typed value in the original-variable-type extension, and comes back typed (`valueString`, `valueInteger`) on the way to R5. Checked in both directions; the id parameters on the same call are checked to stay plain.

## The R6 StructureMap converters, which had no tests

The same map through:

- `StructureMap50_N`: R5 → R6 → R5, typed parameters preserved both ways;
- `StructureMap40_N`: R4 (literals in extensions) → R6 typed → R4 with the extensions again.

Three tests, `mvn test -pl org.hl7.fhir.convertors -Dtest=Convertor_Factory_40_50Test`.

### DIFF
--- a/org.hl7.fhir.convertors/src/test/java/org/hl7/fhir/convertors/conv40_50/StructureMapDependent40_50Test.java
+++ b/org.hl7.fhir.convertors/src/test/java/org/hl7/fhir/convertors/conv40_50/StructureMapDependent40_50Test.java
@@ -1,0 +1,122 @@
+package org.hl7.fhir.convertors.conv40_50;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.hl7.fhir.convertors.conv40_50.resources40_50.StructureMap40_50;
+import org.hl7.fhir.convertors.factory.VersionConvertorFactory_40_50;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+/**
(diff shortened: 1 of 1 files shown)


grahamegrieve:
Thanks, but this repeats what we already have: Convertor_Factory_40_50Test.testBidirectionalStructureMapConversion and testR4ToR5StructureMapConversion check the dependent variable/parameter conversion in both directions, including the R5 → R4 → R5 round trip, on a map parsed from FML. So I'd rather not add a second class that checks the same thing. If you want to strengthen this area, the useful gaps are (a) dependent parameters that aren't ids (e.g. valueString or valueInteger), which go into R4 as an empty variable carrying an original-variable-type extension and have to come back typed, and (b) the R6 converters (StructureMap40_N / 50_N), which have no StructureMap tests at all. Adding those cases to the existing test class would be very welcome.

costateixeira:
Reworked as suggested: the separate class is gone, and the two gaps you named are added to Convertor_Factory_40_50Test - a dependent call with a string and an integer literal round-tripped through R4 (empty variable + original-variable-type extension, back typed), and the same map through StructureMap50_N and StructureMap40_N in both directions. 7 tests pass in the class.

PR #2616, opened 2026-09-03
Copied ValidationEngine shares the original's IgLoader, so loads through the copy land in the original

costateixeira:
## Problem

`ValidationEngine`'s copy constructor copies the worker context but shares the original's `IgLoader`. `IgLoader` captures the context it was constructed against, so `loadIg()` through a copy registers resources in the **original's** context, and the copy never sees them.

Anything that clones a base engine to isolate work is affected - `ValidationService.baseEngines` clones per request, and any server that wants a per-session context (the "clone the worker context first" model) gets no isolation for IG loads. The Lombok `setContext` had the same mismatch: it swapped the context and left the loader bound to the old one.

## Change

- Copy constructor: build the copy a loader bound to its own context (same construction the builder uses), keeping the direct provider.
- Explicit `setContext`: rebinds the loader to the new context.

No behaviour change for a single engine that is never copied.

## Tests

`ValidationEngineCopyTests` (3, all passing on R4 core):

- the copy's loader is bound to the copy's context, the original's to the original's;
- a ValueSet loaded through the copy is present in the copy and **absent** from the original - a shared loader cannot satisfy both assertions;
- after `setContext`, the loader is bound to the new context.

Two commits, one per change.

### DIFF
--- a/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ValidationEngine.java
+++ b/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/ValidationEngine.java
@@ -215,7 +215,7 @@ public class ValidationEngine implements IValidatorResourceFetcher, IValidationP
 
   }
 
-  @Getter @Setter private SimpleWorkerContext context;
+  @Getter private SimpleWorkerContext context;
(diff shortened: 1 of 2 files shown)


grahamegrieve:
This is a serious issue but the PR completely fails to deal with it. I'll take this question up myself.

PR #2617, opened 2026-09-03
HTTP server: /loadIG accepts a server-local path only when bound to loopback

costateixeira:
## Problem

`POST /loadIG` hands the caller's `ig` string straight to `IgLoader.loadIg`, which reads a server-local file or directory when one exists at that path. That is the intended use when the server is bound to loopback - the caller is a process on the same machine, and loading a local IG is the point. But with `-allowNetworkAccess`, any network caller can name a path on the host: the server reads any FHIR-shaped file it can reach, and `/loadIG`'s distinct error messages tell the caller whether a path exists at all.

## Change

When the server is **not** bound to loopback, `ig` must be a package reference (`id` or `id#version`, optionally with a `[version]` prefix) or an http(s) URL; anything else is refused with 400 and an explanation. Loopback behaviour is unchanged.

The classifier mirrors `IgLoader`'s own precedence rather than inventing a new one: archive names (`.tgz`, `.zip`, `.pack`) are never package references, and a package-shaped name that exists as a local file is a file - because that is what `IgLoader` would read. `FhirValidatorHttpService` gains an `isLoopbackOnly()` getter for the handler to consult.

Residual, stated plainly: a package-shaped name that exists in the working directory is refused with a different status than a package that doesn't exist, so *existence* of bare dotted names in the working directory is still inferable. *Contents* are not readable any more.

## Tests

`LoadIgSourcePolicyTests`: the classifier over package references, URLs, and every path shape I could think of (absolute, relative, traversal, UNC, `file:`, archive names); and two servers with a mocked engine - network-bound refuses a path with 400 *before* touching the engine, loopback-bound passes it through.

### DIFF
--- a/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/http/LoadIGHTTPHandler.java
+++ b/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/http/LoadIGHTTPHandler.java
@@ -1,13 +1,14 @@
 package org.hl7.fhir.validation.http;
 
+import org.hl7.fhir.exceptions.FHIRException;
 import org.hl7.fhir.utilities.filesystem.ManagedFileAccess;
 import org.hl7.fhir.utilities.npm.FilesystemPackageCacheManager;
 import com.sun.net.httpserver.HttpExchange;
 import com.sun.net.httpserver.HttpHandler;
 import lombok.extern.slf4j.Slf4j;
-import org.hl7.fhir.r5.model.OperationOutcome;
 import org.hl7.fhir.r5.utils.OperationOutcomeUtilities;
 import org.hl7.fhir.utilities.json.model.JsonObject;
+import org.hl7.fhir.validation.IgLoader;
 
 import java.io.IOException;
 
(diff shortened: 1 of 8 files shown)


org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/http/LoadIGHTTPHandler.java
```diff
@@ -49,4 +58,45 @@ public void handle(HttpExchange exchange) throws IOException {
...
+      return true;
+    }
+    // A dotted file name matches the package pattern too (package.tgz, secrets.json), and
+    // IgLoader resolves that ambiguity in favour of a file that exists. Mirror both rules here:
+    // archive names are never package references (IgLoader makes the same exclusion), and a
+    // package-shaped name that exists on this host's file system is a file, not a package.
+    if (lower.endsWith(".tgz") || lower.endsWith(".zip") || lower.endsWith(".pack")) {
+      return false;
+    }
+    if (!versionLessSrc.matches(FilesystemPackageCacheManager.PACKAGE_VERSION_REGEX_OPT)) {
```
dotasek: issue: This is failing on checkstyle because it does regex processing. It's using a known safe regex, though, so annotate this as below:

```java
    //anchored package name pattern, safe
    boolean matchesPackageVersionRegex = !versionLessSrc.matches(FilesystemPackageCacheManager.PACKAGE_VERSION_REGEX_OPT);
    if (matchesPackageVersionRegex) {
      return false;
    }
```

dotasek (changes requested):
Looks good, except for the regex usage which can be remedied with the code in the comments.

dotasek (approved):
Looks good to me

PR #2621, opened 2026-09-06
2026 09 gg validation enhancements

grahamegrieve:


org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/validation/ValidationOptions.java
```diff
@@ -62,6 +62,7 @@ public ValidationOptions(FhirPublication fhirVersion) {
 
   public ValidationOptions(FhirPublication fhirVersion, String language) {
     super();
+    this.fhirVersion = fhirVersion;
```
dotasek, line 65: nitpick: fhirVersion can be final now, and we should do so.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/TxTester.java
```diff
@@ -899,6 +926,74 @@ private String serverId() throws URISyntaxException {
     return new URI(server).getHost();
   }
 
+  @SuppressWarnings("checkstyle:patternUsage")
+  //a fixed literal reviewed here; never user input
+  private static final Pattern FOLDER_NAME = Pattern.compile("[A-Za-z0-9][A-Za-z0-9._-]*");
+  private static final Set<String> WINDOWS_DEVICE_NAMES = Set.of("con", "prn", "aux", "nul",
```
dotasek: issue: This looks like something that we should be including in ManagedFileAccess, not just in TxTester.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
```diff
@@ -654,6 +656,14 @@ public void initializeFromParameters(InstanceValidatorParameters parameters) {
     setDoImplicitFHIRPathStringConversion(parameters.isDoImplicitFHIRPathStringConversion());
     getSettings().setR5BundleRelativeReferencePolicy(parameters.getR5BundleRelativeReferencePolicy());
     setJurisdiction(CodeSystemUtilities.readCoding(parameters.getJurisdiction()));
+    getSettings().getUsageContexts().clear(); // should be empty - the settings are fresh - but this method is public
```
dotasek, line 659: thought: we only ever initializeFromParameters once in the InstanceValidator's lifecycle. Instead of doing this, maybe we should be throwing an exception if initializeFromParameters is called more than once. Or better yet, providing a factory or constructor that enforces this instead of allowing this method to be public.

dotasek (changes requested):
See comments.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/validation/ValidationOptions.java
grahamegrieve: ok

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
grahamegrieve: it's not a contractual requirement that we do it this way and we do things differently outside ValidationEngine

PR #2623, opened 2026-09-09
Keep package information and master definitions when copying a worker context

oliveregger:
## Problem
`BaseWorkerContext.copy()` does not copy the `packages` map, and `CanonicalResourceManager.copy()` does not copy `masterDefinitions` (nor `listForId` and `supplements`).

A copied context (via the `ValidationEngine` copy constructor, as used by `ValidationService` for sessions with a base engine, by `ValidationTests` with `CLONE = true`, and by downstream tools such as matchbox) therefore resolves an unversioned canonical differently from the original:
- `populatePVList` finds no dependencies for the package of the referencing resource
- `getByPackage` has no master definition to prefer

and falls back to the latest version. With `hl7.fhir.uv.xver-r5.r4` loaded in an R4 context, `http://hl7.org/fhir/encounter-status` resolves to 5.0.0 instead of 4.0.1 and the R4 code `finished` is rejected:

```
Unknown code 'finished' in the CodeSystem 'http://hl7.org/fhir/encounter-status' version '5.0.0'
```

Reported downstream in https://github.com/ahdis/matchbox/issues/538.

## Fix
- `BaseWorkerContext.copy()`: copy `packages`
- `CanonicalResourceManager.copy()`: also copy `masterDefinitions`, `listForId` and `supplements` (and clear all indexes first)

## Tests
- `CanonicalResourceManagerTests.testCopyKeepsMasterDefinitions`: fails on master (`expected: <4.0.1> but was: <5.0.0>`), passes with the fix
- New validator test case in FHIR/fhir-test-cases#283 (`xver-encounter-status-r5cs`, tag `context-copy`): fails on master with the error above, passes with the fix (run with `-DincludedValidationTags=context-copy`)

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,4 +4,4 @@
 
 ## Other code changes
 
-* no changes
+* Fix the copy of a worker context: keep the package information and the master definitions, so that unversioned canonicals resolve to the same version as in the original context (e.g. R4 core CodeSystem instead of the R5 one from hl7.fhir.uv.xver-r5.r4)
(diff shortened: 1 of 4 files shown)


grahamegrieve:
copying contexts is generally broken, and in ways that this PR doesn't address. I'm considering the issue more broadly, and I'll come back to this PR when I decide what to do. But it will probably not be accepted

grahamegrieve (approved):
ok

grahamegrieve (approved):
ok

PR #2624, opened 2026-09-09
License check via Maven plugin

dotasek:
This removes the flaky bash script used for verifying license compatibility and replaces it with org.codehaus.mojo:license-maven-plugin


grahamegrieve (approved):
ok

PR #2627, opened 2026-09-11
2026 09 gg various fixes

grahamegrieve:


org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/Utilities.java
```diff
@@ -1877,6 +1878,57 @@ public static String trimWS(String s) {
...
+      return Base64.getDecoder().decode(value);
+    }
+    byte[] src = new byte[len - ws];
+    int n = 0;
+    for (int i = 0; i < len; i++) {
+      char ch = value.charAt(i);
+      if (!isBase64Whitespace(ch)) {
+        // a non-ASCII character is never valid base64. Map it to a character that isn't either, rather
+        // than letting the cast truncate it into one that is (U+0141 would become 'A')
+        src[n++] = ch < 0x80 ? (byte) ch : (byte) '?';
```
dotasek, line 1918: issue: (del) is included in this, which isn't valid Base64. 

This refactor uses tighter ranges and still passes the tests:

```java
public static byte[] decodeBase64(String value) {
    int len = value.length();
    int ws = 0;
    for (int i = 0; i < len; i++) {
      if (isBase64Whitespace(value.charAt(i))) {
        ws++;
      }
    }
    if (ws == 0) {
      return Base64.getDecoder().decode(value);
    }
    byte[] src = new byte[len - ws];
    int n = 0;
    for (int i = 0; i < len; i++) {
      char ch = value.charAt(i);
      if (!isBase64Whitespace(ch)) {
        // a non-ASCII character is never valid base64. Map it to a character that isn't either, rather
        // than letting the cast truncate it into one that is (U+0141 would become 'A')
        src[n++] = ch == '=' // padding for base64
          || isBase64Char(ch)
            ? (byte) ch
            : (byte) '?';
      }
    }
    return Base64.getDecoder().decode(src);
  }

  private static boolean isBase64Char(char c) {
    return (c >= 'A' && c <= 'Z')
        || (c >= 'a' && c <= 'z')
        || (c >= '0' && c <= '9')
        || c == '+' || c == '/';
  }
```

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
```diff
@@ -3536,8 +3536,11 @@ else if (Utilities.isAllWhitespace(elementValue))
               ok = rule(errors, NO_RULE_DATE, IssueType.INVALID, e.line(), e.col(), path, false, I18nConstants.TYPE_SPECIFIC_CHECKS_DT_BASE64_VALID, value) && ok;
             } else {
               boolean wsok = !Base64Util.base64HasWhitespace(elementValue);
+              // R5+ says base64Binary content "does not include any whitespace or line feeds, but reading applications
+              // should ignore whitespace characters". That is not SHALL NOT, so it's a warning in all versions, but
+              // the language is stronger from R5 on
               if (VersionUtilities.isR5Plus(this.context.getVersion())) {
-                ok = rule(errors, NO_RULE_DATE, IssueType.INVALID, e.line(), e.col(), path, wsok, I18nConstants.TYPE_SPECIFIC_CHECKS_DT_BASE64_NO_WS_ERROR) && ok;
+                warning(errors, NO_RULE_DATE, IssueType.INVALID, e.line(), e.col(), path, wsok, I18nConstants.TYPE_SPECIFIC_CHECKS_DT_BASE64_NO_WS_ERROR);
```
dotasek, line 3543: issue: The constant representing this message is explicitly an ERROR. Is that OK, or should the message constant be changed to omit severity?

dotasek (changes requested):
See comments. Otherwise, this should have green tests before merge (see https://github.com/FHIR/fhir-test-cases/pull/284)

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java
grahamegrieve: Not an obvious thing - the message id has meaning to external users (in the advisor) so I chose not to change it. And it might become an error yet - I asked the committee to clarify

grahamegrieve:
this was overtaken by events, and the code merged elsewhere, and discussion will move to a new PR

PR #2630, opened 2026-09-14
2026 09 gg r6 validation

grahamegrieve:


org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/core/Identifier.java
```diff
@@ -747,5 +747,11 @@ public boolean isEmpty() {
       }
 
 
+  public boolean valueMatches(Identifier other) {
+    @SuppressWarnings("checkstyle:stringImplicitPatternUsage")
+    //system and value are identifiers/URLs; safe
+    boolean result = hasSystem() && hasValue() && getSystem().matches(other.getSystem()) && getValue().matches(other.getValue());
```
dotasek: issue: There is a check to parse the system as a URI, which keeps it from being a malicious regex, but no such check I can see for `value`. Can we just use a String.equals here for both instead of using a regex? It would be less costly as well.

org.hl7.fhir.services/src/main/java/org/hl7/fhir/services/client/FHIRToolingClient.java
dotasek, line 1: issue: I think this move to model from standalone is a bad idea. It's cleaner to keep this separate.

org.hl7.fhir.services/src/main/java/org/hl7/fhir/services/client/EFhirClientException.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  /**
+   * Method returns all OperationOutcome server errors that are associated with
+   * this exception.
+   * 
+   * @return
+   */
+  public List<OperationOutcome> getServerErrors() {
+    return errors;
```
dotasek, line 145: issue: We're returning full access to the internal list with this. Wrapping in List.of(), either here or at instantiation would ensure that callers can't modify this after getting it.

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/client/EFhirClientException.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  /**
+   * Method returns true if exception contains server OperationOutcome errors in
+   * payload.
+   * 
+   * @return
+   */
+  public boolean hasServerErrors() {
+    return errors.size() > 0;
```
dotasek: nitpick: could be !errors.isEmpty()

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/client/EFhirClientException.java
dotasek, line 1: issue: See comments in org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/client/EFhirClientException.java.

We shouldn't expose a mutable list unless it's really intended.

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/client/ResourceAddress.java
dotasek: issue: This is repeated many times in various ResourceAddress classes. Worth moving to `RegexConstants` and documenting there.

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/core/TypeConvertor.java
```diff
@@ -354,7 +358,13 @@ public static Coding castToCoding(Base b) throws FHIRException {
 
     if (b instanceof Coding)
       return (Coding) b;
-    else if (b instanceof ICoding) {
+    else if (b.isMetadataBased()) {
+      ICoding c = b.getAsICoding();
```
dotasek, line 362: question: If this is metadata based, this code assumes that it can also be an ICoding and return the ICoding if it exists? It seems weird that they can co-exist like this, when the case for ICoding directly follows at `b instanceof ICoding`

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/ModelContext.java
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Lazily initialised on first use of {@link #fullCoreContext()}; the class loader
+   * makes that thread safe and exactly-once
+   */
+  private static final class FullCore {
+    private static final ModelContext INSTANCE = build();
+
+    private static ModelContext build() {
+      ModelContext mc = new ModelContext();
```
dotasek, line 19: issue: None of the return values are being used here. This feels like it would have been a place to track which package was used for registration.

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/ModelContext.java
```diff
@@ -0,0 +1,69 @@
...
+      ToolsRegistration.register(mc, false);
```
dotasek, line 20: issue: Does the tools IG not register anything? It seems suspect that it's empty.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/IgLoader.java
```diff
@@ -882,16 +882,25 @@ else if (fn.endsWith(".json") && !fn.endsWith("template.json"))
         res = new org.hl7.fhir.dstu2.formats.JsonParser().parse(new ByteArrayInputStream(content));
       else
         throw new FHIRException("Unsupported format for " + fn);
-      r = VersionConvertorFactory_10_50.convertResource(res, new org.hl7.fhir.convertors.misc.IGR2ConvertorAdvisor5());
+      r = VersionConvertorFactory_10_N.convertResource(res, new org.hl7.fhir.convertors.misc.IGR2ConvertorAdvisor5());
     } else if (fhirVersion.startsWith("5.0")) {
+      org.hl7.fhir.r5.model.Resource res;
       if (fn.endsWith(".xml") && !fn.endsWith("template.xml"))
-        r = new XmlParser().parse(new ByteArrayInputStream(content));
+        res = new org.hl7.fhir.r5.formats.JsonParser().parse(new ByteArrayInputStream(content));
```
dotasek: issue: Using JsonParser for an xml file?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/IgLoader.java
```diff
@@ -882,16 +882,25 @@ else if (fn.endsWith(".json") && !fn.endsWith("template.json"))
         res = new org.hl7.fhir.dstu2.formats.JsonParser().parse(new ByteArrayInputStream(content));
       else
         throw new FHIRException("Unsupported format for " + fn);
-      r = VersionConvertorFactory_10_50.convertResource(res, new org.hl7.fhir.convertors.misc.IGR2ConvertorAdvisor5());
+      r = VersionConvertorFactory_10_N.convertResource(res, new org.hl7.fhir.convertors.misc.IGR2ConvertorAdvisor5());
     } else if (fhirVersion.startsWith("5.0")) {
+      org.hl7.fhir.r5.model.Resource res;
```
dotasek, line 887: issue: This used to have StructureMapUtilities in the r5 branch of this logic. The r6 branch uses StructureMapTools; maybe the r5 code got lost in the rewrite?

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/context/ContextResourceLoaderFactory.java
```diff
@@ -33,7 +37,29 @@ public static IContextResourceLoader makeLoader(String version, ILoaderKnowledge
     if (VersionUtilities.isR6Ver(version)) {
       return new R6ToR5Loader(Utilities.stringSet("CapabilityStatement", "StructureDefinition", "ValueSet", "CodeSystem", "SearchParameter", "OperationDefinition", "Questionnaire", "ConceptMap", "StructureMap", "NamingSystem"), loader);
     }
-    return new R5ToR5Loader(Utilities.stringSet("CapabilityStatement", "StructureDefinition", "ValueSet", "CodeSystem", "SearchParameter", "OperationDefinition", "Questionnaire", "ConceptMap", "StructureMap", "NamingSystem"), loader);
+    return new R6ToR5Loader(Utilities.stringSet("CapabilityStatement", "StructureDefinition", "ValueSet", "CodeSystem", "SearchParameter", "OperationDefinition", "Questionnaire", "ConceptMap", "StructureMap", "NamingSystem"), loader);
+  }
+
+  public static IContextResourceLoaderN makeLoaderN(IModelContext context, String version, ILoaderKnowledgeProviderRN loader) {
```
dotasek, line 43: question: Does this also need a case for R5ToRNLoader?

makeLoader always returns a loader for r5 resources, so it's not needed there, but I suspect it is needed in makeLoaderN

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/conv50_N/resources50_N/NutritionOrder50_N.java
```diff
@@ -353,8 +354,9 @@ public static org.hl7.fhir.r5.model.NutritionOrder.NutritionOrderEnteralFormulaC
       return null;
     org.hl7.fhir.r5.model.NutritionOrder.NutritionOrderEnteralFormulaComponent tgt = new org.hl7.fhir.r5.model.NutritionOrder.NutritionOrderEnteralFormulaComponent();
     ConversionContext50_N.INSTANCE.getVersionConvertor_50_N().copyBackboneElement(src, tgt);
-    if (src.hasType())
-      tgt.setBaseFormulaType(org.hl7.fhir.convertors.conv50_N.datatypes50_N.general50_N.CodeableReference50_N.convertCodeableReference(src.getType()));
+    for (org.hl7.fhir.model.core.CodeableReference t : src.getTypeList()) {
```
dotasek, line 357: issue: Is this correct? It looks like both these models have baseFormulaType. Shouldn't that be what get's converted?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/MeasureValidator.java
```diff
@@ -249,13 +245,14 @@ public boolean validateMeasureReport(ValidationContext hostContext, List<Validat
...
       long t = System.nanoTime();
-      Measure msrc = measure.startsWith("#") ? loadMeasure(element, measure.substring(1)) : context.fetchResource(Measure.class, measure, IWorkerContext.VersionResolutionRules.defaultRule());
+      Measure msrc = measure.startsWith("#") ? loadMeasure(element, measure.substring(1)) : context.fetchResource(Measure.class, measure, VersionResolutionRules.defaultRule());
       timeTracker.sd(t);
       if (warning(errors, NO_RULE_DATE, IssueType.REQUIRED, m.line(), m.col(), stack.getLiteralPath(), msrc != null, I18nConstants.MEASURE_MR_M_NOTFOUND, measure)) {
         boolean inComplete = !"complete".equals(element.getNamedChildValue("status", false));
         MeasureContext mc = new MeasureContext(msrc, element);
         NodeStack ns = stack.push(m, -1, m.getProperty().getDefinition(), m.getProperty().getDefinition());
-        hint(errors, NO_RULE_DATE, IssueType.BUSINESSRULE, m.line(), m.col(), ns.getLiteralPath(), mc.scoring() == null || Utilities.existsInList(mc.scoring(), "proportion", "ratio", "continuous-variable", "cohort", "composite", "attestation"), I18nConstants.MEASURE_MR_M_SCORING_UNK);
+        // todo disabled in R6?
```
dotasek, line 254: issue: Should we put an if statement here to check if this is r6? I worry the //todo will get lost in all these changes.

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/client/FHIRToolingClient.java
dotasek, line 157: question: I know there isn't an RN client for terminology (or otherwise), but maybe withVer should be updated pre-emptively.

dotasek (changes requested):
Initial review. This could use at least another pass while the comments are being addressed.

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/core/Identifier.java
grahamegrieve: probably we can. In fact, matches isn't more right that equals - it's something else needed, I think, but I'll wait until test cases are provided by real users

org.hl7.fhir.services/src/main/java/org/hl7/fhir/services/client/FHIRToolingClient.java
grahamegrieve: ok. I moved it to services - it must be there and not standalone

org.hl7.fhir.services/src/main/java/org/hl7/fhir/services/client/EFhirClientException.java
grahamegrieve: why would they bother? And if they bothered. why would it be a problem?

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/client/EFhirClientException.java
grahamegrieve: well, ok

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/utils/client/EFhirClientException.java
grahamegrieve: see response

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/client/ResourceAddress.java
grahamegrieve: done. and fixed a bug in it while I moved it.

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/ModelContext.java
grahamegrieve: Agree. I'll look at it at some stage in the next few days - a different PR

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/IgLoader.java
grahamegrieve: fixed

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/IgLoader.java
grahamegrieve: I split StructureMapUtilities, and renamed one half of it to StructureMapTools

grahamegrieve: ah but that wasn't the issue. good pick. Fixesd

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/context/ContextResourceLoaderFactory.java
grahamegrieve: good pick.

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/core/TypeConvertor.java
grahamegrieve: what's weird about this? it's been like this for a long time.

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/ModelContext.java
grahamegrieve: it is tracked. There's no need to do anything with the return value

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/conv50_N/resources50_N/NutritionOrder50_N.java
grahamegrieve: I think it's correct. baseFormulaType is renamed to type in R6

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/type/MeasureValidator.java
grahamegrieve: I'm talking to Bryn about what's intended here. but it's never been in any of the test cases

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/client/FHIRToolingClient.java
grahamegrieve: no, it shouldn't be. Stays at 5 until we really upgrade it

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/core/Reference.java
```diff
@@ -545,8 +545,7 @@ public boolean valueMatches(Reference value) {
         return false;
       }
       @SuppressWarnings("checkstyle:stringImplicitPatternUsage")
```
dotasek: issue: We can drop the suppression here if we're no longer using `matches`

org.hl7.fhir.services/src/main/java/org/hl7/fhir/services/terminology/ConceptMapEngine.java
```diff
@@ -0,0 +1,101 @@
...
+      return translateByJustCode(cm, source.getCode());
+  }
+
+  private Coding translateByJustCode(ConceptMap cm, String code) throws FHIRException {
+    SourceElementComponent ct = null;
+    ConceptMapGroupComponent cg = null;
+    for (ConceptMapGroupComponent g : cm.getGroupList()) {
+      for (SourceElementComponent e : g.getElementList()) {
+        if (code.equals(e.getCode())) {
+          if (e != null)
```
dotasek: issue: in the line above, `e.getCode()` will throw a null pointer if `e` is null. So this `e != null` is always true at this point.

It follows that `ct` will never get set, and the whole method will always return null.

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/ModelContext.java
dotasek: OK, I see the tracking in `packageList.add(packageName);` 

Then why return the package name at all if we don't use it?

Also, if each of these packages uses `register(IModelContext modelContext, boolean overridesBase)`, why not make it an interface they all implement?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
```diff
@@ -759,18 +761,18 @@ private List<NpmPackage> load() throws IOException {
         NpmPackage core = pcm.loadPackage(VersionUtilities.packageForVersion(v));
         NpmPackage tho = pcm.loadPackage("hl7.terminology");
         log.info("Load FHIR from "+core.name()+"#"+core.version());
-        SimpleWorkerContext ctxt = new SimpleWorkerContext.SimpleWorkerContextBuilder().withAllowLoadingDuplicates(true).fromPackage(core);
-        TerminologyClientFactory factory = new TerminologyClientFactory(ctxt.getVersion());
+        SimpleWorkerContext ctxt = new SimpleWorkerContext.SimpleWorkerContextBuilder(context.getModelContext()).withAllowLoadingDuplicates(true).fromPackage(core);
```
dotasek: issue: context is guaranteed to be null here, so will always throw an exception.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/codegen/JavaResourceGenerator.java
```diff
@@ -551,9 +543,9 @@ private void writeSearchParameterField(String name, JavaGenClass clss, boolean i
 
     Set<String> providesMembershipIn = new TreeSet<String>();
     for (CompartmentDefinition next : this.definitions.getCompartments().getList()) {
-      for (CompartmentDefinitionResourceComponent nextEntry : next.getResource()) {
+      for (CompartmentDefinitionResourceComponent nextEntry : next.getResourceList()) {
         if (nextEntry.getCode().equals(upFirst(name))) {
```
dotasek: issue: There's a type mismatch here, resulting in this always being false. The code that follows will never get entries in providesMembershipIn

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/codegen/JavaResourceGenerator.java
dotasek: issue: not in this PR, but this branch of the `if` statement is a duplicate of the previous. It won't be executed.

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/loaders/XVersionLoader.java
```diff
@@ -4,100 +4,110 @@
...
+        return VersionConvertorFactory_50_N.convertResource(new org.hl7.fhir.r5.formats.JsonParser().parse(stream));
+      case "6.0":
+        return new org.hl7.fhir.model.core.formats.JsonParser(context).parse(stream);
     }
     throw new FHIRException("Unknown version " + version + " loading resource");
   }
 
-  public static void saveXml(String version, Resource resource, OutputStream stream) throws FHIRFormatError, IOException {
+  public static void saveXml(IModelContext context, String version, Resource resource, OutputStream stream) throws FHIRFormatError, IOException {
     if (Utilities.noString(version)) {
```
dotasek, line 63: issue: Should there be a return here, like all the switch cases below?

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/loaders/loaderRN/R4ToRNLoader.java
```diff
@@ -0,0 +1,191 @@
...
+      b.setId(UUID.randomUUID().toString().toLowerCase());
+      b.setType(BundleType.COLLECTION);
+      b.addEntry().setResource(r5).setFullUrl(r5 instanceof CanonicalResource ? ((CanonicalResource) r5).getUrl() : null);
+    }
+    for (CodeSystem cs : advisor.getCslist()) {
+      BundleEntryComponent be = b.addEntry();
+      be.setFullUrl(cs.getUrl());
+      be.setResource(cs);
+    }
+    if (killPrimitives) {
```
dotasek: issue: This functionality is repeated multiple times across all R*toRNLoader. It should go in a reusable method in IContextResourceLoaderN or some other class.

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/loaders/loaderRN/R2016MayToRNLoader.java
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  @Override
+  public Bundle loadBundle(InputStream stream, boolean isJson) throws FHIRException, IOException {
+    Resource r2016may = null;
+    if (isJson)
+      r2016may = new JsonParser().parse(stream);
+    else
+      r2016may = new XmlParser().parse(stream);
+    org.hl7.fhir.model.core.Resource r5 = VersionConvertorFactory_14_N.convertResource(r2016may, advisor);
```
dotasek, line 68: issue: still uses r5 as a variable name instead of rN

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/loaders/loaderRN/R3ToRNLoader.java
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  @Override
+  public org.hl7.fhir.model.core.Resource loadResource(InputStream stream, boolean isJson) throws FHIRException, IOException {
+    Resource r3 = null;
+    if (isJson)
+      r3 = new JsonParser().parse(stream);
+    else
+      r3 = new XmlParser().parse(stream);
+    org.hl7.fhir.model.core.Resource r5 = VersionConvertorFactory_30_N.convertResource(r3);
```
dotasek: issue: The call in R2toRNLoader includes `advisor` in this method call. Is it missing here, or is there a deliberate reason for this difference?

A number of R*toRNLoader classes have this, which I've tried to note in file level comments.

This may also be present in all of the R*toR5Loader classes in the loaderR5 package, from which I assume it was copied.

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/loaders/loaderRN/R2ToRNLoader.java
dotasek, line 1: issue: still uses r5 variable names (see https://github.com/hapifhir/org.hl7.fhir.core/pull/2630/changes#r4018543795)

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/loaders/loaderRN/R4BToRNLoader.java
dotasek, line 1: issue: still uses r5 variable names (see https://github.com/hapifhir/org.hl7.fhir.core/pull/2630/changes#r4018543795)

May need to pass advisor: https://github.com/hapifhir/org.hl7.fhir.core/pull/2630/changes#r4018570185

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/loaders/loaderRN/R2016MayToRNLoader.java
dotasek, line 1: issue: 
May need to pass advisor: https://github.com/hapifhir/org.hl7.fhir.core/pull/2630/changes#r4018570185

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/loaders/loaderRN/R5ToRNLoader.java
dotasek, line 1: issue: May need to pass advisor: https://github.com/hapifhir/org.hl7.fhir.core/pull/2630/changes#r4018570185

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/misc/ProfileVersionAdaptor.java
```diff
@@ -44,48 +42,51 @@ public ConversionMessageStatus getStatus() {
...
   public StructureDefinition convert(StructureDefinition sd, List<ConversionMessage> log) throws FileNotFoundException, IOException {
-    if (sd.getKind() == StructureDefinitionKind.LOGICAL) {
+    if (sd.getKind() == StructureDefinition.StructureDefinitionKind.LOGICAL) {
       return convertLogical(sd, log);
     }
     if (sd.getDerivation() != TypeDerivationRule.CONSTRAINT || !"Extension".equals(sd.getType())) {
       return null; // nothing to say right now
     }
-    sd = sd.copy();
+    sd = sd.copy(Base.COPY_NOTHING);
```
dotasek: nitpick: I would change the param `sd` to `originalSd` and then assign everything below to `sd` like so:

`StructureDefinition sd = originalSd.copy(Base.COPY_NOTHING)`

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/factory/VersionConvertorFactory_10_N.java
dotasek, line 1: issue: There are a lot of BaseAdvisor_10_50 uses here. Does this not require a BaseAdvisor_10_N?

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/factory/VersionConvertorFactory_14_N.java
dotasek, line 1: issue: There are a lot of BaseAdvisor_14_50 uses here. Does this not require a BaseAdvisor_14_N?

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/factory/VersionConvertorFactory_30_N.java
dotasek, line 1: issue: There are a lot of BaseAdvisor_30_50 uses here. Does this not require a BaseAdvisor_30_N?

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/txClient/TerminologyClientNR3.java
```diff
@@ -0,0 +1,357 @@
...
+    return client.getActualVersion();
+  }
+  
+  @Override
+  public TerminologyCapabilities getTerminologyCapabilities() throws FHIRException {
+    // via the factory, not TerminologyCapabilities50_N directly: the generated converters read
+    // their VersionConvertor out of a ThreadLocal that only the factory sets up, so calling one
+    // directly NPEs in copyDomainResource
+    return (TerminologyCapabilities) VersionConvertorFactory_50_N.convertResource(
+      TerminologyCapabilities30_50.convertTerminologyCapabilities(client.getTerminologyCapabilities(), false));
```
dotasek: issue: the second param here is unused in the implementation. Maybe a good time to remove it?

org.hl7.fhir.services/src/main/java/org/hl7/fhir/services/context/CanonicalResourceProxy.java
dotasek, line 1: issue: this looks to be a direct copy of the internal CanonicalResourceProxy in a couple of CanonicalResourceManager classes. It would be easier to review in the future if all of these were lifted out of CanonicalResourceManager.

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/logging/Slf4JLoggingService.java
dotasek, line 1: issue: There appear to be 3 identical copies of this, and unless I'm missing something, none of them are bound to a particular FHIR version. Maybe this can live in `org.hl7.fhir.utilities`?

org.hl7.fhir.services/src/main/java/org/hl7/fhir/services/context/PackageResourceLoader.java
dotasek, line 1: issue: This originally lived as an internal class on SimpleWorkerContext. It would be far easier to compare across versions if this was lifted to its own class in r5.

dotasek (changes requested):
See comments.

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/core/Reference.java
grahamegrieve: ok

org.hl7.fhir.services/src/main/java/org/hl7/fhir/services/terminology/ConceptMapEngine.java
grahamegrieve: well, that doesn't follow, but there's a real bug there. Fixed

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/special/PackageReGenerator.java
grahamegrieve: how do you figure it's guaranteed to be null? It'll be set on line 187 of RePackageCommand, long before this code is called

grahamegrieve: oh no, I see. Sorted

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/codegen/JavaResourceGenerator.java
grahamegrieve: I"m not following this. What type mismatch?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/codegen/JavaResourceGenerator.java
grahamegrieve: I have no context for this.

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/loaders/XVersionLoader.java
grahamegrieve: added

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/loaders/loaderRN/R4ToRNLoader.java
grahamegrieve: ok. done

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/loaders/loaderRN/R2016MayToRNLoader.java
grahamegrieve: ok

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/loaders/loaderRN/R2ToRNLoader.java
grahamegrieve: fixed

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/loaders/loaderRN/R3ToRNLoader.java
grahamegrieve: fixed

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/loaders/loaderRN/R4BToRNLoader.java
grahamegrieve: fixed

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/loaders/loaderRN/R2016MayToRNLoader.java
grahamegrieve: fixed

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/loaders/loaderRN/R5ToRNLoader.java
grahamegrieve: not for R5

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/misc/ProfileVersionAdaptor.java
grahamegrieve: ok

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/factory/VersionConvertorFactory_10_N.java
grahamegrieve: no. because the conversion is a two step conversion, and it actually uses the advisor the 10_50 step.

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/factory/VersionConvertorFactory_14_N.java
grahamegrieve: same answer

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/factory/VersionConvertorFactory_30_N.java
grahamegrieve: same answer

org.hl7.fhir.convertors/src/main/java/org/hl7/fhir/convertors/txClient/TerminologyClientNR3.java
grahamegrieve: ok

org.hl7.fhir.services/src/main/java/org/hl7/fhir/services/context/CanonicalResourceProxy.java
grahamegrieve: ok

org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/logging/Slf4JLoggingService.java
grahamegrieve: ok

org.hl7.fhir.services/src/main/java/org/hl7/fhir/services/context/PackageResourceLoader.java
grahamegrieve: ok

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/codegen/JavaResourceGenerator.java
dotasek: `upFirst(name)` returns a String. `getCode()` returns ExtendedResourceTypes, which is an enum. Enums use reference equality in equals, so `nextEntry.getCode().equals(upFirst(name))` will always return false because  ExtendedResourceTypes.ANY_ENUM.equals("anyString") is always false.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/codegen/JavaResourceGenerator.java
dotasek: Sorry, for some or other reason, GitHub didn't include the line number.

The `if` statements at lines 263 and 275 are identical. The branch starting at 275 won't ever get executed because of this.

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/codegen/JavaResourceGenerator.java
grahamegrieve: oh. I looked at the wrong line of code when I looked at this - fixed

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/codegen/JavaResourceGenerator.java
grahamegrieve: not sure what it was trying to do. I'll remove both dead branches

org.hl7.fhir.standalone/src/main/java/org/hl7/fhir/standalone/context/SimpleWorkerContext.java
```diff
@@ -560,6 +579,7 @@ public int loadFromPackageInt(NpmPackage pi, IContextResourceLoader loader, Set<
             }
             t++;
           } catch (FHIRException e) {
+            e.printStackTrace();
```
dotasek, line 582: nitpick: leftover debug?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidatorFactory.java
dotasek, line 46: issue: Both this, and the method below set host services to `null`. The constructor specifies that it needs to be non-null. 

This is used in org.hl7.fhir.standalone.context.SimpleWorkerContext#newValidator, which is in turn called in our unit tests and FhirPathHostServices, so there's a good chance this will have an impact.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/igs/testing/TestPlan.java
```diff
@@ -1257,6 +1269,10 @@ public TestPlanParameterComponent copy() {
         return dst;
       }
 
+      public void assign(Base b) {
```
dotasek: question: is there a more descriptive name for `assign`? I see it copying values here and in the majority of places it's defined. If this is always the case, can we just call it `copyValuesFromBase`?

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/codegen/JavaParserXmlGenerator.java
```diff
@@ -269,7 +269,7 @@ private void genElement(Analysis analysis, TypeInfo ti, ElementDefinition ed, El
         boolean enShared = vs.hasUserData("shared") || vs.hasUserData("java.core.enum");
         String en;
         if (vs.hasUserData("java.core.enum")) {
-          en = (isR6() ? "org.hl7.fhir.model.core" : "org.hl7.fhir.r5.model")+".Enumerations."+ei.getName();
+          en = (isR6() ? "org.hl7.fhir.model.core" : "org.hl7.fhir.model.core")+".Enumerations."+ei.getName();
```
dotasek, line 272: issue: this piece of logic shows up many times. It would be worth a utility method.

dotasek (changes requested):
See comments.

org.hl7.fhir.standalone/src/main/java/org/hl7/fhir/standalone/context/SimpleWorkerContext.java
grahamegrieve: no I want that in production

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/igs/testing/TestPlan.java
grahamegrieve: that's confusing in several directions. I'll rename it to assignValues()

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/codegen/JavaParserXmlGenerator.java
grahamegrieve: next time I work on the generator

org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidatorFactory.java
grahamegrieve: this also has no line, but the test cases are passing

dotasek (approved):
Approved.

dotasek (approved):

Issue #2592, opened 2026-09-01
The "..." append convention for label has swapped operands in mergeStrings, corrupting the merged label


ewoutkramer:
Found during a systematic side-by-side review of the Java and .NET snapshot generators. Line references are pinned to commit b06c7ee (master, 2026-08-21) via permalinks; verified still present on master `4f52ba6` (2026-09-01).

`Utilities.appendDerivedTextToBase(baseText, derivedText)` returns `baseText + "\r\n" + derivedText.substring(3)` — the `"..."` append convention.

`ProfileUtilities.mergeMarkdown` (definition/comment/requirements) calls it correctly: `appendDerivedTextToBase(source.getValue() /* base */, mergedMarkdown.getValue() /* diff */)`.

`mergeStrings` (used for `label` only, [PU:3152-3168](https://github.com/hapifhir/org.hl7.fhir.core/blob/b06c7ee7c57367305ff72b6bb09ce779b5dbd6ce/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java#L3152-L3168)) has the operands swapped:

```java
} else if (res.hasValue() && source.hasValue() && res.getValue().startsWith("...")) {
  res.setValue(Utilities.appendDerivedTextToBase(res.getValue() /* DIFF */, source.getValue() /* BASE */));
```

For a differential label starting with `"..."` the result is `"...difflabel" + CRLF + baselabel.substring(3)` — the `"..."` marker survives into the snapshot and the *base* label loses its first three characters. Should be `appendDerivedTextToBase(source.getValue(), res.getValue())`, mirroring `mergeMarkdown`.

Issue #2462, opened 2026-05-29
Locale-sensitive test failures in I18nBaseTestandXMLUtilTests` break local Maven builds on non-English systems (de_DE JVM locale)


bassspieler:
## Summary

Running `mvn install` on a macOS machine whose JVM default locale resolves to `de_DE` causes at least two test classes in the `org.hl7.fhir.utilities` module to fail with assertion errors. The failures are locale-sensitive: the tests assert exact English-language strings, but the JVM or the underlying XML parser produces localised (German) output instead.

The parser/security behavior itself appears **correct** in both cases — the bug is in the test assertions, which are fragile against non-English runtime environments.

---

## Environment

| Property | Value |
|---|---|
| OS | macOS (Apple Silicon) |
| OS locale | `de_DE` |
| JVM default locale (`Locale.getDefault()`) | `de_DE` |
| JDK | Eclipse Temurin OpenJDK 25.0.3 (only JDK installed during reproduction) |
| Maven | Apache Maven 3.9.16 |
| Branch | `master` (fork of `hapifhir/org.hl7.fhir.core`) |
| Command | `mvn install --no-transfer-progress` |
| First failing module | `org.hl7.fhir.utilities` |

---

### 1. `org.hl7.fhir.utilities.i18n.I18nBaseTest`

Several assertions in this test class expect English error message strings, but the test runs under the JVM's default locale without explicitly setting a deterministic locale.

**Example failure:**

```
expected : "Error parsing JSON: test"
actual   : "Fehler beim Parsen von JSON: test"
```

The i18n infrastructure appears to be picking up the JVM default locale (`de_DE`) and returning the German translation, while the test asserts the English root-locale string. This suggests the test does not explicitly force a known locale before asserting message content.

---

### 2. `org.hl7.fhir.utilities.xml.XMLUtilTests`

This test verifies that the XML parser correctly rejects DOCTYPE declarations (a security-relevant feature). The parser **does** throw an exception as expected — the security behavior is correct — but the test asserts an exact English-language exception message that comes from the underlying Apache Xerces/JDK XML stack, which is localised on this machine.

**Example failure:**

```
Expected exception message to contain:
  "DOCTYPE is disallowed"

Actual exception message (localised by the JDK/XML parser):
  "DOCTYPE ist nicht zulässig, wenn das Feature
   \"http://apache.org/xml/features/disallow-doctype-decl\"
   auf \"true\" gesetzt ist."
```

[...]

PR #2631, opened 2026-09-14
Fix mergeAdditionalBinding writing "any" back to the source

### LINKED ISSUE #2590: mergeAdditionalBinding: the "any" flag merge is a self-assignment no-op

In `ProfileUtilities.mergeAdditionalBinding` ([PU:3219-3235](https://github.com/hapifhir/org.hl7.fhir.core/blob/b06c7ee7c57367305ff72b6bb09ce779b5dbd6ce/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java#L3219-L3235)):

```java
if (source.getAny()) {
  source.setAny(true);   // sets SOURCE, not dest — self-assignment no-op
}
```

When a differential's additional binding matches an inherited one (same valueSet + purpose) and carries `any = true`, the merged binding never receives it. Every other property in the method writes to `dest`. One-line fix: `dest.setAny(true)`.

AltangerelG:
Fixes #2590.

## The bug

`ProfileUtilities.mergeAdditionalBinding` merges a differential's additional
binding into the matching one inherited from the base. Every property is written
to `dest` except `any`, which is written back to `source`:

```java
if (source.getAny()) {
  source.setAny(true);   // self-assignment; dest never receives it
}
```

So when a differential's additional binding matches an inherited one on value set
and purpose, and sets `any = true`, the merged binding silently loses the flag.

One word: `source.setAny(true)` becomes `dest.setAny(true)`, as reported.

## Test

`ProfileUtilitiesAdditionalBindingTest` drives the real merge path through
`updateFromDefinition` rather than reaching for reflection, since
`mergeAdditionalBinding` is private and `updateFromDefinition` is protected.

- `mergedAdditionalBindingTakesAnyFromDifferential` fails on master and passes
  with this change.
- `mergedAdditionalBindingLeavesAnyUnsetWhenTheDifferentialDoesNotSetIt` passes
  either way. It is there so the first test cannot be satisfied by setting `any`
  unconditionally.

Verified by reverting the one-word change with the tests in place: 1 failure,
`expected: <true> but was: <false>`. Restoring it: both pass.

The first assertion in each test is on `getAdditional().size()`. Without it, an
empty list would surface as `IndexOutOfBoundsException` rather than as a clear
statement that the bindings never merged — which is exactly the failure I hit
while writing this, see below.

## Module suite

`mvn test -pl org.hl7.fhir.r5` reports 5211 tests with 8 failures and 3 errors,
all of them in `TerminologyCacheTests`. They are not from this change: running
`TerminologyCacheTests` alone on the unmodified parent commit gives the same 8
failures and 3 errors. They look environment-dependent (cache files under the
Windows temp folder), in the same family as #2462.

## One thing worth knowing about the test

The base element is given a bindable type:

```java
ed.addType().setCode("code");
```

That is not decoration. `updateFromDefinition` ends with

```java
// now, check that we still have a bindable type; if not, delete the binding - see task 8477
if (dest.hasBinding() && !hasBindableType(dest)) {
  dest.setBinding(null);
}
```

so an element with no type has its binding dropped after the merge, and the test
would assert against a binding that no longer exists. The type is on the base
element only: giving one to the differential as well would pull
`checkTypeDerivation` into the test, which needs a populated worker context.

`BindingStrength.EXTENSIBLE` is likewise deliberate. The `REQUIRED` branches call
`context.findTxResource(...)`, so `EXTENSIBLE` keeps the test free of a worker
context while still exercising the merge.

## Related, not changed here

`org.hl7.fhir.services` carries the same defect verbatim, in its own copy of
`ProfileUtilities.mergeAdditionalBinding`. I have left it alone: that module's
readme describes it as experimental R6 work, and how it tracks the r5 sources is
your call rather than mine. Happy to include it here, or to open a separate
issue, whichever you prefer.

The same applies to `mergeStrings` (#2592), which is also duplicated across both
modules.


AltangerelG:
@grahamegrieve asked for the R6 code to be updated too on #2632 and #2633, and
the same applies here, so I've done it rather than wait to be asked — the
`org.hl7.fhir.services` copy of `mergeAdditionalBinding` had the identical
`source.setAny(true)`.

`org.hl7.fhir.services` has no `src/test`, so that half is the source change
alone; the two tests in `org.hl7.fhir.r5` cover the behaviour, and the first
fails without the r5 half.

Happy to drop that commit if you'd rather the R6 module moved separately.

AltangerelG:
Added the R6 counterpart of the additional-binding merge test in `org.hl7.fhir.standalone`, following your

```
  Tests run: 2, Failures: 0
```

and with the R6 half of the fix reverted, the "any" case fails. The module sets

of using reflection.

PR #2632, opened 2026-09-14
Fix swapped operands in mergeStrings, corrupting the merged label

### LINKED ISSUE #2592: The "..." append convention for label has swapped operands in mergeStrings, corrupting the merged label

AltangerelG:
Fixes #2592.

## The bug

`Utilities.appendDerivedTextToBase(baseText, derivedText)` returns
`baseText + "\r\n" + derivedText.substring(3)`, so the base value goes first and
the `"..."` is stripped off the derived value.

`mergeMarkdown` calls it that way round. `mergeStrings`, which serves
`ElementDefinition.label`, does not:

```java
// mergeMarkdown (correct)
mergedMarkdown.setValue(Utilities.appendDerivedTextToBase(source.getValue(), mergedMarkdown.getValue()));

// mergeStrings (operands swapped)
res.setValue(Utilities.appendDerivedTextToBase(res.getValue(), source.getValue()));
```

Both are called as `merge*(derived, base)` — see the `base.setLabelElement(...)`
and `base.setDefinitionElement(...)` call sites — so in `mergeStrings` `res` is
the differential and `source` is the base. Passing them in that order produces
`"...difflabel" + CRLF + baselabel.substring(3)`.

The change swaps them back, matching `mergeMarkdown`.

## What it produced

With a base label of `Base label` and a differential label of `...and more`, the
merged label on master is:

```
expected: <Base label\r\nand more>
but was:  <...and more\r\ne label>
```

Both symptoms in one value: the `"..."` marker survives into the snapshot, and
the base label loses its first three characters.

## Test

`ProfileUtilitiesLabelMergeTest` drives the real merge through
`updateFromDefinition`, which is protected, rather than reaching for reflection
into the private `mergeStrings`.

- `labelStartingWithEllipsisIsAppendedToTheBaseLabel` fails on master with the
  value above, and passes with this change.
- `labelNotStartingWithEllipsisReplacesTheBaseLabel` passes either way, so the
  first test cannot be satisfied by changing how non-`"..."` labels are handled.

Verified by reverting the one-line change with the tests in place: 1 failure,
1 pass. Restoring it: both pass.

## Note

This is a sibling of #2631 (for #2590). Both are one-line fixes in
`ProfileUtilities` from @ewoutkramer's snapshot-generator review, but they touch
different methods and different lines, and each is independent — either can be
merged first. The test helper is duplicated across the two test classes
deliberately, to keep the pull requests independent of each other; happy to fold
them together if you would rather have one.

`mergeStrings` is also duplicated verbatim in `org.hl7.fhir.services`, with the
same swapped operands. I have not touched it, for the same reason as in #2631:
that module describes itself as experimental R6 work and how it tracks the r5
sources is your call.


grahamegrieve:
These changes also need to be made to the R6 code as well

AltangerelG:
Done — af66bfe applies the same operand swap to
`org.hl7.fhir.services/.../ProfileUtilities.java`, which had the identical
`appendDerivedTextToBase(res.getValue(), source.getValue())`.

`org.hl7.fhir.services` has no `src/test`, so that half is the source change
alone; the test in `org.hl7.fhir.r5` covers the behaviour and fails without the
r5 half.

AltangerelG:
Added the R6 counterpart of the label merge test in `org.hl7.fhir.standalone`, following your

```
  Tests run: 2, Failures: 0
```

and with the R6 half of the fix reverted, the "..." case fails. The module sets

of using reflection.

grahamegrieve (commented):
thanks.

grahamegrieve (approved):
ok

PR #2633, opened 2026-09-14
Fix "modiferExtension" typo in isExtensionSlicing

### LINKED ISSUE #2588: Typo "modiferExtension" in isExtensionSlicing: modifierExtension slicings never qualify as standard extension slicing

`SnapshotGenerationPreProcessor.isExtensionSlicing` checks

```java
Utilities.existsInList(ed.getName(), "extension", "modiferExtension")
```

[sic] ([PRE:1078](https://github.com/hapifhir/org.hl7.fhir.core/blob/b06c7ee7c57367305ff72b6bb09ce779b5dbd6ce/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/SnapshotGenerationPreProcessor.java#L1078)). Because of the `"modiferExtension"` typo, a slicing on `modifierExtension` never qualifies as standard extension slicing, so the preprocessor treats it as a real SliceInfo: its trailing content gets propagation treatment, and its presence inside another slicing's trailing region can trigger the `UNSUPPORTED_SLICING_COMPLEXITY` total bail-out (PRE:713-723) — so the typo is behavioral, not just cosmetic.

Second, separable question in the same predicate (design, maybe intentional): the clause `(!ed.getSlicing().hasOrdered() || ed.getSlicing().getOrdered())` ([PRE:1081](https://github.com/hapifhir/org.hl7.fhir.core/blob/b06c7ee7c57367305ff72b6bb09ce779b5dbd6ce/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/SnapshotGenerationPreProcessor.java#L1081)) means a slicing that *omits* `ordered` — the common case in the wild — is also not exempted; only an explicit `ordered=false` passes.

AltangerelG:
Fixes #2588.

## The bug

`SnapshotGenerationPreProcessor.isExtensionSlicing` exempts standard extension
slicing from being treated as a real slice, by matching the element name:

```java
Utilities.existsInList(ed.getName(), "extension", "modiferExtension")
```

The second name is misspelled, and `ed.getName()` returns the last path segment,
so a slicing on `modifierExtension` never matched. As #2588 notes, that is
behavioural rather than cosmetic: the slicing became a real `SliceInfo`, its
trailing content received propagation treatment, and its presence inside another
slicing's trailing region could trigger the `UNSUPPORTED_SLICING_COMPLEXITY`
bail-out, which abandons slicing processing for the whole profile.

## Test

`isExtensionSlicing` is private, and its three callers act on accumulated slice
state rather than returning anything, so observing the classification through
`process()` would need a differential big enough to reach the bail-out. The test
invokes the predicate directly instead, which is what the typo actually breaks.
Reflection is used sparingly elsewhere in the repo (`TurtleGeneratorTests`,
`CapabilityStatementUtilitiesTests`), so it seemed the lesser evil here — happy
to write the larger fixture instead if you would prefer the coverage.

Three cases: `Patient.extension` and `Patient.modifierExtension` are recognised,
`Patient.identifier` is not. Reverting the one-word change with the tests in
place fails exactly one of them, `modifierExtensionSlicingIsRecognised`, leaving
the other two passing.

## Not changed: the `ordered` clause

#2588 raises a second, separable point about the same predicate:

```java
(!ed.getSlicing().hasOrdered() || ed.getSlicing().getOrdered())
```

which means a slicing that omits `ordered` — the common case in the wild — is
also not exempted, and only an explicit `ordered=false` passes. The issue marks
that as possibly intentional, so I have left it alone; this pull request is the
typo only. The test sets `ordered` explicitly for that reason, and says so in a
comment, so it is not silently depending on the current behaviour.

Worth noting that if the `ordered` clause is later relaxed, this typo fix is
what makes the `modifierExtension` half of the predicate reachable at all.

## Related

Third of @ewoutkramer's snapshot-generator review issues, after #2631 (#2590)
and #2632 (#2592). All three are independent — different methods, and this one a
different file — so they can be merged in any order.


grahamegrieve:
this fix needs to be applied to the R6 code as well

AltangerelG:
Done — 9b949c0 applies the same fix to
`org.hl7.fhir.services/.../SnapshotGenerationPreProcessor.java`, which had the
identical `"modiferExtension"` spelling.

One note: `org.hl7.fhir.services` has no `src/test`, so that half is the source
change on its own. The three tests in `org.hl7.fhir.r5` still pin the predicate's
behaviour, and they fail without the r5 half of the change.

grahamegrieve:
All the R6 test cases are in standalone - because mostly they require infrastructure from that package to execute the tests

AltangerelG:
Thank you — I had that wrong. I looked in `org.hl7.fhir.services/src/test`, found
nothing, and wrote "no test source directory" in all four of these PRs, when what
I had actually established was only that I hadn't found them.

4738e2b adds the R6 counterpart in `org.hl7.fhir.standalone`, following the
convention there of dropping the `services.` segment from the test package
(`org.hl7.fhir.conformance.profile` against
`org.hl7.fhir.services.conformance.profile`).

Verified locally with the invocation in the standalone pom's comment:

```
mvn test -pl org.hl7.fhir.standalone -DskipTests=false -Dtest=SnapshotGenerationPreProcessorExtensionSlicingTest
  Tests run: 3, Failures: 0
```

and with the R6 half of the fix reverted:

```
  modifierExtensionSlicingIsRecognised:29 expected: <true> but was: <false>
  Tests run: 3, Failures: 1
```

so it does pin the R6 behaviour, even though `skipTests` means it compiles rather
than runs in the pipeline.

I'll add the equivalents to #2631, #2632 and #2649. Those three drive
`updateFromDefinition`, which is protected, so from a standalone test package
they'll need reflection rather than same-package access — tell me if you'd rather
they lived somewhere that avoids that.

grahamegrieve (approved):
ok. thanks

PR #2635, opened 2026-09-15
Render inline answerOption choices in Questionnaire form view

### LINKED ISSUE #2634: Questionnaire form rendering: inline answerOption choices are never rendered; dropdowns show only "??"
## Summary

When a Questionnaire item defines its choices inline with `answerOption`, the dropdown rendered in the **Sample Rendering** tab contains no actual options — only a single entry showing `??`.

## Reproduce

Image

HL7's own published SDC IG is a ready example. Note that the page opens on the **Tree view** tab, which renders correctly — you need to click the **Sample Rendering** tab to see the problem:

https://build.fhir.org/ig/HL7/sdc/en/Questionnaire-CardiologyForm.html

All 55 dropdowns on that tab are identical:

```html
<select size="1" name="patient_gender"><option value="a" selected="false">??</option></select>
<select size="1" name="additionalinfo_sexassignedatbirth"><option value="a" selected="false">??</option></select>
```

That Questionnaire has **72 items using inline `answerOption` and 0 using `answerValueSet`**.

Note the same page also shows `Answers: 3 options` next to the Gender item — the options are present in the resource, they are simply never rendered into the `<select>`.

## Root cause

I added temporary diagnostic output to `QuestionnaireRenderer.listOptions()` and ran a real build. Every item printed:

```
listOptions linkId=feeding hasAVS=true hasAO=true avsValue=null
```

`ResourceWrapper.has("answerValueSet")` returns `true` even when the element carries no value (`primitiveValue()` is `null`). The flow is therefore:

1. enters `if (i.has("answerValueSet"))`
2. value is null, so the value set cannot be resolved/expanded
3. that branch does **not** `return`
4. `else if (i.has("answerOption"))` is skipped entirely, and execution falls through to `select.option("a", "??", false)`

`renderItemOptions()` is never invoked (diagnostic count was 0), so no option is ever produced.

## Impact

Specification content itself is unaffected — the Tree view, the answer option list and the raw resource all render correctly. Only the form preview is affected, where a reader may reasonably conclude the Questionnaire defines no answer options.

## Environment

IG Publisher 2.3.4. The same logic is still present in the current upstream source.

---

A PR with a proposed fix follows.

hard25670559:
Fixes #2634

## Problem

`listOptions()` entered the `answerValueSet` branch whenever that element was present, even when it carried no value. Since that branch does not `return` when the value set cannot be resolved, execution skipped the `answerOption` branch entirely and fell through to `select.option("a", "??", false)`.

As a result, `renderItemOptions()` was never invoked for items using inline `answerOption`, and every dropdown in the Sample Rendering tab contained nothing but `??`.

## Change

```java
// before
if (i.has("answerValueSet")) {
  ...
} else if (i.has("answerOption")) {
  renderItemOptions(select, i);
}
select.option("a", "??", false);

// after
if (i.has("answerValueSet") && !Utilities.noString(i.primitiveValue("answerValueSet"))) {
  ...
} else if (i.has("answerOption")) {
  renderItemOptions(select, i);
  return;
}
select.option("a", "??", false);
```

1. only take the `answerValueSet` branch when it actually has a value
2. `return` after rendering `answerOption`, so the placeholder is not appended on top of real options

The `??` placeholder is left in place for the case where neither is present — happy to remove it instead if you prefer.

## Verification

Built an IG using inline `answerOption` with IG Publisher 2.3.4, overriding `org.hl7.fhir.r5` on the classpath with this change:

| | before | after |
|---|---|---|
| `<option>` elements | 10 | 30 |
| showing `??` | 10 | 0 |

The option count matches the 2–4 answers defined per item, with `value` set to the correct `system#code` and the text taken from `display`.


grahamegrieve:
As a fix, this makes sense, but as a PR, I have two issues: 

* the change needs to be made to R6 as well (we're migrating to R6)
* there needs to be a test case for this in the narrative generation test cases for R5 and R6 in the test cases repository

PR #2637, opened 2026-09-15
Send required supplements when validating a server-side include

### LINKED ISSUE #2636: Required supplement not found when the supplemented code system is only on the terminology server
IG Publisher 2.3.4 (core 6.10.4) rejects valid codes from a ValueSet that requires a supplement to a code system the validator doesn't have locally:

```
Required supplement not found: https://terminology.dhp.uz/fhir/core/CodeSystem/iso3166-two-letter-cs|0.8.0
None of the codings provided are in the value set 'ISO 3166 part 1 2-letter codes' (https://terminology.dhp.uz/fhir/core/ValueSet/iso-3166-2-vs|0.8.0), and a coding from this value set is required) (codes = urn:iso:std:iso:3166#UZ)
```

`iso-3166-2-vs` (uz.dhp.core#0.8.0) includes `urn:iso:std:iso:3166` and requires the `iso3166-two-letter-cs` supplement. THO 7 dropped the `content: not-present` ISO 3166 stub, so that code system is now only on tx.fhir.org. `ValueSetValidator.inComponent` checks the include on the server as a transient `iso-3166-2-vs--0` without the supplement, the server says the code is in, and `checkRequiredSupplements` then fails because a supplement only counts as used once it's merged into a local code system. Before #2540 this was hidden by an NPE.

tx.fhir.org accepts the code against the original ValueSet when the supplement is sent as a tx-resource.

To reproduce, build https://github.com/vadi2/VadimTestIG/tree/server-side-supplement-repro (one profile bound to `iso-3166-2-vs`, one example with `urn:iso:std:iso:3166#UZ`). 2.3.4 reports the error above, and a publisher built with #2637 reports 0 errors. The standalone validator doesn't show it, because it still loads hl7.terminology.r5#6.2.0, which has the stub.

vadi2:
Fixes #2636

A ValueSet that requires a supplement to a code system we have no full copy of (e.g. `urn:iso:std:iso:3166`, which THO 7 no longer carries a stub for) failed every code with "Required supplement not found". There was nothing local to merge the supplement into, so it was never counted as used, and the value set the server checked codes against didn't require it.

- `ValueSetValidator`: when an include is checked on the server, the transient ValueSet now requires the supplements that apply to that code system. When there is a server to check them, they count as used up front, as supplements merged into local code systems do - otherwise codes from another, local, code system in the same value set, and membership-only checks, still failed
- A supplement to a particular version (`supplements` = `url|version`) only applies to that version, the same rule as `CanonicalResourceManager.getSupplements(url, version)`. tx.fhir.org doesn't check this (it applies a supplement for `urn:iso:std:iso:3166|1999` to 2018 without complaint), so the client has to
- `BaseWorkerContext`: a value set's own required supplements now go to the server as tx-resources. Through its code systems, a supplement to a particular version was only found when the copy of the code system we resolve is that version
- The same changes are applied to the copies in `org.hl7.fhir.standalone`

A required supplement that can't be found is still reported, as is one to a different version of the code system.

Tested with an IG Publisher build against the first commit of this branch. The IG that failed on THO 7.3.0 then built with 0 errors, and an unknown code (`UU`) was still rejected, without the misleading supplement warning.


org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/terminologies/validation/ValueSetValidator.java
```diff
@@ -651,6 +651,23 @@ public CodeSystem resolveCodeSystem(String system, String version, Element ctxt,
...
 
+  private List<String> supplementsFor(String system) {
+    List<String> res = new ArrayList<>();
+    for (String s : requiredSupplements) {
+      CodeSystem scs = context.fetchResource(CodeSystem.class, s, IWorkerContext.VersionResolutionRules.defaultRule());
+      if (scs != null && scs.hasSupplements()) {
+        String base = scs.getSupplements();
+        if (base.contains("|")) {
+          base = base.substring(0, base.indexOf("|"));
+        }
```
grahamegrieve: Why strip the person here? I think that this is wrong

vadi2: @grahamegrieve addressed

grahamegrieve (approved):
ok

PR #2641, opened 2026-09-17
Type split() as an ordered collection in FHIRPath static analysis

### LINKED ISSUE #2640: False `FHIRPATH_NOT_A_COLLECTION` QA warning for `split(...).first()`
## Problem

QA reports a warning for a valid invariant:

```
Found a use of a collection operator on something that is not a collection
at 'split('-').first().lower()'
```

`split(separator)` returns an ordered collection of strings
([spec](http://hl7.org/fhirpath/#splitseparator-string-collection)), so `first()` on it is legal.
Both the Java and the JS engine in FHIRPath Lab evaluate the expression fine.

## Reproduction

A `StructureDefinition.constraint.expression` containing:

```fhirpath
entry.resource.language.all(
  $this.split('-').first().lower() = %resource.language.split('-').first().lower()
)
```

Note: the check only runs on snapshot elements, so a differential-only profile does not trigger it.

## Versions

- IG Publisher 2.3.4 / 2.3.5-SNAPSHOT, `core_version` 6.10.4
- Also present on `org.hl7.fhir.core` master (6.10.5-SNAPSHOT, `e9cb40e7b4`)

## Cause

Not in the publisher. `FHIRPathEngine.evaluateFunctionType(...)` in `org.hl7.fhir.core` types the
*return* of `split()` as a singleton:

```java
case Split:
  checkParamTypes(...);
  return new TypeDetails(CollectionStatus.SINGLETON, TypeDetails.FP_String);  // wrong
```

The same method warns whenever `first()`/`last()`/`tail()`/`skip()`/`take()` is applied to a
`SINGLETON` focus. `funcSplit()` returns all segments at runtime — only the static type is wrong.
`toChars()`, with identical semantics, is already typed `ORDERED`.

## Fix

Return `CollectionStatus.ORDERED`. The `checkParamTypes(...)` call stays `SINGLETON` — the
separator parameter really is a single string. Static analysis only; runtime unchanged.

`FHIRPathEngine.java` is copied per version, and the wrong line is in each copy that has `split()`:

| | `split()` | wrong typing | warning emitted |
| --- | --- | --- | --- |
| `r4`, `r5`, `services` (R6) | yes | yes | yes |
| `r4b` | yes | yes | no — the engine has no collection-status warnings at all |
| `dstu2`, `dstu2016may`, `dstu3` | no | — | — |

`r4b` has no `typeWarnings`, no `IssueMessage` and no `checkOnTypes(...)`, so nothing there turns
the wrong type into a warning — but its public `check(...)` still hands callers
`CollectionStatus.SINGLETON` for `split()`.

patrick-werner:
Fixes #2640.

`evaluateFunctionType()` types the return of `Split` as `SINGLETON`, and the same method warns when `first()`/`last()`/`tail()`/`skip()`/`take()` is applied to a `SINGLETON` focus — so `split(x).first()` raised a spurious `FHIRPATH_NOT_A_COLLECTION` warning that lands in IG QA reports. Per [spec](http://hl7.org/fhirpath/#splitseparator-string-collection) `split(separator)` returns an ordered collection of strings, and `funcSplit()` already returns every segment. Only the static type was wrong. `toChars()`, same semantics, was already `ORDERED`.

One line per engine: return `CollectionStatus.ORDERED`. `checkParamTypes(...)` stays `SINGLETON` — the separator parameter really is a single string. Static analysis only; runtime unchanged.

Applied to R4, R4B, R5 and R6. R4B emits no collection-status warnings, so nothing there turned the wrong type into a warning, but its `check()` still handed callers `SINGLETON`. DSTU2/DSTU2016May/DSTU3 have no `split()`.

**Tests**, per module in `FHIRPathTests` plus a dedicated `FHIRPathSplitCollectionTests` in R5 covering the `StructureDefinition.constraint` path, the unchanged evaluation results, and a control asserting `first()` on a genuine singleton still warns. R4 944, R5 1104, R6 1073 pass.

R4B has one pre-existing failure unrelated to this change: `name.prefix.join(',')`. `503ec1c0` ("Return empty from join() on an empty collection") added the `if (focus.isEmpty()) return new ArrayList<>()` guard to `funcJoin()` in R4, R5 and R6 but not R4B. Verified against an unmodified R4B: 938 tests / same single failure, versus 940 tests / same single failure with this change.


patrick-werner:
Both points covered.

**All versions**: applied to R4, R4B, R5 and R6. DSTU2, DSTU2016May and DSTU3 have no `split()` at all, so there is nothing to change there.

**Test cases**: FHIR/fhir-test-cases#285 adds `first()`, `last()`, `skip()` and `take()` on `split()` to the R4, R4B, R5 and R6 fhirpath files — none of the existing `testSplit1-4` used a positional collection function, so this combination was untested in every version.

One caveat: those cases pass with and without this fix, so they are coverage rather than a guard for this regression. The harness calls `fp.check(...)` and only catches exceptions, while `FHIRPATH_NOT_A_COLLECTION` is a warning collected via `checkOnTypes(..., warnings)` that no runner inspects. Verified by reverting the fix: testSplit5-8 still pass, while the engine-level type tests fail. That is why this PR also carries Java tests per engine. Filed separately as #2644.

For the R4B `join()` fix (#2642): `testJoinEmpty` already exists in all four fhirpath files from `c905291a` — that case is exactly what fails on R4B today, and the fix makes it pass.

grahamegrieve (approved):
ok

PR #2642, opened 2026-09-17
Return empty from join() on an empty collection in R4B

patrick-werner:
FHIRPath: "If the input is empty, the result is empty". `503ec1c0` added this guard to `funcJoin()` in R4, R5 and R6 but left R4B out, so R4B still returns an empty string where a null is expected — the shared fhir-test-cases suite catches it as `name.prefix.join(',')`.

Adds the same two tests that `503ec1c0` added to the other engines. R4B `FHIRPathTests` goes from 938 tests / 1 failure to 940 / 0; the full R4B module is 4251 / 0.

No release-notes entry, matching `503ec1c0`, which did not add one either.

Found while working on #2640 / #2641, but independent of it — this branch is off `master`.


grahamegrieve:
I can't see what this change actually does as compared to master?

PR #2643, opened 2026-09-17
Fix setting XHTML property in elementmodel

qligier:
While working on Matchbox, I got failing tests with the FHIR mapping language and transformations after updating to 6.10.4: `Error: Don't use setValue() for xhtml`.

After commit 56d75fbd051dee436787e42607cc8dc1613e97e2, Element.setProperty hasn't been updated to copy an XHTML value, which can't be set through setValue any more. StructureMapUtilities relies on setProperty, so some transformations failed.

I've updated the R5 and R6 Element classes, and added a simple test case with a minimalistic map copying an XHTML and another primitive value.

Original error and stracktrace for reference:
```
java.lang.Error: Don't use setValue() for xhtml

	at org.hl7.fhir.r5.elementmodel.Element.setValue(Element.java:271)
	at org.hl7.fhir.r5.elementmodel.Element.setProperty(Element.java:544)
	at org.hl7.fhir.r5.utils.structuremap.StructureMapUtilities.processTarget(StructureMapUtilities.java:2164)
	at org.hl7.fhir.r5.utils.structuremap.StructureMapUtilities.executeRule(StructureMapUtilities.java:1746)
	at org.hl7.fhir.r5.utils.structuremap.StructureMapUtilities.executeRule(StructureMapUtilities.java:1750)
	at org.hl7.fhir.r5.utils.structuremap.StructureMapUtilities.executeGroup(StructureMapUtilities.java:1733)
	at org.hl7.fhir.r5.utils.structuremap.StructureMapUtilities.transform(StructureMapUtilities.java:1695)
```


grahamegrieve (approved):
ok.

grahamegrieve (approved):
ok

PR #2645, opened 2026-09-17
2026 09 gg base adaptors

grahamegrieve:


org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/client/FHIRToolingClient.java
```diff
@@ -552,6 +571,29 @@ public String getAddress() {
...
+    } catch (IOException e) {
+      throw new FHIRException(e);
+    }
+    if (result.isUnsuccessfulRequest()) {
+      throw new EFhirClientException(result.getHttpStatus(), "Server returned error code " + result.getHttpStatus(), (OperationOutcome) result.getPayload());
+    }
+    return (ConceptMap) result.getPayload();
+  }
+
   public ConceptMap initializeClosure(String name) {
```
dotasek: issue: Not in this PR, but across all implementations of FHIRToolingClient, this appears to be uncalled. I think it should be deleted in each.

org.hl7.fhir.r4/src/main/java/org/hl7/fhir/r4/utils/CmdLineApp.java
```diff
@@ -39,7 +39,7 @@
 import org.hl7.fhir.utilities.npm.FilesystemPackageCacheManager;
 import org.hl7.fhir.utilities.npm.NpmPackage;
 
-
+@Deprecated
```
dotasek: issue: This is totally unused except in the main method in this class. I think we should add a note to the deprecation indicating it will be deleted and delete it (in 6 months?)

PR #2646, opened 2026-09-18
Import leftover Crowdin translations and remove Crowdin headers

vadi2:
This PR imports the 6,465 translations (plus 65 updates to outdated ones) that were stranded on the `l10n_crowdin_translations_master` branch, adds Ukrainian, fixes the Lithuanian plural forms, and removes the leftover `X-Crowdin-*` headers. The .properties files were regenerated with POGenerator. The imported strings were fixed so they display correctly: single apostrophes are doubled so text and values aren't lost, and two translator typos are corrected.


grahamegrieve:
Claude:

VALUE_SET_LOINCV (lt): "Loinc v" is translated as "Lina Vaitkevičiūtė", a person's name. It would appear wherever a LOINC version label is shown in Lithuanian-rendered guides. It could be machine translation reading "Loinc v" as a name plus initial, or a Crowdin contributor signing their work; I can't tell which.
VALUE_SET_LOINCV (uk): "Рішення у справі «Лоінк проти Німеччини» (Lo", meaning "Judgment in the case Loinc v. Germany". This is clearly a machine-translation hallucination, but it's equally unusable.

STRUC_DEF_SUIT_SHALL_CODE (lt, uk, uz): "Unless not suitable, these codes SHALL…" becomes "If not suitable, these codes SHALL…", which reverses the extensible binding rule. Russian has the same error, but it was already in master before this PR.
QUESTIONNAIRE_QR_ITEM_BAD_ATTACHMENT_MIMETYPE (ru): the text was copied from the date-maximum message, so a bad MIME type is reported as "value {0} exceeds maximum {1}".
REFERENCE_RESOLUTION_FAILED (ru): "failed" is missing, so a failed fetch reads like a progress note.
Questionnaire_QR_Item_GroupAnswer (ru): "not" is missing, so the rule is applied to group items, which are exactly the wrong ones.
FHIRPATH_LOCATION and "(at {n})" in This_property_must_be_* (lt): "at" is rendered as "ne", which means "not".
TYPE_SPECIFIC_CHECKS_DT_BASE64_NO_WS_WARNING (ru): the SHOULD NOT becomes a statement of fact, and the advice that readers should accept whitespace comes out reversed.
This_element_does_not_match_any_known_slice_ (ru): "this may not be a problem" becomes "this cannot be a problem".
Boolean messages (lt): "must be 'true' or 'false'" has the literal values translated, so users are told to enter 'teisinga' or 'klaidinga'.
Rendering labels that change what readers understand:
"Binding" is translated as "Mandatory" in lt and uk.
SHOULD becomes MUST in uk STRUC_DEF_SHOULD_CODE and CAPABILITY_SHOULD_SUPP.
SHOULD becomes "ZAL" ("WILL") in nl CONF_SHOULD and CONF_SHOULDNOT.
"draft" is dropped from ru SDR_DRAFT.
"Right Structure" becomes "Correct structure" in lt and uk profile comparisons.

Do you want to address some of these? the first two, in particular?

PR #2647, opened 2026-09-18
Add FHIRPathPatch: apply a FHIRPath Patch (the PATCH Parameters document) to a resource

costateixeira:
Replaces #2606. That PR added a resource manipulator with its own `{op, path, value|parts}` format and its own path parser. This one does the same job with the standard **FHIRPath Patch** document instead - the `Parameters` a FHIR server accepts on `PATCH` - so a patch written for a test step also works against a real server, and the path language is FHIRPath rather than something of ours.

## What's added

`org.hl7.fhir.services.utilities.FHIRPathPatch` applies a FHIRPath Patch to a resource in the element model. It lives in the services module because that is what `ValidationEngine` on master is built on now; there is no r5 copy, since nothing in r5 would call it. It implements the five operations of the [specification](http://hl7.org/fhir/fhirpatch.html):

| type | path selects | other parts |
| --- | --- | --- |
| `add` | the element that gets a new child (exactly one) | `name`, `value` |
| `insert` | a repeating element, e.g. `Patient.name` | `index`, `value` |
| `delete` | at most one element; nothing is a no-op | |
| `replace` | exactly one element | `value` |
| `move` | a repeating element | `source`, `destination` |

`value` is either a typed `value[x]` (primitive or complex data type, converted with `ObjectConverter`) or nested `part`s named after the child elements. Choice elements are named concretely (`deceasedBoolean`). Paths are evaluated with `FHIRPathEngine`; an operation whose path selects the wrong number of elements fails with a message naming the operation, the path and the count. Operations apply in order; the first failure throws and leaves the resource partly patched, which the javadoc says.

`ValidationEngine.patchResource(bytes, format, Parameters, outputFormat)` is the glue the HTTP layer will call: parse, apply, serialise. It validates nothing; the caller decides.

Nothing else changes. No profile is involved, which is why the "ProfileBased" name from #2606 is gone: the element model already carries the definitions the applier needs.

## Tests

`FHIRPathPatchTests`, in the standalone module on the shared worker context, no terminology server (standalone skips its tests by default: `-DskipTests=false`): each operation on a small Patient, values as typed `value[x]` and as parts, insert at start / end / into an empty list, the cardinality and range errors, and a malformed-document case. Tests that reorder children serialise the result and parse it back, so a tree that no longer serialises fails here rather than downstream.


grahamegrieve:
This needs a lot of work. First, it must be based on Base, not Element. That has some problems with existing code, but I asked Claude to evaluate it, and the list is long:
The biggest problems are that it gets choice-element naming the opposite way round from the spec, it doesn't check value types, and it has no limit on how long a client's path expression can run.

1. Base instead of Element

Yes, it can be done, and it's worth doing. The path evaluation already works on Base, and the five operations fit the Base API:

add / replace: makeProperty/addChild then fill, or setProperty.
delete: removeChild(name, value).
Definitions and cardinality: getNamedProperty.

On the generated classes setProperty casts the value to the property's type, so you'd get the type checking that is missing now (see 2) without writing it.

Three things block it at present:

The element model's Element only half-implements the Base contract. It doesn't override addChild, removeChild(String, Base) or getNamedProperty. Its listChildren lists only the children that are present, although Base says it should list every defined child. So through Base alone you can't ask an Element whether gender is a legal child, or whether it's a list. Element has the definitions, so adding those overrides is cheap.
Base has no way to insert or reorder at an index. Insert and move could remove all the list's values and add them back in the new order with setProperty, which appends for lists. Or you could add a small insertChild(name, index, value) hook to Base.
No parent pointer. Finding the parent for delete would walk the tree with getChildValues("*"), which the PR's findParent already effectively does.
2. Spec features not supported or not tested

The PR's own tests don't use the fhir-test-cases suite. These are the three cases that fail on it:

Case	Result
Add with choice element	fails: "BackboneElement has no element named 'time'"
Add extension	fails the same way (the extension's value part)
Full Resource	fails: "Don't use setValue() for xhtml"

Choice naming is backwards. The spec says the name part and nested part names use the bare name (deceased, time, value) and the type comes from value[x]. The PR requires the type-suffixed name (deceasedBoolean), and its addChoiceElement test uses that form. The spec form is rejected: "Patient has no element named 'deceased'".

No type checking, which the spec says servers SHALL do. All of these are applied without complaint:

birthDate replaced with valueBoolean gives "birthDate":"true".
birthDate replaced with "notadate" is accepted.
name[0] replaced with an Address gives "name":[{"city":"Lisbon"}].
Replacing Patient.deceased (currently deceasedBoolean) with a valueDateTime writes the date string into deceasedBoolean, which then serialises as false.

Replacing a choice element with a different type isn't supported. Replacing Observation.value (a Quantity) with valueString throws.

Other gaps:

Narrative: the div can't be added or replaced.
Contained resources: a value part that carries resource fails with "neither a value nor parts".
Extensions on a primitive value (e.g. valueString with an extension) are silently dropped. Only primitiveValue() is copied.
resolve() on a contained reference (#pr1) throws a NullPointerException, because the engine has no host services. That's the one use of resolve() the spec explicitly allows.
Insert and move work out the list's parent by cutting the path at its last .. So the path has to end in a plain element name, and a relative path like identifier, which the other operations accept, is refused.
Deleting the last children of an element leaves an empty element, and the JSON output contains "name":[null].
The result isn't validated. The spec says servers SHALL error if the result is invalid. The PR deliberately leaves that to the caller, but none of the cases above would be caught.

Not tested in the PR: choices in the spec form, extensions, narrative, contained resources and resolve(), extensions on primitives, wrong-type values, a delete that empties its parent, and paths that use where(). The simplest fix is to run the fhir-test-cases suite in the PR's tests.

3. Denial of service and other security issues

CPU denial of service is real. The path is client-controlled FHIRPath, and the engine has no limit on evaluation time or effort, apart from a 500 ms timeout on each regex call.

On a Patient with about 2,000 elements:

Patient.descendants().where(%resource.descendants().count() < 0) took 0.5 s.
Nesting it one level deeper took 437 s, from a single operation.
Deeper nesting grows polynomially: each extra level multiplies the time by roughly the element count. A patch can also contain any number of operations.

Once patchResource is exposed over HTTP, it needs an evaluation budget, a limit on path length and complexity, a cap on operation count, or all three.

Other issues:

resolve() is safe only by accident. Right now an external reference resolves to nothing because there are no host services. If someone later passes host services in, paths could reach other resources or the network, which the spec says SHALL NOT happen. The operation should check for this explicitly.
Terminology functions such as memberOf() run inside paths, so a client can trigger terminology server calls with value set URLs it chooses.
Operations aren't atomic: the first failure leaves the resource partly patched. patchResource works on a freshly parsed copy, so that's fine there, but other callers need to know.
Crash on resolve(): as well as being a spec gap, the NullPointerException from 2 is a 500 error any client can trigger.

PR #2649, opened 2026-09-19
Fold in additional bindings from obligation profiles

### LINKED ISSUE #2593: Obligation-profile additional bindings never propagate: inverted hasAdditional guard in updateFromDefinition

In `updateFromDefinition`, the block folding **additional bindings from obligation profiles** into the working element's binding ([PU:2947-2952](https://github.com/hapifhir/org.hl7.fhir.core/blob/b06c7ee7c57367305ff72b6bb09ce779b5dbd6ce/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java#L2947-L2952)) has an inverted guard:

```java
for (ElementDefinitionBindingAdditionalComponent ab : ed.getBinding().getAdditional()) {
  if (!Utilities.existsInList(ab.getPurpose().toCode(), "maximum", "required", "extensible")) {
    if (binding.hasAdditional(ab)) {          // <-- should be !binding.hasAdditional(ab)
      binding.getAdditional().add(ab.copy());
    }
  }
}
```

The parallel extension-form block a few lines earlier ([PU:2573-2578](https://github.com/hapifhir/org.hl7.fhir.core/blob/b06c7ee7c57367305ff72b6bb09ce779b5dbd6ce/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java#L2573-L2578)) correctly uses `!binding.hasAdditional(ab)`. As written, a *new* additional binding coming from an obligation profile is never added, and an already-present one is duplicated.

AltangerelG:
Fixes #2593.

`updateFromDefinition` folds an obligation profile's additional bindings into the
working element, but the guard is inverted:

```java
if (!Utilities.existsInList(ab.getPurpose().toCode(), "maximum", "required", "extensible")) {
  if (binding.hasAdditional(ab)) {     // adds only when already present
    binding.getAdditional().add(ab.copy());
```

So a new additional binding coming from an obligation profile was never added,
and one already present would be added a second time. The parallel block in
`updateFromObligationProfiles` has it the right way round, and this now matches
it.

Applied to `org.hl7.fhir.services` as well, per @grahamegrieve's note on #2632
and #2633 — it carries the same inverted guard, with `getAdditionalList()` and
`copy(Base.COPY_DATA)` in place of the r5 spellings.

## Tests

Two, in `org.hl7.fhir.r5`, since `org.hl7.fhir.services` has no test source
directory:

- `additionalBindingFromAnObligationProfileIsAdded` fails on master with
  `expected: <1> but was: <0>`, and passes with this change.
- `anAdditionalBindingAlreadyPresentIsNotDuplicated` passes either way. It is
  there so the first cannot be satisfied by adding unconditionally, which would
  reintroduce the duplication half of the bug.

`obligationProfiles` is private and populated from extensions on the derived
profile, which needs a populated worker context, so the test injects a profile
directly by reflection. There is precedent for that in the repo
(`CapabilityStatementUtilitiesTests`, `TurtleGeneratorTests`), but say if you
would rather it drove the extension path instead.

Fourth of @ewoutkramer's snapshot-generator review issues, after #2631, #2632 and
#2633. Independent of those three — different method, and it can merge in any
order.


AltangerelG:
Added the R6 counterpart of the obligation binding test in `org.hl7.fhir.standalone`, following your

```
  Tests run: 2, Failures: 0
```

and with the R6 half of the fix reverted, the first case fails. The module sets

of using reflection.

grahamegrieve (approved):
ok. thanks

PR #2650, opened 2026-09-19
fix issues in type convertor in R5 and R6

grahamegrieve:


org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/model/TypeConvertorPrimitiveTests.java
```diff
@@ -0,0 +1,176 @@
...
+  @Test
+  void testElementModel() throws Exception {
+    Element wrapper = new JsonParser(TestingUtilities.getSharedWorkerContext())
+      .parse("{\"value\":\"42\",\"_value\":{\"id\":\"p1\",\"extension\":[{\"url\":\"" + EXT_URL + "\",\"valueString\":\"x\"}]}}", "string", true);
+    Element e = wrapper.getNamedChild("value", false);
+    IntegerType i = TypeConvertor.castToInteger(e);
+    Assertions.assertEquals(42, i.getValue());
+    Assertions.assertEquals("p1", i.getId());
+    Assertions.assertEquals(EXT_URL, i.getExtension().get(0).getUrl());
+
```
dotasek, line 244: question: Do we care about testing that the extension value is "x" here? We're testing everything else, why not: 'Assertions.assertEquals("x", i.getExtension().get(0).getValue().primitiveValue());
'?

org.hl7.fhir.standalone/src/test/java/org/hl7/fhir/model/TypeConvertorPrimitiveTests.java
```diff
@@ -0,0 +1,181 @@
...
+  @Test
+  void testElementModel() throws Exception {
+    Element wrapper = new JsonParser(TestingUtilities.getSharedWorkerContext())
+      .parse("{\"value\":\"42\",\"_value\":{\"id\":\"p1\",\"extension\":[{\"url\":\"" + EXT_URL + "\",\"valueString\":\"x\"}]}}", "string", true);
+    Element e = wrapper.getNamedChild("value", false);
+    IntegerType i = TypeConvertor.castToInteger(e);
+    Assertions.assertEquals(42, i.getValue());
+    Assertions.assertEquals("p1", i.getId());
+    Assertions.assertEquals(EXT_URL, i.getExtensionList().get(0).getUrl());
+
```
dotasek, line 249: question: Do we care about testing that the extension value is "x" here? We're testing everything else, why not: 'Assertions.assertEquals("x", i.getExtension().get(0).getValue().primitiveValue());
'?

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/core/Enumeration.java
```diff
@@ -168,6 +168,14 @@ public Enumeration(EnumFactory<T> theEnumFactory, T theValue) {
     setValue(theValue);
   }
 
+  /**
+   * the enum factories map a code with no value (only an id and/or extensions) to their NULL
+   * constant, but the enumeration should have no value, like the code it came from
+   */
+  private static boolean hasNoValue(Element source) {
```
dotasek: issue: why not `hasValue`? I know most code quality tools will complain that this would always be inverted, but matching the PrimitiveType `hasValue` might be more consistent.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/model/Enumeration.java
```diff
@@ -118,14 +118,22 @@ public Enumeration(EnumFactory<T> theEnumFactory, T theValue) {
 		setValue(theValue);
 	}
 	
+  /**
+   * the enum factories map a code with no value (only an id and/or extensions) to their NULL
+   * constant, but the enumeration should have no value, like the code it came from
+   */
+  private static boolean hasNoValue(Element source) {
```
dotasek: issue: why not hasValue? I know most code quality tools will complain that this would always be inverted, but matching the PrimitiveType hasValue might be more consistent.

dotasek (changes requested):
See issues, but also consider checking canonical, URL, and URI with dedicated regexs.

org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/model/Enumeration.java
grahamegrieve: ok

org.hl7.fhir.model/src/main/java/org/hl7/fhir/model/core/Enumeration.java
grahamegrieve: ok

org.hl7.fhir.standalone/src/test/java/org/hl7/fhir/model/TypeConvertorPrimitiveTests.java
grahamegrieve: ok

org.hl7.fhir.r5/src/test/java/org/hl7/fhir/r5/model/TypeConvertorPrimitiveTests.java
grahamegrieve: ok

dotasek:
> See issues, but also consider checking canonical, URL, and URI with dedicated regexs.

Spec defines these as `\S*`, NOT as actual URL/URI. See: https://www.hl7.org/fhir/datatypes.html#uri

dotasek (approved):
Approved, assuming further commits resolve comments.

PR #2653, opened 2026-09-20
Fix NPE and message arity in snapshot generation (#2594)

### LINKED ISSUE #2594: Two small snapshot-generation fixes: ATTEMPT_TO_CHANGE_SLICING message arity; NPE risk in chooseMatchingBaseSlice
Found during a systematic side-by-side review of the Java and .NET snapshot generators — bundling two small, low-impact items into one issue. Line references are pinned to commit b06c7ee (master, 2026-08-21) via permalinks; verified still present on master `4f52ba6` (2026-09-01).

**1. `ATTEMPT_TO_CHANGE_SLICING` info message passes 2 args to a 4-placeholder message.**

The INFORMATION-severity variant (slicing restated *identically* on a later slice element) calls `formatMessage(I18nConstants.ATTEMPT_TO_CHANGE_SLICING, id0, idI)` ([PPP:355-357](https://github.com/hapifhir/org.hl7.fhir.core/blob/b06c7ee7c57367305ff72b6bb09ce779b5dbd6ce/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfilePathProcessor.java#L355-L357)), but the message text is "The element at {0} defines the slicing {1} but then an element in the slicing {2} tries to redefine the slicing to {3}" — `{1}` renders an element id where a slicing summary belongs and `{2}`/`{3}` render unfilled. The ERROR variant (PPP:351-353) passes all four. Either pass the same four args or use a dedicated "slicing restated redundantly" message.

**2. NPE risk in `chooseMatchingBaseSlice` for a base slice with no type.**

`chooseMatchingBaseSlice` ([PU:1732-1739](https://github.com/hapifhir/org.hl7.fhir.core/blob/b06c7ee7c57367305ff72b6bb09ce779b5dbd6ce/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java#L1732-L1739)) compares `bs.getType().equals(type)`; `findBaseSlices` (PU:1742-1758) builds `BaseTypeSlice` with `type = sliceElement.getTypeFirstRep().getCode()`, which is null for a typeless base slice row → `null.equals(...)` NPE on the next type-slicing pass over that base. Guard the null.

AltangerelG:
Fixes #2594 — both of the two issues reported there.

### 1. NPE in `chooseMatchingBaseSlice`

```java
if (bs.getType().equals(type)) {
```

`BaseTypeSlice.getType()` may be null, so a base slice with no type threw
`NullPointerException` rather than simply not matching. Guarded with a null check:
a typeless slice now fails to match, and a later slice in the list can still match.

### 2. `ATTEMPT_TO_CHANGE_SLICING` called with the wrong number of arguments

The message takes four parameters — two ids and two slicing summaries. The ERROR
variant passes all four; the INFORMATION variant directly below it passed only the
two ids, so both slicing summaries came out empty in the rendered message. It now
passes the same four arguments as the ERROR variant.

### Scope

Both fixes are applied in R5 and in the R6 sources under `org.hl7.fhir.services`:

- `org.hl7.fhir.r5/.../profile/ProfileUtilities.java`
- `org.hl7.fhir.r5/.../profile/ProfilePathProcessor.java`
- `org.hl7.fhir.services/.../profile/ProfileUtilities.java`
- `org.hl7.fhir.services/.../profile/ProfilePathProcessor.java`

### Tests

`ProfileUtilitiesBaseSliceTest` covers the null guard, once in `org.hl7.fhir.r5` and
once in `org.hl7.fhir.standalone` for the R6 code: a typeless slice does not match, a
typeless slice does not hide a later matching one, and a matching slice is still returned.

Checked that each test fails without the fix — reverting the guard gives
`Tests run: 3, Failures: 0, Errors: 2`, both the expected NPE, in the R5 and the R6
suite respectively; with the fix both run 3/0/0.

The message-arity change is not covered by a test — it alters only the text of an
INFORMATION message, and there was no existing assertion on that output to extend.


grahamegrieve:
this needs proper test cases in the test-cases repository. I think

PR #2654, opened 2026-09-20
Fix three defects in the additional-base merge (#2591)

### LINKED ISSUE #2591: Additional-base merge: pattern/pattern branch passes the getFixed() operands to checkPatternValues

In `SnapshotGenerationPreProcessor.mergeElementDefinitions` (the additional-base merge, [PRE:449-450](https://github.com/hapifhir/org.hl7.fhir.core/blob/b06c7ee7c57367305ff72b6bb09ce779b5dbd6ce/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/SnapshotGenerationPreProcessor.java#L449-L450)):

```java
} else if (source.hasPattern() && base.hasPattern()) {
  merged.setPattern(checkPatternValues(baseSD.getVersionedUrl(), source.getPath()+".pattern", source.getFixed(), base.getFixed(), true));
```

In this branch neither side has `fixed` (the enclosing `if (source.hasFixed() || base.hasFixed())` handled that), so both operands are null — `checkPatternValues` NPEs or misbehaves. Clearly meant `source.getPattern(), base.getPattern()`.

Two related small defects in the same method family:

- Dead disjunct in `checkPatternValues` at [PRE:670](https://github.com/hapifhir/org.hl7.fhir.core/blob/b06c7ee7c57367305ff72b6bb09ce779b5dbd6ce/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/SnapshotGenerationPreProcessor.java#L670): `p1.getValues().size() > 1 || p1.getValues().size() > 2` — the second clause is unreachable; presumably meant `p2.getValues().size()`.
- `isLower`'s Quantity guard is inverted ([PRE:638-642](https://github.com/hapifhir/org.hl7.fhir.core/blob/b06c7ee7c57367305ff72b6bb09ce779b5dbd6ce/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/SnapshotGenerationPreProcessor.java#L638-L642)): it **throws when the units are equal** and happily compares values when the units differ — the opposite of the intended "only comparable when units match".

AltangerelG:
Fixes #2591 — all three defects reported there, in `SnapshotGenerationPreProcessor`.

### 1. The pattern/pattern branch passed the `fixed` operands

```java
} else if (source.hasPattern() && base.hasPattern()) {
  merged.setPattern(checkPatternValues(..., source.getFixed(), base.getFixed(), true));
```

The enclosing `if (source.hasFixed() || base.hasFixed())` has already established that
neither side has a fixed value, so both operands are always null and `checkPatternValues`
NPEs on `v1.fhirType()`. Now passes `source.getPattern(), base.getPattern()`.

### 2. Dead disjunct in `checkPatternValues`

```java
if (p1.getValues().size() > 1 || p1.getValues().size() > 2) {
```

The second clause is unreachable. Both sides are dereferenced with `get(0)` on the very
next line, and the sibling branch below already guards `p2.getValues().size() > 1`, so the
intent was to guard `p2` here too — otherwise extra values on the base pattern are dropped
without complaint. Changed to `p1.getValues().size() > 1 || p2.getValues().size() > 1`.

### 3. `isLower`'s Quantity guard was inverted

```java
if (Utilities.stringsEqual(q1.getUnit(), q2.getUnit())) {
  throw new FHIRException(... path+"."+property+".unit" ...);
}
```

This threw when the units were *equal* and fell through to comparing the values when they
*differed* — the opposite of "only comparable when the units match". Negated.

### Scope

All three applied in R5 and in the R6 sources under `org.hl7.fhir.services`:

- `org.hl7.fhir.r5/.../profile/SnapshotGenerationPreProcessor.java`
- `org.hl7.fhir.services/.../profile/SnapshotGenerationPreProcessor.java`

### Tests

`SnapshotGenerationPreProcessorPatternMergeTest`, four tests, once in `org.hl7.fhir.r5`
and once in `org.hl7.fhir.standalone` for the R6 code: patterns merge from the pattern
operands, a repeating child on the base pattern is refused rather than truncated, equal
units compare, and differing units are rejected.

Verified each test fails without its fix. Reverting the three defects one at a time in R5
fails exactly the tests that target each:

| reverted | failing tests |
| --- | --- |
| pattern operands | `patternsAreMergedFromThePatternOperands`, `aRepeatingChildOnTheBasePatternIsNotSilentlyTruncated` |
| dead disjunct | `aRepeatingChildOnTheBasePatternIsNotSilentlyTruncated` |
| inverted unit guard | `quantitiesWithTheSameUnitAreCompared`, `quantitiesWithDifferentUnitsAreRejected` |

Reverting all three in R6 fails all four standalone tests; restoring them passes 4/4.

`mergeElementDefinitions` and `isLower` are private and their callers need a fully
navigable pair of profiles, so the tests reach them by reflection, following the existing
`SnapshotGenerationPreProcessorExtensionSlicingTest`. Happy to relax the modifiers instead
if you prefer.

### One thing I did not change

The unit-mismatch message reports `v1.fhirType(), v2.fhirType()`, which for two Quantities
renders as "Quantity" and "Quantity" — no help on a `.unit` path. It looks like it should
report the two units, but that is beyond what the issue describes, so I have left it alone.
Say the word and I will include it.


grahamegrieve:

Issue #2589, opened 2026-09-01
UNSUPPORTED_SLICING_COMPLEXITY bail-out abandons ALL slicings, not just the offending one


ewoutkramer:
Found during a systematic side-by-side review of the Java and .NET snapshot generators. Line references are pinned to commit b06c7ee (master, 2026-08-21) via permalinks; verified still present on master `4f52ba6` (2026-09-01).

When one slicing's trailing region contains one non-extension nested slicing, `SnapshotGenerationPreProcessor.processSlices` logs a warning and `return`s ([PRE:713-723](https://github.com/hapifhir/org.hl7.fhir.core/blob/b06c7ee7c57367305ff72b6bb09ce779b5dbd6ce/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/SnapshotGenerationPreProcessor.java#L713-L723)) — abandoning trailing-content propagation for **every** slicing in the differential and skipping the final `markExtensions` pass, not just skipping the offending SliceInfo.

A log-warn is also easy to miss for something that changes the generated snapshot substantially.

Suggested direction: skip only the offending SliceInfo, and/or surface a `ValidationMessage` instead of a log line.

AltangerelG:
Confirmed on current master: the pre-pass at :713-723 `return`s out of `processSlices`
entirely, so one unsupported nested slicing abandons trailing-content propagation for every
other slicing and skips the final `markExtensions` pass too.

Worth noting for whoever picks this up: "skip only the offending SliceInfo" is not a local
change. The later loops iterate `slicings` directly, so the offending entries have to be
collected in the pre-pass and then skipped in each subsequent pass — a small restructure,
not a `return` to `continue`.

I stopped there rather than guessing, because the issue says "suggested direction" and both
halves are behaviour changes:

1. Skipping only the offender means profiles that currently get *no* pre-processing would
   start getting partial pre-processing, which changes generated snapshots in the field.
2. Raising the `log.warn` to a `ValidationMessage` — at what severity? ERROR would fail
   publication for profiles that build clean today.

Related: #2605 (PR #2655) removes one common way of reaching this bail-out at all, by
treating an absent `slicing.ordered` as "not ordered" so ordinary hand-written extension
slicings stop being classified as unsupported nested slicing. That shrinks the blast radius
here but does not fix the all-or-nothing behaviour.

Tell me which shape you want and I will do it in R5 and R6 with tests.

PR #2655, opened 2026-09-20
Treat absent slicing.ordered as not ordered in isExtensionSlicing (#2605)

### LINKED ISSUE #2605: SnapshotGenerationPreProcessor: an extension slicing entry without an explicit `ordered=false` is treated as unsupported nested slicing, and the whole profile's slice pre-processing is abandoned with only a warning
**Where:** `SnapshotGenerationPreProcessor.isExtensionSlicing`, master `f725fb7` line 1077 (same at b06c7ee):

```java
private boolean isExtensionSlicing(ElementDefinition ed) {
  if (!Utilities.existsInList(ed.getName(), "extension", "modiferExtension")) {
    return false;
  }
  if (ed.getSlicing().getRules() != SlicingRules.OPEN || (!ed.getSlicing().hasOrdered() || ed.getSlicing().getOrdered()) || ed.getSlicing().getDiscriminator().size() != 1) {
    return false;
  }
  ...
}
```

`(!hasOrdered() || getOrdered())` makes the exemption require an **explicit** `ordered = false`. `ElementDefinition.slicing.ordered` is 0..1 and its absence means "not ordered" in the spec, so a perfectly ordinary hand-written extension slicing entry

```xml
<element id="Patient.identifier.extension">
  <path value="Patient.identifier.extension"/>
  <slicing>
    <discriminator><type value="value"/><path value="url"/></discriminator>
    <rules value="open"/>
  </slicing>
</element>
```

placed between an outer slicing entry and its first named slice is *not* recognised as an extension slicing. `processSlices` (:695-720) then hits the "unsupported nested slicing" check and `return`s — abandoning slice pre-processing for the **entire** profile (cf. #2589) — with a `log.warn` only:

```
Unsupported feature in slicing at Patient.identifier: the slicing element at path Patient.identifier.extension has slicing open, by value=url, and this nested slicing is not supported
```

**Observed:** three otherwise identical profiles (validator_cli 6.10.2, JUnit-driver config): (a) with the entry above → warning, nothing from the slicer's sub-tree propagates into the named slice; (b) without any explicit extension slicing entry → pre-processing runs; (c) with the same entry plus `<ordered value="false"/>` → pre-processing runs, with the same propagation into the named slice as (b). Tools that always emit `ordered: false` (SUSHI) never see this; hand-authored or other-tool profiles do, and get a silently different snapshot.

(Aside, already filed as #2588: the same list contains the typo `"modiferExtension"`.)

**Suggested fix:** treat absent `ordered` as `false` in `isExtensionSlicing`; consider raising the bail-out to an ERROR-level `ValidationMessage` so the author learns that slice pre-processing was skipped.

Found while reverse-engineering snapshot generation for a spec write-up (Firely); the three fixtures are available on request.

AltangerelG:
Fixes #2605.

`isExtensionSlicing` required an **explicit** `ordered = false` before it would
exempt an entry from slice pre-processing:

```java
if (ed.getSlicing().getRules() != SlicingRules.OPEN
    || (!ed.getSlicing().hasOrdered() || ed.getSlicing().getOrdered())
    || ed.getSlicing().getDiscriminator().size() != 1) {
  return false;
}
```

`ElementDefinition.slicing.ordered` is 0..1 and its absence means "not ordered", so an
ordinary hand-written extension slicing entry that simply omits it was not recognised.
`processSlices` then hit the unsupported-nested-slicing check and returned, abandoning
slice pre-processing for the **entire** profile with only a `log.warn` — a silently
different snapshot. Tools that always emit `ordered: false` never see it.

`getOrdered()` already returns `false` when the element is absent:

```java
public boolean getOrdered() {
  return this.ordered == null || this.ordered.isEmpty() ? false : this.ordered.getValue();
}
```

so the `hasOrdered()` clause is the whole problem and the condition reduces to
`ed.getSlicing().getOrdered()`.

### Scope

- `org.hl7.fhir.r5/.../profile/SnapshotGenerationPreProcessor.java`
- `org.hl7.fhir.services/.../profile/SnapshotGenerationPreProcessor.java`

### Tests

Extends the existing `SnapshotGenerationPreProcessorExtensionSlicingTest` in
`org.hl7.fhir.r5` and in `org.hl7.fhir.standalone`, from 3 tests to 5: an entry that
leaves `ordered` absent is recognised, and one with `ordered = true` still is not.

Its fixture used to carry this comment —

> Ordered is set explicitly, since the predicate does not accept a slicing that leaves it absent.

— which was this bug, written down as though it were intended. Removed.

Verified both ways: 5/5 with the fix in each tree, and restoring the `hasOrdered()` clause
fails exactly one test, `extensionSlicingWithoutAnExplicitOrderedIsRecognised`, in each.
The other four set `ordered` explicitly and are unaffected either way.

### Not included

The issue also suggests raising the `processSlices` bail-out from `log.warn` to an
ERROR-level `ValidationMessage`, so an author learns that pre-processing was skipped. That
is a reporting-behaviour change and overlaps #2589, so I have left it out of this PR.
Happy to do it here or separately, whichever you prefer.


grahamegrieve:

Issue #2603, opened 2026-09-03
Snapshot generation: MappingAssistant mishandles a mapping-identity collision between base and derived profile (rename applied to the wrong side, and to `name` instead of `identity`)


ewoutkramer:
**Setup:** a base profile declares `StructureDefinition.mapping` with `identity = m`, `uri = http://example.org/maps/A`, and uses it on `Patient.active`. A derived profile re-declares `identity = m` with a different `uri = http://example.org/maps/B` (different `name` too) and uses `m` (meaning B) on `Patient.gender`.

`MappingAssistant` (master `f725fb7`, same lines at b06c7ee) detects the collision and decides to rename the base's `m` to `m1` (`nameExists` → `renames.put("m", "m1")`, MA:69-76). Two things then go wrong:

1. **The rename lands in `name`, not `identity`** (MA:76): `masterList.add(m.copy().setName(n));` — so the derived SD ends up with **two** `mapping` entries whose `identity` is `m` (one named "m1"), and no declaration with `identity = m1` exists at all.

2. **The rename table is applied to the differential's element mappings instead of the inherited ones.** `updateFromDefinition` calls `mappings.merge(derived, base); // note reversal of names to be correct in .merge()` (PU:3084), and `merge(ElementDefinition base, ElementDefinition derived)` does `addMappings(list, base.getMapping(), renames)` … `addMappings(list, derived.getMapping(), null)` (MA:173-177). With the caller's reversal, `renames` (keyed on *base-SD* identities) is applied to the **differential's** mappings — which carry the derived SD's identities — while the inherited element mappings, the ones that actually use the base identity, are copied unrenamed.

**Observed output** (validator_cli 6.10.2, JUnit-driver config, R5 core):

```xml
<!-- StructureDefinition.mapping of the derived snapshot -->
<mapping><identity value="m"/><uri value="http://example.org/maps/B"/><name value="Map B (declared in the derived profile)"/></mapping>
<mapping><identity value="m"/><uri value="http://example.org/maps/A"/><name value="m1"/></mapping>
…
<!-- Patient.active (inherited from the base profile) -->
<mapping><identity value="m"/><map value="A.active (base mapping, identity m = Map A)"/></mapping>   <!-- still m: now resolves to Map B -->
<!-- Patient.gender (the derived profile's own row) -->
<mapping><identity value="m1"/><map value="B.gender (derived mapping, identity m = Map B)"/></mapping> <!-- m1: declared nowhere -->
[...]

PR #2659, opened 2026-09-20
Give the renamed mapping declaration the new identity (#2603)

AltangerelG:
Fixes **half** of #2603. Please read the second section before reviewing — I have removed
the other half, and I think the issue is wrong about it.

### What this fixes

When a derived profile re-declares a mapping identity the base already uses for a different
URI, `MappingAssistant` renames the base's declaration out of the way. The rename was
written into `name`:

```java
masterList.add(m.copy().setName(n));
```

so the derived SD ended up with two declarations both claiming `identity = m` (one merely
*named* "m1"), and nothing declaring `identity = m1`. `findDefinition` resolves by identity,
so an element mapping naming the renamed map resolved to nothing.

One line in each tree, plus a test in `org.hl7.fhir.r5` and in `org.hl7.fhir.standalone`.
`SnapShotGenerationTests` stays at 164/0/0, and reverting the fix fails the test.

### What I removed, and why — the second defect does not hold up

The issue also reports that `merge` applies the rename table to the differential's mappings
instead of the inherited ones, and that the operands should swap. I implemented that, and it
**fails three cases** in `SnapShotGenerationTests` that pass on master:
`profile-mapping-3`, `pat-choice-ms`, `au-med-k`.

`profile-mapping-3` is the clearest. Base declares `map1 -> http://example.org/map1`; the
derived re-declares the same URI as `map2`, so `renames = {map1 -> map2}`. The base's
`Patient.birthDate` carries `map1 = "some.path.here"`, the differential carries
`map2 = "some.other.path.here"`.

| | `Patient.birthDate` mappings |
|---|---|
| master, and expected | `map2 = "some.other.path.here,some.path.here"` — appended, no orphan |
| with the operands swapped | `map2 = "some.other.path.here"` **and** `map1 = "some.path.here"` |

`map1` is declared nowhere in that snapshot, so the swap *creates* exactly the orphan the
issue is about. The current assignment is getting the inherited row renamed and merged
correctly.

I could not reconstruct from the code why that is so — following `base = dest` /
`derived = source` and the reversed call at `ProfileUtilities:3094`, I expected the renames
to land on the differential. The output says otherwise, and that gap is itself a reason not
to change it on my say-so.

There *is* still something wrong nearby: `pat-choice-ms` expects `w51` on ten elements while
declaring only `w5`, which is an undeclared identity in a golden file. But it is not fixed by
swapping these operands, and `update()`'s `notExcluded` resolving element identities against
`base.getMapping()` looks a likelier cause. I have written that up on the issue.

Happy to take another run at it with direction.


grahamegrieve:
this needs proper test cases in the test-cases repository. But probably we need to decide on this as a group

PR #2661, opened 2026-09-20
Check slice groups that reach the end of the snapshot (#2587)

### LINKED ISSUE #2587: Slice-cardinality sum check never runs for slice groups at the end of the snapshot

The post-generation sweep in `ProfileUtilities` that checks slice min/max sums (and silently raises auto-added entries' `min`) — [PU:976-1036](https://github.com/hapifhir/org.hl7.fhir.core/blob/b06c7ee7c57367305ff72b6bb09ce779b5dbd6ce/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java#L976-L1036) — flushes an `ElementDefinitionCounter` only when a later element with shallower-or-equal path depth appears. Counters still open when the element loop ends are **abandoned**: a sliced element whose slice group extends to the last snapshot element is never checked or adjusted. Whether the check runs thus depends on where the sliced element happens to sit in the resource.

Suggested fix: flush remaining counters after the loop.

Related cosmetic issue in the same block: the min>max warning ([PU:1012-1016](https://github.com/hapifhir/org.hl7.fhir.core/blob/b06c7ee7c57367305ff72b6bb09ce779b5dbd6ce/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java#L1012-L1016)) prints the entry's own min/max although the failed comparison was between the *sums* (`checkMinMax` compares countMin vs countMax) — the message text can't match the numbers the check used.

AltangerelG:
Fixes #2587.

The slice min/max sum check opens an `ElementDefinitionCounter` when it meets a sliced
element, and flushes it when a later element of shallower-or-equal path depth turns up.
Counters still open when the element loop ended were simply **abandoned**.

So a sliced element whose slice group runs to the last row of the snapshot was never
checked — and an auto-added slicing entry there never had its `min` raised. Whether the
check ran at all depended on where the sliced element happened to sit in the resource.

### What changed

The sweep moves out of `generateSnapshot` into `checkSliceCardinalitySums(derived)`, and the
three per-counter checks into `checkSliceCounts(slice)`, so the same checks can run from
inside the loop and once more after it:

```java
// counters still open when the elements run out belong to slice groups that reach the
// end of the snapshot; without this they were never checked or adjusted at all
for (ElementDefinitionCounter slice : slices.values()) {
  checkSliceCounts(slice);
}
```

The extraction is a pure move — no logic edited, only indentation and the loop counter made
local. I did it because the check was otherwise only reachable through a full snapshot
generation, and I wanted the fix under test. Happy to inline it again if you would rather
keep `generateSnapshot` whole.

### Scope

- `org.hl7.fhir.r5/.../profile/ProfileUtilities.java`
- `org.hl7.fhir.services/.../profile/ProfileUtilities.java`

### Tests

`ProfileUtilitiesSliceSumTest`, three tests, in `org.hl7.fhir.r5` and in
`org.hl7.fhir.standalone` for the R6 code: a slice group that reaches the last element is
still checked, an auto-added slicer in that position has its `min` raised, and a group
followed by another element keeps working as before.

Verified by removing only the post-loop flush. The two trailing-group tests fail; the third
keeps passing, which is what shows the extraction did not disturb the existing path:

```
ProfileUtilitiesSliceSumTest.slicesThatRunToTheEndOfTheSnapshotAreStillChecked   FAILED
ProfileUtilitiesSliceSumTest.aTrailingSliceGroupOnAnAutoAddedSlicerHasItsMinRaised  FAILED
Tests run: 3, Failures: 2
```

Same in R6: 3/0/0 with the flush, 2 failures without it.

One fixture note, in case it reads oddly: the slicer is left at `min = 0` with two slices of
`min = 1`, because `checkMin()` only reports when the slices sum to **more** than the
slicer's own min. My first fixture set both to 2 and reported nothing.

### Not included

The issue also notes that the min>max warning prints the entry's own min/max although
`checkMinMax` compares the *sums*, so the text cannot match the numbers the check used.
That is a message-wording change rather than this defect, so I have left it out — say the
word and I will add it here or separately.


grahamegrieve:

PR #2662, opened 2026-09-20
Guard the empty baseSlices case (#2585)

### LINKED ISSUE #2585: IndexOutOfBoundsException in processPathWithSlicedBaseWhereDiffsConstrainTypes when findBaseSlices returns no slices

`ProfilePathProcessor.processPathWithSlicedBaseWhereDiffsConstrainTypes` ends with

```java
cursors.baseCursor = baseSlices.get(baseSlices.size() - 1).getEnd() + 1;
```

([PPP:1652](https://github.com/hapifhir/org.hl7.fhir.core/blob/b06c7ee7c57367305ff72b6bb09ce779b5dbd6ce/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfilePathProcessor.java#L1652)), which throws `IndexOutOfBoundsException` when `baseSlices` is empty. Two ways to get there:

1. the base snapshot has a slicing entry on the `[x]` element but zero named slice rows;
2. the slicing entry has children in the base snapshot: [`findBaseSlices` (PU:1742-1758)](https://github.com/hapifhir/org.hl7.fhir.core/blob/b06c7ee7c57367305ff72b6bb09ce779b5dbd6ce/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java#L1742-L1758) is called with `newBaseLimit` = the end of the slicer's own scope, so it anchors on the slicer's **last child**; its same-path scan then keys on the child's path, matches no named slice, and returns an empty list.

Suggested fix: guard the empty case (fall back to `newBaseLimit + 1`).

AltangerelG:
Fixes #2585.

`processPathWithSlicedBaseWhereDiffsConstrainTypes` ends with

```java
cursors.baseCursor = baseSlices.get(baseSlices.size() - 1).getEnd() + 1;
```

which throws `IndexOutOfBoundsException` when `findBaseSlices` returned an empty list — a
base slicing entry with no named slice rows, or a slicing entry that has children, where
`findBaseSlices` is called with `newBaseLimit` at the end of the slicer's own scope and so
anchors on the slicer's **last child**, whose path then matches no named slice.

Falls back to `newBaseLimit + 1`, as the issue suggests: with no slices found, the slicer's
own scope is what has just been dealt with.

### Scope

- `org.hl7.fhir.r5/.../profile/ProfilePathProcessor.java`
- `org.hl7.fhir.services/.../profile/ProfilePathProcessor.java`

### On testing — no test, and here is why

I could not write an honest one. The method recurses into `processPaths`, so reaching it
needs a real base and differential driven through snapshot generation rather than a
constructed call. The harness for that, `SnapShotGenerationTests`, loads its cases from the
`hl7.fhir.test.cases` package through `TestingUtilities.loadTestResourceStream` — they are
not in this repository, so the fixture would have to land in `fhir-test-cases` first.

I would rather say that plainly than add a test that exercises something adjacent and reads
like coverage. If you would like the fixture contributed to `fhir-test-cases`, I am happy to
do that and link it here.


grahamegrieve:
indeed, this and others need a test case in the test-cases repository

